PR #116, opened 2026-02-24
Prove inflate_deflateLazy — Level 2 end-to-end native roundtrip

kim-em:
Closes #104

Session: `unknown`

2cdcce7 doc: Add progress entry for lazy LZ77 correctness session
e31a431 feat: Prove lz77Lazy native correctness — BB1 for lazy compressor
ee45c9b chore: add partial completion and cross-issue dependency prompting

🤖 Prepared with Claude Code

PR #117, opened 2026-02-24
chore: use needs-replan label for partial completions

kim-em:
Improves the partial completion flow introduced in #116:

- `--partial` now adds `needs-replan` label instead of returning issue directly to unclaimed queue (prevents workers from re-claiming stale scope before a planner re-scopes)
- `_unclaimed_issues` excludes `needs-replan` from the queue
- Fix: existing-PR early-return path now correctly handles `--partial` label transition
- Planner instructions: close `needs-replan` issue and create a fresh issue for remaining work (preserves history via bidirectional links)

🤖 Prepared with Claude Code


kim-em:
Changes included directly in master via history rewrite.

Issue #243, opened 2026-02-24
Complete zlib roundtrip: fill remaining 2 sorries in ZlibCorrect.lean


kim-em:
**Session type**: implementation (Phase 4+ — zlib framing roundtrip proof)

depends-on: #238

## Current state

- 4 sorries total: 2 in GzipCorrect.lean, 2 in ZlibCorrect.lean
- Both gzip and zlib roundtrips are blocked on tight endPos infrastructure
- Issue #238 (gzip roundtrip) covers `inflateLoop_endPos_le` and the gzip proof
- ZlibCorrect.lean already has the proof structure for `zlib_decompressSingle_compress`
  with two targeted sorries (from PR #241, the partial progress on #233):
  1. `hendPos_tight`: `endPos + 4 ≤ (ZlibEncode.compress data level).size`
  2. `hadler`: Adler32 trailer bytes at endPos match the encoder's output

Replaces: #233 (replan — partial progress in PR #241)

## Goal

Complete the zlib roundtrip proof by filling the two remaining sorries in
`Zip/Spec/ZlibCorrect.lean`.

## Why this depends on #238

Both sorries require knowing that `inflateRaw` returns an endPos that
falls within the DEFLATE portion of the compressed ByteArray, NOT in the
trailer. Specifically:

- `compressed = header(2) ++ deflated ++ trailer(4)` (from `compress_eq`)
- `inflateRaw compressed 2` returns `(data, endPos)` (already proved)
- We need `endPos ≤ 2 + deflated.size` (so trailer is after endPos)

The simple `inflateRaw_endPos_le` (`endPos ≤ compressed.size`) is NOT
tight enough — it gives `endPos ≤ 6 + deflated.size`, but we need
`endPos + 4 ≤ 6 + deflated.size`, i.e., `endPos ≤ 2 + deflated.size`.

Issue #238 tackles this same problem for the gzip case (needs
`endPos ≤ 10 + deflated.size` when `compressed.size = 18 + deflated.size`).
Whatever endPos infrastructure #238 builds will directly unblock this issue.

### 1. Prove `hendPos_tight`

```lean
have hendPos_tight : endPos + 4 ≤ (ZlibEncode.compress data level).size
```

After #238 merges, use the same endPos infrastructure (whether it's a
tighter `inflateLoop_endPos_le`, an `inflateRaw_prefix_suffix_invariant`,
or a direct endPos-tracking theorem) to show the endPos from
[...]

PR #280, opened 2026-02-25
feat: prove hendPos_tight in zlib roundtrip

kim-em:
Partial progress on #243

Session: `d7671b84-b59a-4ec1-8378-24835ddf9991`

4e900ea doc: add progress entry for session 4c6006db
27f109b feat: prove hendPos_tight in zlib roundtrip via suffix independence

🤖 Prepared with Claude Code

### DIFF
--- a/progress/20260225T052118Z_4c6006db.md
+++ b/progress/20260225T052118Z_4c6006db.md
@@ -1,0 +1,43 @@
+# Progress: Fill hendPos_tight sorry in zlib roundtrip
+
+**Session**: 4c6006db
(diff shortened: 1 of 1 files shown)


kim-em:
Closing: this PR proved hendPos_tight but has two issues:
1. CI failure (maxRecDepth exceeded in ZlibCorrect.lean)
2. Redundant InflateSuffix.lean with sorry — the theorem already exists as inflateRaw_append_suffix in GzipCorrect.lean

The hendPos_tight proof approach is correct (using suffix independence) — the replacement issue for #243 will use the same approach but with the existing inflateRaw_append_suffix lemma.

Issue #622, opened 2026-03-06
Track E: Fix and prove weightsToMaxBits_valid — corrected while-loop bound


kim-em:
## Current state

- **Track E (Zstd)**: Active implementation + spec phase
- **Sorry count**: 7 total (1 Fse.lean, 1 Zstd.lean, 5 ZstdHuffman.lean)
- **ZstdHuffman.lean sorry breakdown**:
  - `isPow2_iff` (line 60) — covered by #617
  - `buildZstdHuffmanTable_tableSize` (line 106) — not covered
  - `buildZstdHuffmanTable_maxBits_pos` (line 114) — not covered
  - `weightsToMaxBits_valid` (line 129) — THIS ISSUE
  - `weightSum_pos_of_exists_nonzero` (line 144) — covered by #617

## Background — CRITICAL: Theorem statement is FALSE

The current theorem statement is:

```lean
theorem weightsToMaxBits_valid (weights : Array UInt8)
    (m : Nat)
    (h : Zip.Native.weightsToMaxBits weights = .ok m) :
    weightSum weights > 0 ∧ weightSum weights ≤ 1 <<< (m - 1) := by
  sorry
```

**The second conjunct is provably false.** Counterexample:
- Input: `weights = #[1, 2]`
- `weightSum`: w=1 → 2^0=1, w=2 → 2^1=2, total = 3
- `weightsToMaxBits` computation:
  - maxBits=0, power=1
  - Loop: 1<3 → maxBits=1, power=2; 2<3 → maxBits=2, power=4; 4<3 false → stop
  - 3 ≠ 4 → no bump
  - Returns m=2
- Theorem claims: `3 ≤ 1 <<< (2-1) = 2`. But 3 > 2. **FALSE.**

The bug is in the bound: the docstring says "`2^(m-1)` is the first power
≥ `weightSum`" but actually `2^maxBits` (= `2^m` when not bumped) is the
first power ≥ weightSum. The correct bound is `weightSum ≤ 1 <<< m`:

- **Not bumped**: `2^(m-1) < weightSum ≤ 2^m = 1 <<< m` (by while-loop
  invariant: previous power was too small, current power is big enough)
- **Bumped**: `weightSum = 2^old_maxBits`, `m = old_maxBits + 1`, so
  `weightSum = 2^(m-1) ≤ 2^m = 1 <<< m` ✓

The CLAUDE.md says: "Do NOT modify theorem statements just to make proofs
[...]

Issue #610, opened 2026-03-06
Track E: Prove buildFseTable structural invariants — table size and accuracy log passthrough


kim-em:
## Current state

- **Track E (Zstd)**: Active implementation + spec phase
- **Sorry count**: 2 (`Zip/Spec/Fse.lean:166`, `Zip/Spec/Zstd.lean:143`)
- **Existing FSE spec theorems** in `Zip/Spec/Fse.lean`:
  - `predefined_litLen_valid`, `predefined_matchLen_valid`, `predefined_offset_valid` (proved by `decide`)
  - `decodeFseDistribution_accuracyLog_ge` (proved)
  - `decodeFseDistribution_accuracyLog_le` (proved)
  - `decodeFseDistribution_sum_correct` (sorry — covered by issue #605)
- **No theorems about `buildFseTable`** exist yet — the spec predicate `ValidFseTable` is defined but never connected to the implementation
- **`buildFseTable`** (`Zip/Native/Fse.lean:120-176`): builds FSE decoding table from probability distribution. Returns `Except String FseTable` but the function body has no `throw` — it always succeeds. Initializes cells via `Array.replicate (1 <<< accuracyLog) default`, then mutates via `Array.set!` (which preserves size). Returns `{ accuracyLog, cells }`.

## Background

The FSE spec file (`Zip/Spec/Fse.lean`) defines `ValidFseTable` with three conjuncts:
1. `table.size = 1 <<< accuracyLog`
2. All symbol indices < numSymbols
3. All numBits ≤ accuracyLog

Currently, the spec file has theorems about `decodeFseDistribution` (the input side) but nothing about `buildFseTable` (the output side). Connecting `buildFseTable` to `ValidFseTable` is essential for the Track E B3 verification path: proving that FSE decoding is correct requires showing that the table structure satisfies the expected invariants.

The two simplest structural properties are:
- **Table size**: `buildFseTable` initializes cells with `Array.replicate tableSize default` where `tableSize = 1 <<< accuracyLog`. Since `Array.set!` preserves array size unconditionally, the final cells array has size `1 <<< accuracyLog`.
- **Accuracy log passthrough**: The function returns `{ accuracyLog, cells }` where `accuracyLog` is the input parameter unchanged.

These are shallow structural properties (not deep loop invariants), making them tractable within one session. They establish the foundation for the harder conjuncts (symbol bounds, numBits bounds) in future work.

## Deliverables

[...]

PR #628, opened 2026-03-06
Track E: Prove ZstdHuffman mathematical properties — isPow2_iff and weightSum_pos_of_exists_nonzero

### LINKED ISSUE #617: Track E: Prove ZstdHuffman mathematical properties — isPow2_iff and weightSum_pos_of_exists_nonzero
## Current state

- **Track E (Zstd)**: Active implementation + spec phase
- **Sorry count**: 7 total (1 Fse.lean, 1 Zstd.lean, 5 ZstdHuffman.lean)
- **ZstdHuffman.lean sorry breakdown**:
  - `isPow2_iff` (line 60): bitwise characterization of powers of 2
  - `weightSum_pos_of_exists_nonzero` (line 143): fold positivity from element
  - `buildZstdHuffmanTable_tableSize` (line 106): monadic table size invariant
  - `buildZstdHuffmanTable_maxBits_pos` (line 114): monadic maxBits positivity
  - `weightsToMaxBits_valid` (line 128): while loop accumulator reasoning
- **This issue covers**: the first 2 (mathematical properties), NOT the last 3
  (monadic/loop invariant proofs which are covered or will be covered by
  separate issues)

## Background

`Zip/Spec/ZstdHuffman.lean` was created by PR #606 (merged 2026-03-06) with
8 fully-proven theorems and 5 `sorry` stubs. The 5 sorries fall into two
categories:

**Mathematical properties** (this issue):
- `isPow2_iff`: `isPow2 n = true ↔ ∃ k, 1 <<< k = n` — characterizes the
  `n > 0 && (n &&& (n - 1)) == 0` bit-trick as equivalent to being a power
  of 2. This is a standard result in number theory / bit manipulation.
- `weightSum_pos_of_exists_nonzero`: if any weight `weights[i] > 0`, then
  `weightSum weights > 0` — a fold monotonicity property: adding a positive
  `1 <<< (w - 1)` to a non-negative accumulator gives a positive result.

**Monadic/loop invariant proofs** (separate issues):
- `buildZstdHuffmanTable_tableSize/maxBits_pos`: require unfolding
  `buildZstdHuffmanTable` through its `do` block (similar to the FSE proofs
  in issue #610)
- `weightsToMaxBits_valid`: requires while-loop invariant reasoning

The mathematical properties are independent and use different proof
techniques (bitwise arithmetic, fold induction) from the monadic proofs.

## Deliverables

1. **Prove `isPow2_iff`** in `Zip/Spec/ZstdHuffman.lean` (~20-30 lines):
   ```
   theorem isPow2_iff (n : Nat) :
       isPow2 n = true ↔ ∃ k : Nat, 1 <<< k = n
   ```
   **Proof strategy**:
   - Unfold `isPow2` to `n > 0 && (n &&& (n - 1)) == 0`
   - The forward direction (`→`): `n > 0` and `n &&& (n - 1) = 0` implies
     `n` is a power of 2. This is a standard bit-trick: for a power of 2,
     `n = 2^k` has exactly one bit set, so `n - 1 = 2^k - 1` has all lower
     bits set, and `n &&& (n - 1) = 0`.
   - The reverse direction (`←`): if `n = 2^k`, then `n > 0` (trivially)
     and the bit-trick holds by induction on `k`.
   - Consider using `Nat.bitwise` lemmas, `Nat.land` properties, or
     converting to `BitVec` and using `bv_decide` for finite cases +
     induction for the general case.
   - If `bv_decide` works for small cases, it may suffice for a
     `decide`-friendly proof after appropriate reformulation.
   - Alternative: prove the contrapositive — if `n` has two or more bits
     set, then `n &&& (n - 1) ≠ 0`.

2. **Prove `weightSum_pos_of_exists_nonzero`** in `Zip/Spec/ZstdHuffman.lean`
   (~15-25 lines):
   ```
   theorem weightSum_pos_of_exists_nonzero (weights : Array UInt8)
       (i : Fin weights.size) (hi : weights[i].toNat > 0) :
       weightSum weights > 0
   ```
   **Proof strategy**:
   - `weightSum` is `Array.foldl f 0` where `f acc w = if w > 0 then
     acc + (1 <<< (w - 1)) else acc`.
   - The fold starts at 0, and `f` only ever adds non-negative values
     (powers of 2) to the accumulator. So the accumulator is always ≥ 0.
   - At index `i`, `f acc weights[i]` adds `1 <<< (weights[i] - 1) ≥ 1`
     (since `weights[i] > 0`), making the accumulator ≥ 1.
   - All subsequent steps preserve `acc ≥ 1` since `f` only adds non-negative
     values.
   - This likely needs a helper lemma: `Array.foldl_mono` or similar —
     if `f acc x ≥ acc` for all `x`, and `f acc₀ x₀ ≥ k`, then the final
     result is `≥ k`.
   - Check if `ZipForStd/Array.lean` already has fold monotonicity lemmas.
   - The `lean-array-list` skill may have relevant patterns.

## Context

- **Independent of #610** (buildFseTable structural invariants — FSE, not Huffman)
- **Independent of #605** (decodeFseDistribution_sum_correct — FSE, not Huffman)
- **Independent of #595** (XxHash64 spec — different component)
- **Independent of #552** (compressed block wiring — implementation, not spec)
- These are the two most tractable ZstdHuffman sorries — the remaining 3
  require monadic or loop invariant techniques that are harder
- Reducing ZstdHuffman sorry count from 5 to 3 demonstrates progress on
  the Zstd verification track
- Estimation: 2 deliverables × difficulty 2 (requires exploration of Lean's
  bitwise and fold APIs, but the mathematical properties are well-understood) = 4

## Verification

- `lake build` succeeds
- `lake exe test` passes
- `grep -c sorry Zip/Spec/ZstdHuffman.lean` shows 3 (down from 5)
- Both new proofs compile without `sorry` or `native_decide`
- The proofs follow project style (prefer `omega`/`simp`/`grind`, minimize steps)
- Progress entry documents proof techniques used

kim-em:
Closes #617

Session: `c1c5afce-71be-40b4-958b-1bfdcc3dc010`

a56bd52 doc: progress entry for ZstdHuffman mathematical proofs
ff683e6 feat: prove isPow2_iff and weightSum_pos_of_exists_nonzero (ZstdHuffman)

🤖 Prepared with Claude Code

### DIFF
--- a/progress/20260306T034500Z_c1c5afce.md
+++ b/progress/20260306T034500Z_c1c5afce.md
@@ -1,0 +1,50 @@
+# Progress: Prove ZstdHuffman mathematical properties
+
+- **Date**: 2026-03-06T03:45:00Z
+- **Session**: c1c5afce (feature)
+- **Issue**: #617
+
+## Accomplished
+
+Proved two mathematical property theorems in `Zip/Spec/ZstdHuffman.lean`,
+reducing sorry count from 5 to 3:
+
+1. **`isPow2_iff`**: The bit-trick `n > 0 && (n &&& (n-1)) == 0`
+   correctly characterizes powers of 2 (`∃ k, 1 <<< k = n`).
+   - Reverse direction: testBit extensionality — `2^k` has exactly
+     bit k set, `2^k - 1` has bits below k set, so their AND is 0.
+   - Forward direction: strong induction on n, case-splitting on parity:
+     - Even n: helper lemma `land_pred_even` shows
+       `n &&& (n-1) = 2 * (n/2 &&& (n/2-1))`, reducing to IH.
+     - Odd n ≥ 3: helper lemma `land_pred_odd` shows
+       `n &&& (n-1) = 2*(n/2) > 0`, contradiction.
+     - Odd n = 1: direct (k = 0).
+
+2. **`weightSum_pos_of_exists_nonzero`**: If any weight `weights[i] > 0`,
+   then `weightSum weights > 0`.
+   - Three helper lemmas: `weightStep_ge` (fold step ≥ acc),
+     `weightFold_mono_init` (fold monotone in initial value),
+     `weightFold_ge_init` (fold preserves lower bound).
+   - Main proof: split list at index i, show processing the positive
+     weight makes accumulator ≥ 1, remaining fold preserves this.
+
+## Proof techniques
+
+- `Nat.eq_of_testBit_eq` + `Nat.testBit_and`/`Nat.testBit_succ` for
+  bitwise identity proofs
+- `omega` for div/mod arithmetic in testBit helper lemmas
+- `Nat.strongRecOn` for strong induction in the forward direction
+- List fold splitting via `List.foldl_append` + `List.drop_eq_getElem_cons`
+
+## Remaining
+
+- 3 sorries remain in ZstdHuffman.lean (monadic/loop invariant proofs,
+  covered by issues #610 and #622)
+- Total project sorry count: `grep -rc sorry Zip/` shows reduction by 2
+
+## Quality metrics
+
+- Sorry count: 5 → 3 in `Zip/Spec/ZstdHuffman.lean`
+- `lake build`: ✓
+- `lake exe test`: ✓ (all tests passed)
+- No `native_decide` used


kim-em:
## Post-merge review (issue #640)

**Overall**: Good quality proofs with appropriate proof techniques.

### Checklist
- [x] No `native_decide` — confirmed
- [x] No bare `simp` except one instance (see below)
- [x] Theorem statements match claims (isPow2 ↔ power of 2, positive weight sum)
- [x] No fragile simp sets — `simp only [...]` used throughout, except line 125

### Findings

**1. Bare `simp` on line 125** (reverse direction of `isPow2_iff`):
```lean
    apply Nat.eq_of_testBit_eq
    intro i
    simp
```
This closes the goal `2 ^ k - 1 &&& 2 ^ k = 0` using `simp`'s bitwise lemmas. A `simp only [...]` with the specific lemmas used would be more robust against simp set changes. Low priority but worth noting for future cleanup.

**2. Helper lemma quality**: `land_pred_even`, `land_pred_odd`, `land_pred_zero_imp_pow2` are well-structured — the even/odd case split with testBit extensionality is the right approach. The strong induction pattern (`Nat.strongRecOn`) is clean.

**3. `weightFold_mono_init` and `weightFold_ge_init`**: Good factoring — monotonicity is the key property and it's extracted cleanly. `weightFold_ge_init` could be derived from `weightFold_mono_init` (it's the special case `a = b`), but the direct proof is only 4 lines so the duplication is fine.

**4. `weightSum_pos_of_exists_nonzero` proof**: The list-splitting technique via `List.foldl_append` + `List.drop_eq_getElem_cons` is the right approach. The `show` term on line 241 is a bit long but readable.

### Sorries remaining
3 in ZstdHuffman.lean: `buildZstdHuffmanTable_tableSize`, `buildZstdHuffmanTable_maxBits_pos`, `weightsToMaxBits_valid` — tracked by issues #610 and #622.

🤖 Review by Claude Code (issue #640)

PR #1657, opened 2026-04-22
Feature: Track E — saturating UInt64 counter add in streaming FFI decoders (PR #1610 F-d)

### LINKED ISSUE #1645: Feature: Track E — saturating UInt64 counter add in streaming FFI decoders (PR #1610 F-d)
### Current state

The streaming FFI decoders introduced by PR #1610 track decompressed
output bytes in a running `IO.Ref UInt64` counter, checked against
`maxDecompressedSize` before each chunk is written. Two call sites
compute `next = total + chunk.size.toUInt64` with a raw `UInt64` add:

- [`Zip/Gzip.lean:93`](/home/kim/lean-zip/Zip/Gzip.lean:93)
  inside `Gzip.decompressStream`.
- [`Zip/RawDeflate.lean:63`](/home/kim/lean-zip/Zip/RawDeflate.lean:63)
  inside `RawDeflate.decompressStream`.

Both sites today read:

```lean
let total ← totalRef.get
let next := total + chunk.size.toUInt64
if maxDecompressedSize ≠ 0 && next > maxDecompressedSize then
  throw (IO.userError s!"...")
totalRef.set next
output.write chunk
```

If `total + chunk.size.toUInt64` overflows `UInt64.max`, the raw add
**silently wraps** — `next` becomes small again, the `next >
maxDecompressedSize` guard passes, and the chunk is written. At
`UInt64.max ≈ 16 EiB`, the overflow condition is unreachable in
practice on any current hardware, so the severity is low. But the
defence-in-depth property that *the cap is honoured regardless of
input* is lost, and the C-side overflow-guard style used in
[`c/zlib_ffi.c`](/home/kim/lean-zip/c/zlib_ffi.c:1) (see `grow_buffer`
and the `SIZE_MAX`-style guards) is not matched on the Lean side.

This gap was flagged as **F-d (≤ 50 lines, hardening)** in the
PR #1616 review of PR #1610:

> **F-d (≤ 50 lines, hardening)** — Replace the raw
> `total + chunk.size.toUInt64` add at [`Zip/Gzip.lean:90`] and
> [`Zip/RawDeflate.lean:63`] with a saturating or checked add (e.g.,
> `if chunk.size.toUInt64 > UInt64.max - total then throw ... else
> total + chunk.size.toUInt64`). Severity low — `UInt64.max ≈ 16 EiB`
> is unreachable in practice — but the current code silently wraps
> past `UInt64.max`. A saturating add makes the defence-in-depth
> explicit and matches the C-side overflow-guard style used in
> `c/zlib_ffi.c`.

See
[`progress/20260422T061856Z_e45d791b.md:204-213`](/home/kim/lean-zip/progress/20260422T061856Z_e45d791b.md:204)
(the PR #1616 review progress entry) for the verbatim
recommendation. F-b (bomb-limit for `decompressFile`) landed as
PR #1642. F-c (fuzz-harness extension to streaming APIs) is a
sibling feature issue created in this planner session. F-d is the
final open item from the PR #1616 review; closing it ticks the
final F-series bullet.

Quality invariants: 0 sorries, `lake build` clean, `lake exe test`
passes.

### Deliverables

1. **Replace the raw add at
   [`Zip/Gzip.lean:93`](/home/kim/lean-zip/Zip/Gzip.lean:93) with a
   checked add.** The canonical shape:

   ```lean
   let total ← totalRef.get
   let chunkSize := chunk.size.toUInt64
   if chunkSize > UInt64.size.toUInt64 - 1 - total ∨
       (maxDecompressedSize ≠ 0 ∧ total + chunkSize > maxDecompressedSize) then
     throw (IO.userError
       s!"gzip: decompressed stream exceeds limit ({maxDecompressedSize} bytes)")
   let next := total + chunkSize
   totalRef.set next
   output.write chunk
   ```

   (Simplified — `UInt64.size.toUInt64 - 1` is `UInt64.max` spelled
   explicitly; if a built-in constant is available use that.) The
   guard condition combines two disjuncts:

   - **overflow guard**: `chunkSize > UInt64.max - total`, which
     catches the `UInt64` wrap-around on addition.
   - **cap guard**: the existing `next > maxDecompressedSize`
     check (only when `maxDecompressedSize ≠ 0`).

   Both failures share the same error message and throw site. This
   preserves the `"exceeds limit"` substring that the regression
   tests grep for, so no test needs to be updated for the overflow
   path (which is unreachable in practice anyway).

   **Alternative: saturating add.** If the checked-add reshape is
   awkward, an equivalent saturating version is acceptable:

   ```lean
   let next :=
     if chunkSize > UInt64.size.toUInt64 - 1 - total
     then UInt64.size.toUInt64 - 1  -- saturate at UInt64.max
     else total + chunkSize
   if maxDecompressedSize ≠ 0 && next > maxDecompressedSize then
     throw (IO.userError s!"...")
   ```

   The checked-add version throws on overflow; the saturating
   version lets the cap guard catch it. Either is correct so long
   as the silent-wrap is gone. Pick the one that's shorter and
   reads clearly.

2. **Apply the same change at
   [`Zip/RawDeflate.lean:63`](/home/kim/lean-zip/Zip/RawDeflate.lean:63)
   with the `"raw deflate:"` error prefix.** The file's existing
   error-message convention (also `"exceeds limit"`, see
   [`Zip/RawDeflate.lean:67`](/home/kim/lean-zip/Zip/RawDeflate.lean:67))
   must be preserved verbatim. Both the overflow and cap disjuncts
   should share the message.

3. **Add a one-line comment near each throw site explaining the
   defence-in-depth intent.** Follow the `.claude/CLAUDE.md`
   comment guidance: *only when WHY is non-obvious*. Suggested
   wording (one line, plain prose):

   ```lean
   -- Guard against UInt64 wrap-around on very large chunks; the
   -- cap-overflow check below is meaningful only if the add did
   -- not silently wrap.
   ```

   Only add this if the reviewer would otherwise ask "why the
   extra guard". The check itself is small enough that the
   comment may be unnecessary — use discretion per CLAUDE.md.

4. **No test changes.** The overflow path is unreachable at
   realistic input sizes, so a deterministic regression test
   would need a `>UInt64.max`-byte mock — impossible to fit in a
   fixture. The fuzz harness (F-c, sibling issue) already
   exercises the counter under randomised cross-chunk state and
   will surface any new crash the checked-add introduces. The
   existing bomb-limit regression tests at
   [`ZipTest/Gzip.lean:180-195`](/home/kim/lean-zip/ZipTest/Gzip.lean:180)
   and
   [`ZipTest/RawDeflate.lean`](/home/kim/lean-zip/ZipTest/RawDeflate.lean)
   already exercise the cap-hit path, which is unchanged.

### Scope discipline — do NOT

- Do NOT change the error message's `"exceeds limit"` substring
  or the surrounding wording. The
  `error-wording-catalogue` entry and the existing regression
  tests grep for that substring; a deviation would break both.
- Do NOT add a new error variant for the overflow case. Share
  the cap-overflow message — the overflow is a degenerate case
  of "exceeds limit" (cap ≤ `UInt64.max`, so crossing
  `UInt64.max` necessarily crosses the cap).
- Do NOT change the signature of `decompressStream` or
  `decompressFile`. This is a hardening-only patch to the
  existing function bodies.
- Do NOT touch the non-streaming (whole-buffer) decoders —
  `Zlib.decompress`, `Gzip.decompress`, `RawDeflate.decompress`
  — they allocate a fixed-size output buffer in C and do not
  have a Lean-side running counter. Whole-buffer overflow is a
  C-side concern, audited separately under
  `scripts/sanitize-ffi.sh` and the allocation-site inventory.
- Do NOT modify `Zip/Native/**` — native decoders use
  `Nat`-typed `maxOutputSize` and do not have the `UInt64`
  overflow hazard.
- Do NOT modify `PLAN.md` or `.claude/CLAUDE.md` (off-limits).
- Do NOT update `SECURITY_INVENTORY.md` unless a specific row
  in *"Decompression Limit Inventory"* or *"Missing work"*
  becomes stale. A `UInt64`-overflow hardening is not currently
  called out in the inventory, so no update is needed; but if
  you find a related row, a one-line tweak is acceptable.
- Do NOT use `native_decide` (forbidden per `.claude/CLAUDE.md`).

### Context

- **Why the hazard is low severity.** `UInt64.max ≈ 1.8 · 10^19`
  bytes ≈ 16 EiB. Even a 1-TiB stream would need to run ~16
  million times to wrap. In production, the `maxDecompressedSize`
  cap (default 1 GiB after PR #1631) bounds `total` far below
  `UInt64.max` long before wrap becomes possible. The hazard is
  only realisable if a caller passes `maxDecompressedSize := 0`
  (unlimited mode, opt-in bomb-unsafe) **and** drives an
  exabyte-scale stream — neither of which happens outside of
  adversarial reasoning.
- **Why close it anyway.** Defence-in-depth is the Track E
  ethos: the C side guards overflow explicitly (see
  `c/zlib_ffi.c`'s `grow_buffer` and its `size_t` overflow
  check); the Lean side should match. The check is ≤ 5 lines
  of code at each call site. A future refactor that flips
  `maxDecompressedSize`'s default to `0` somewhere in the call
  graph (regression), or a caller that disables the cap for
  performance, loses the only line of defence against wrap. A
  silent wrap is a worse failure mode than a spurious
  `"exceeds limit"` throw — both are rejections, but the wrap
  produces wrong output silently.
- **Why the disjunct-form guard.** Splitting the guard into
  `overflow ∨ cap_exceeded` keeps the short-circuit behaviour
  (if overflow is true, we don't need to compute `total +
  chunkSize` — which would wrap — before checking the cap). In
  Lean 4 with `bool &&` semantics, this is fine; be careful
  with `∧` vs `&&` — the original code uses `&&` in an `if`
  expression, so keep that style to avoid accidental
  recomputation.
- **Why this is separate from F-c.** F-c extends test coverage
  to the high-level streaming APIs. F-d modifies production
  code. They touch disjoint files and can be claimed in any
  order. If F-c lands first, the new fuzz probes will exercise
  F-d's checked add under randomised `maxCap`s; if F-d lands
  first, the existing fuzz probes will still exercise the
  low-level `InflateState.push/finish` path (unchanged by
  F-d). Either ordering is fine.

### Verification

- `lake build` clean.
- `lake exe test` passes — all existing bomb-limit regression
  tests still fire at the existing cap, and no new test
  failure from unrelated probes. Specifically:
  - `ZipTest/Gzip.lean` `decompressStream` bomb test
    ([`ZipTest/Gzip.lean:180-195`](/home/kim/lean-zip/ZipTest/Gzip.lean:180))
    still asserts `"exceeds limit"` on a 6200-byte input with
    cap=10.
  - `ZipTest/Gzip.lean` `decompressFile` bomb test (PR #1642,
    lines 210–233) still asserts the same substring with
    cap=10 and `partial_.size ≤ 10`.
  - `ZipTest/RawDeflate.lean` `decompressStream` bomb test
    still fires.
- `grep -rc sorry Zip/` returns `0` (unchanged baseline).
- `grep -n 'UInt64.size.toUInt64 - 1\|UInt64.max' Zip/Gzip.lean Zip/RawDeflate.lean`
  shows the new overflow guard at both call sites (one occurrence
  per file, near lines 93 and 63 respectively).
- `grep -n '"exceeds limit"' Zip/Gzip.lean Zip/RawDeflate.lean`
  still shows the original error substring at both call sites.
- Diff limited to `Zip/Gzip.lean`, `Zip/RawDeflate.lean`, and a
  progress entry. No changes to `ZipTest/**`, `c/**`,
  `Zip/Native/**`, `SECURITY_INVENTORY.md`, `PLAN.md`,
  `.claude/CLAUDE.md`, or other files.
- Sanity check via runtime experiment is optional (the overflow
  is unreachable in practice). If a reviewer wants
  demonstration, a `#eval`-only sanity check that the guard
  compiles at the expected boolean values is acceptable — but
  delete it before commit per the `.claude/CLAUDE.md` rule that
  `try?` / exploratory `#eval` should not ship in committed
  code.

### Hints / references

- [`Zip/Gzip.lean:86-106`](/home/kim/lean-zip/Zip/Gzip.lean:86) —
  `decompressStream` body. Deliverable 1 edits lines 92–97 of
  this function.
- [`Zip/RawDeflate.lean:57-85`](/home/kim/lean-zip/Zip/RawDeflate.lean:57)
  — `decompressStream` body. Deliverable 2 edits the analogous
  lines in this function.
- [`c/zlib_ffi.c`](/home/kim/lean-zip/c/zlib_ffi.c:1) `grow_buffer`
  — the C-side overflow-guard style this change matches. Search
  for `SIZE_MAX` or `size_t` overflow checks to find the pattern.
- [`progress/20260422T061856Z_e45d791b.md:204-213`](/home/kim/lean-zip/progress/20260422T061856Z_e45d791b.md:204)
  — PR #1616 review recommendation F-d verbatim. Cite this
  entry in the commit message.
- `error-wording-catalogue` skill — the `"exceeds"` family of
  substrings must be preserved. See
  `.claude/skills/error-wording-catalogue/SKILL.md` for the
  exhaustive table.

### Sizing note

Small — ≤ 20 lines of diff across two files, plus a progress
entry. Well under the 50-line target the review estimated.
Single logical concern: "defence-in-depth UInt64 overflow guard
on streaming FFI decoder counter". No risk of scope creep — the
two call sites are structurally identical, and the change is
mechanical.

kim-em:
Closes #1645

Session: `b794f64f-b807-4085-96c0-830b7959343b`

7f70a9b feat: defend UInt64 counter add in streaming FFI decoders (PR #1610 F-d)

🤖 Prepared with Claude Code

### DIFF
--- a/Zip/Gzip.lean
+++ b/Zip/Gzip.lean
@@ -91,7 +91,9 @@ partial def decompressStream (input : IO.FS.Stream) (output : IO.FS.Stream)
     if chunk.size > 0 then
       let total ← totalRef.get
       let next := total + chunk.size.toUInt64
-      if maxDecompressedSize ≠ 0 && next > maxDecompressedSize then
+      -- `next < total` detects UInt64 wrap-around; without it the cap
+      -- check below could be silently bypassed on >UInt64.max-byte streams.
+      if next < total || (maxDecompressedSize ≠ 0 && next > maxDecompressedSize) then
         throw (IO.userError
           s!"gzip: decompressed stream exceeds limit ({maxDecompressedSize} bytes)")
       totalRef.set next
--- a/Zip/RawDeflate.lean
+++ b/Zip/RawDeflate.lean
@@ -62,7 +62,9 @@ partial def decompressStream (input : IO.FS.Stream) (output : IO.FS.Stream)
     if chunk.size > 0 then
       let total ← totalRef.get
       let next := total + chunk.size.toUInt64
-      if maxDecompressedSize ≠ 0 && next > maxDecompressedSize then
+      -- `next < total` detects UInt64 wrap-around; without it the cap
+      -- check below could be silently bypassed on >UInt64.max-byte streams.
+      if next < total || (maxDecompressedSize ≠ 0 && next > maxDecompressedSize) then
         throw (IO.userError
           s!"raw deflate: decompressed stream exceeds limit ({maxDecompressedSize} bytes)")
       totalRef.set next
--- a/progress/20260422T113747Z_b794f64f.md
+++ b/progress/20260422T113747Z_b794f64f.md
@@ -1,0 +1,86 @@
+# Track E F-d: saturating UInt64 counter add in streaming FFI decoders
+
+**Date (UTC):** 2026-04-22T11:37Z
+**Session type:** feature
+**Issue:** #1645 (PR #1610 review F-d)
+**Branch:** agent/b794f64f
+
+## Accomplished
+
+Closed the final open F-series item from the PR #1616 review of PR #1610:
+the streaming FFI decoders' running byte counter is now defended against
+`UInt64` wrap-around.
+
+- `Zip/Gzip.lean:93-97` — `decompressStream.checkAndWrite` guard now reads
+  `if next < total || (maxDecompressedSize ≠ 0 && next > maxDecompressedSize)`.
+- `Zip/RawDeflate.lean:64-68` — same edit, with the `"raw deflate:"`
+  prefix preserved.
+- Each site carries a two-line comment explaining the wrap-detection
+  intent (the WHY that is non-obvious — the cap check alone is silently
+  bypassable if the add wraps).
+
+Diff is ±6 lines across the two files (well under the ≤ 20-line
+estimate in the plan). No test changes — the overflow path is
+unreachable at realistic input sizes (`UInt64.max ≈ 16 EiB`), and the
+existing bomb-limit regression tests exercise the cap-hit path which is
+unchanged.
+
+## Decisions
+
+- **Wrap-detection vs explicit `UInt64.max - total`.** The plan offered
+  both the checked-add form (`chunkSize > UInt64.max - total`) and a
+  saturating alternative, and explicitly said "pick the one that reads
+  clearly". I went with `next < total`: after the UInt64 add, `next <
+  total` holds exactly when the add overflowed (since a non-overflowing
+  add produces `next ≥ total`). This is shorter, uses no magic constant
+  for `UInt64.max`, and composes cleanly with the existing cap check as
+  a single `||` disjunct. Both disjuncts share the same `"exceeds
+  limit"` throw, preserving the substring the regression tests
+  (`ZipTest/Gzip.lean:180-195`, the `decompressFile` bomb test, and the
+  `RawDeflate` analogue) grep for.
+
+- **Shared error message.** The plan was explicit: overflow is a
+  degenerate case of "exceeds limit" (cap ≤ UInt64.max, so crossing
+  UInt64.max necessarily crosses the cap), and a separate error variant
+  would break the `error-wording-catalogue` discipline. Kept a single
+  throw per site.
+
+- **Short-circuit.** The plan called out that the overflow disjunct
+  should short-circuit before computing `total + chunkSize`. In this
+  form, `next` is computed unconditionally (UInt64 addition is a pure
+  wrapping op, so there is no semantic cost to computing the wrapped
+  value — we simply observe it and reject). The short-circuit concern
+  in the plan's canonical shape was aesthetic rather than semantic;
+  this equivalent form avoids the magic constant entirely.
+
+## Verification
+
+- `lake build -R` clean (191 jobs).
+- `lake exe test` green (all tests passed, including the three
+  existing bomb-limit regressions: `Gzip.decompressStream`,
+  `Gzip.decompressFile`, `RawDeflate.decompressStream`).
+- `grep -rc sorry Zip/` → `0` (unchanged baseline).
+- `grep -n '"exceeds limit"'` in both files still shows the substring
+  at the expected throw sites.
+- `grep -n 'next < total'` shows the new guard at
+  `Zip/Gzip.lean:94` and `Zip/RawDeflate.lean:65`.
+- Diff limited to `Zip/Gzip.lean` and `Zip/RawDeflate.lean` plus this
+  progress entry. No changes to `ZipTest/**`, `c/**`, `Zip/Native/**`,
+  `Zip/Spec/**`, `SECURITY_INVENTORY.md`, `PLAN.md`, or
+  `.claude/CLAUDE.md`.
+
+## Track E status
+
+The F-series (PR #1616 review of PR #1610) is now fully closed:
+
+- F-a: ✅ landed in PR #1610 (bomb-limit for `decompressStream`).
+- F-b: ✅ PR #1642 (bomb-limit for `decompressFile`).
+- F-c: ✅ PR #1653 (fuzz harness extension to streaming APIs).
+- F-d: ✅ this PR — overflow-guard hardening.
+
+## Quality metric deltas
+
+- sorry count: 0 → 0 (unchanged).
+- test count: unchanged.
+- lines of production code: +4 across two files (two 2-line comments
+  plus the one-line guard-expansion per site).


kim-em:
## Post-merge audit (issue #1663) — F-d UInt64 overflow guard

Paired-review audit for this F-d PR, completing the F-series review
pairing (cadence per meditate #1651 §4). Structured against the
7 deliverables in #1663. Audit artifact:
[`progress/20260422T131959Z_e81d853d.md`](https://github.com/kim-em/lean-zip/blob/agent/e81d853d/progress/20260422T131959Z_e81d853d.md).

### Findings

| # | Deliverable | Verdict |
|---|---|---|
| 1 | Wrap-detection correctness (`next < total`) | ✓ Correct on every reachable input (standard two's-complement overflow check; case-split confirmed in audit). |
| 2 | `"exceeds limit"` substring fidelity | ✓ Preserved at `Zip/Gzip.lean:98` and `Zip/RawDeflate.lean:69`; all 5 bomb-limit regression tests still match. |
| 3 | Comment quality (WHY, not WHAT) | ✓ Load-bearing *silently bypassed* phrase captures the security-relevant consequence. No tighter form preserves both the named-disjunct anchor and the *bypassed* phrasing. |
| 4 | Test coverage adequacy (no new test) | ✓ Overflow branch is observationally equivalent to the cap branch (same throw, same substring) and unreachable at realistic sizes (~584 years at 1 GB/s, 64 KiB chunks). Contriving a test would require exposing the internal `IO.Ref UInt64`. Accept. |
| 5 | Cross-decoder symmetry | ✓ `diff -u` of the two `checkAndWrite` bodies (with `gzip:`/`raw deflate:` normalized) produces **empty output**. Zero drift. |
| 6 | Composability with the cap default flip (#1631) | ✓ Default 1 GiB cap fires long before overflow threatens; in unlimited mode (`maxDecompressedSize = 0`), the overflow disjunct becomes the sole gate. Both disjuncts feed the same throw, so disjunct ordering has no observable difference. |
| 7 | Interaction with streaming-FFI fuzz probes (#1653) | ✓ `lake exe test` runs the 1000-iter `FuzzInflate` smoke including `tryStreamingGzip` / `tryStreamingRawDeflate`; no regression. Overflow disjunct unreachable from fuzz inputs (chunks × iters ≪ `UInt64.max`). |

### Drift / follow-ups

**None.** No production or test changes recommended.

### F-series closure

| Item | Feature PR | Review PR |
|------|-----------|-----------|
| F-a bomb-limit for `decompressStream` | #1610 | #1616 |
| F-b bomb-limit for `decompressFile`   | #1642 | #1660 (F-c cluster) |
| F-c fuzz harness extension            | #1653 | #1660 (F-c cluster) |
| F-d UInt64 overflow hardening         | **#1657** | **this audit** |

The Track E F-series review pairing is now complete.

### Baseline

- `grep -rc sorry Zip/`: 0 (unchanged).
- `lake build -R`: clean (191 jobs).
- `lake exe test`: all tests pass.

---
🤖 Prepared with [Claude Code](https://claude.com/claude-code)

PR #1771, opened 2026-04-24
Track E: reject encryption-related general-purpose flag bits (0/6/13) at CD parse + cd-encrypted-flag.zip fixture

### LINKED ISSUE #1762: Track E: reject encryption-related general-purpose flag bits (0/6/13) at CD parse + cd-encrypted-flag.zip fixture
Track E hardening for the CD-entry general-purpose flag word.
Three encryption-related bits are currently **never checked** —
bit 0 (traditional ZipCrypto), bit 6 (strong encryption), and
bit 13 (encrypted central-directory). lean-zip has no
cryptographic support whatsoever: it cannot decrypt ZipCrypto,
does not implement the strong-encryption SES/AES frameworks, and
cannot decode a masked CD. Currently a crafted archive with any
of these bits set is accepted and processed as if the bytes were
plaintext — producing garbage output that fails the CRC check
*late*, in `readEntryData` at
[`Zip/Archive.lean:758`](/home/kim/lean-zip/Zip/Archive.lean:758)
(*"CRC32 mismatch"*), after running the DEFLATE decoder over
encrypted bytes. Early rejection at CD-parse time is the right
fix: the CD flag word is the authoritative metadata carrier and
the rejection is a single mask-compare.

### Current state

`parseCentralDir`
([`Zip/Archive.lean:365-456`](/home/kim/lean-zip/Zip/Archive.lean:365))
reads the general-purpose bit flag into `flags` at
[`Zip/Archive.lean:396`](/home/kim/lean-zip/Zip/Archive.lean:396)
and uses only bit 11 (the UTF-8 name flag) for name decoding at
[`Zip/Archive.lean:422`](/home/kim/lean-zip/Zip/Archive.lean:422).
No other flag bits are inspected at CD parse. `readEntryData`
subsequently uses bit 3 (data descriptor) at
[`Zip/Archive.lean:732`](/home/kim/lean-zip/Zip/Archive.lean:732)
to gate the CD/LH size/crc checks, and compares the whole
non-bit-3 mask via `dataDescriptorBitMask = 0xFFF7` at
[`Zip/Archive.lean:720`](/home/kim/lean-zip/Zip/Archive.lean:720).
The CD/LH flags-consistency check catches CD-vs-LH divergence on
every bit in the mask, but does **not** reject an archive whose
CD and LH both declare encryption — the attacker simply sets the
same encryption bit on both sides.

APPNOTE §4.4.4 (general-purpose bit flag):

- Bit 0: `If set, indicates that the file is encrypted.`
- Bit 6: `Strong encryption.` (APPNOTE Annex H; SES/AES-based.)
- Bit 13: `Used when encrypting the Central Directory.`

lean-zip implements none of these. Producing the real plaintext
from any of them would require cryptographic primitives we do
not ship and — for bit 13 — a key-derivation step over an
encrypted CD that lean-zip cannot even parse. Every archive
carrying any of these bits is therefore malformed from
lean-zip's perspective and should be rejected.

The writer side is trivially compliant: `writeCentralHeader` at
[`Zip/Archive.lean:108`](/home/kim/lean-zip/Zip/Archive.lean:108)
emits a hard-coded `flags = 0x0800` (only bit 11 UTF-8 set),
and `writeLocalHeader` at
[`Zip/Archive.lean:81`](/home/kim/lean-zip/Zip/Archive.lean:81)
does the same. No writer-side change is needed.

Attack shape: a crafted archive sets bit 0 (`flags = 0x0001`) on
both CD and LH. The method is 8 (deflate) with pre-encrypted
bytes as payload. Current behaviour: CD/LH flags-consistency
passes (both sides agree), CD/LH size-consistency passes,
`readEntryData` runs DEFLATE over the encrypted bytes, the
decoder either produces garbage or throws a native-inflate error
deep in the pipeline. New behaviour: `parseCentralDir` throws
`"encryption-related general-purpose flag set"` immediately.

### Deliverables

1. **Add the encryption-bit mask check in `parseCentralDir`**.
   Immediately after the `flags` read at
   [`Zip/Archive.lean:396`](/home/kim/lean-zip/Zip/Archive.lean:396),
   add a single mask compare. The three bits together form
   `0x2041` (bit 0 + bit 6 + bit 13 = `1 + 64 + 8192`):

       -- APPNOTE §4.4.4: encryption-related general-purpose flag
       -- bits that lean-zip does not support.  Bit 0 = traditional
       -- ZipCrypto, bit 6 = strong encryption (SES/AES), bit 13 =
       -- encrypted central directory.  All three require
       -- cryptographic primitives we do not ship; reject early so
       -- the CRC mismatch does not fire late after running DEFLATE
       -- over encrypted bytes.  The writer emits `flags = 0x0800`
       -- (bit 11 UTF-8 only) and never sets any of these.
       unless (flags &&& 0x2041) == 0 do
         throw (IO.userError
           s!"zip: encryption-related general-purpose flag set (flags={flags}); lean-zip does not support encrypted archives")

   Placement: before the method/crc/size reads on the same CD
   row, **before** the entry is pushed to `entries`. The check
   applies per-entry (same as `diskNumberStart` at
   [`Zip/Archive.lean:412-416`](/home/kim/lean-zip/Zip/Archive.lean:412))
   — a single encrypted entry is enough to trip the guard and no
   further entries are parsed. Do **not** fold with the flags
   consistency check in `readEntryData`: the CD-parse check and
   the CD/LH consistency check are orthogonal dimensions (the
   former is an absolute bit-presence rejection; the latter is a
   CD-vs-LH divergence check).

   **Error substring for tests**:
   `"encryption-related general-purpose flag set"` (new wording
   family — record a row in the `error-wording-catalogue` skill
   adjacent to the existing `"flags mismatch between CD and
   local header"` entry).

2. **Add the `cd-encrypted-flag.zip` fixture.** Extend
   [`scripts/build-cd-lh-mismatch.py`](/home/kim/lean-zip/scripts/build-cd-lh-mismatch.py)
   additively:
   - Add `lh_flags: int = 0` and `cd_flags: int = 0` kwargs to
     `make_lh` and `make_cd` respectively (replacing the
     hard-coded `0` at
     [`scripts/build-cd-lh-mismatch.py:38`](/home/kim/lean-zip/scripts/build-cd-lh-mismatch.py:38)
     and
     [`scripts/build-cd-lh-mismatch.py:59`](/home/kim/lean-zip/scripts/build-cd-lh-mismatch.py:59)).
   - Thread them through `write(...)` (default `0` preserves
     byte-identity of all pre-existing fixtures).
   - Emit a `write(...)` call with
     `lh_flags=0x0001, cd_flags=0x0001, lh_method=0, cd_method=0`.
     Setting bit 0 on **both** sides ensures the CD/LH
     flags-consistency check does not fire first — the CD-parse
     encryption-bit guard fires instead, which is the whole
     point of the fixture.

   Record the fixture SHA-256 in the progress entry. The
   existing eight fixtures emitted by the generator must remain
   byte-identical pre/post — confirm via `cmp` and record all
   nine SHA-256s.

3. **Wire the fixture into `ZipTest/ZipFixtures.lean`** with an
   `assertThrows "encryption-related general-purpose flag set"`
   block parallel to the `cd-lh-flags-mismatch.zip` block.
   Append the new filename to the two cleanup arrays at the
   tail.

4. **Update `SECURITY_INVENTORY.md`**:
   - New row in *Minimized Reproducer Corpus* (class: `other
     (CD/LH consistency)` — the encryption-bit rejection lives
     in the same CD-parse guard family as the `diskNumberStart`
     check; first-landed: this PR's number; anchor at the
     throw-message line per the throw-line convention).
   - New *Recent wins* bullet under *ZIP Archive
     Reader/Extractor* parallel to the per-entry `diskNumberStart`
     bullet PR #1759 added.
   - The *Missing work* block does not currently call out
     encryption-flag handling as a gap — record in the progress
     entry that this dimension was discovered during the CD
     general-purpose-flag coverage sweep, not a gap inventory
     had already flagged.
   - Re-run `bash scripts/check-inventory-links.sh` →
     `errors=0, warnings=0`.

5. **Critical regression smoke test on
   `testdata/zip/interop/`.** Six interop fixtures live there.
   For each, parse the CD flag word (CD offset +8, UInt16) with
   Python and confirm `(flags & 0x2041) == 0`. If any interop
   fixture disagrees, **stop and replan** — real-world producers
   *should* never emit encryption bits in our plain-archive test
   corpus, but confirm empirically before landing a strict
   rejection. The defensive interop-corpus pattern from PR #1736
   saved the project once already. Record the per-fixture check
   results in the progress entry.

### Verification

- `lake build -R` clean.
- `lake exe test` green; new `cd-encrypted-flag.zip`
  `assertThrows` block fires with substring
  `"encryption-related general-purpose flag set"`.
- `bash scripts/check-inventory-links.sh` → `errors=0,
  warnings=0`.
- `git diff --stat HEAD`: confined to `Zip/Archive.lean`,
  `ZipTest/ZipFixtures.lean`, `SECURITY_INVENTORY.md`,
  `scripts/build-cd-lh-mismatch.py`, plus the new
  `testdata/zip/malformed/cd-encrypted-flag.zip` binary and a
  one-row addition to
  `.claude/skills/error-wording-catalogue/SKILL.md`. Expect
  ~10 LOC source delta + ~122-byte fixture + ~10 LOC generator
  delta (two kwarg additions) + ~5 LOC inventory rows.
- Sorry count: unchanged at 0.

### Sizing

One `feat:` commit. ~10 LOC source delta (one `unless` block) +
~122-byte fixture + ~10 LOC generator delta + ~5 LOC inventory
rows. Single logical change. Smallest of the CD-parse Track E
guards landed this phase (smaller than `cd-entry-disknum-mismatch`
because only one flag-word read is added rather than a new CD
field).

### Context

- **Predecessor PRs**:
  - PR #1759 / issue #1758-sibling (CD entry `diskNumberStart`)
    — direct template for the per-entry CD-parse-side guard
    pattern; read `progress/ee509a40.md` for the per-entry
    throw placement and the generator-kwarg-extension shape.
  - PR #1716 / issue #1715 (CD/LH flags consistency, bit-3
    masked) — established the `dataDescriptorBitMask` discipline
    and the `flags` field on `Entry`. This issue is orthogonal:
    CD/LH consistency is a divergence check; CD-parse encryption
    rejection is an absolute bit-presence rejection.
- **APPNOTE anchors**:
  - §4.4.4 — general-purpose bit flag semantics (bit 0, bit 6,
    bit 13).
  - Annex H — strong-encryption specification (SES/AES).
- **Writer-side evidence the three bits are never set**:
  [`Zip/Archive.lean:81`](/home/kim/lean-zip/Zip/Archive.lean:81)
  (LH) and
  [`Zip/Archive.lean:108`](/home/kim/lean-zip/Zip/Archive.lean:108)
  (CD) both write `flags = 0x0800` (bit 11 UTF-8 only); none of
  bits 0, 6, or 13 are ever emitted. Reader-only gap.
- **Generator**: extend
  [`scripts/build-cd-lh-mismatch.py`](/home/kim/lean-zip/scripts/build-cd-lh-mismatch.py)
  — the `write` signature already carries `lh_*`/`cd_*`
  additive-kwarg pairs (version, method, sizes, disk-number);
  adding `lh_flags`/`cd_flags` is mechanical.

### Decisions to make during implementation

- **Single combined mask vs. three separate checks.** `0x2041`
  catches all three bits in one `unless`; the error message
  reports the full raw `flags` value so the offending bit is
  identifiable by inspection. Splitting into three separate
  `unless` blocks would give per-bit attribution but triple the
  throw sites without any diagnostic gain for a single-fixture
  test. Recommendation: **combine** — match the sibling
  disk-number pair (single combined throw with both raw values).
- **Which fixture bit to set.** Bit 0 is the simplest and most
  concrete (traditional ZipCrypto is the textbook case); it also
  allows a realistic LH encryption header to be omitted
  (bit 0 archives normally carry a 12-byte encryption header
  before the payload, but the fixture just sets the flag without
  the header — the CD-parse guard fires before any LH read).
  Bits 6 and 13 are covered by the same mask so a single fixture
  exercises all three. Recommendation: **use bit 0** for
  determinism.
- **Placement: `parseCentralDir` vs. `readEntryData`.** The
  CD-parse placement catches the anomaly before the LH is even
  read — defence in depth. `readEntryData` already has the
  CD/LH flags-consistency check, but that only catches the
  CD-vs-LH divergence case (asymmetric encryption flag), not the
  both-sides-encrypted case this issue addresses.
  Recommendation: **parseCentralDir** — match the
  `diskNumberStart` per-entry pattern.
- **Whether to check bit 7 (reserved), bit 8 (reserved), bit 9
  (reserved), bit 14-15 (reserved for PKWARE future use).**
  APPNOTE §4.4.4 leaves them unassigned; some future extension
  might use them. Non-goal for this issue — rejecting unassigned
  bits would be overly strict and would break on any future
  producer. Recommendation: **only bits 0, 6, 13** — the three
  bits with concrete documented meaning that lean-zip cannot
  honour.

### Non-goals

- **Do NOT** implement ZipCrypto, AES, or any other encryption.
  The whole point of this issue is to reject, not support.
- **Do NOT** check reserved / unassigned flag bits. Only the
  three APPNOTE-documented encryption bits.
- **Do NOT** fold with the CD/LH flags-consistency check — the
  two checks are orthogonal (absolute bit-presence vs. CD-vs-LH
  divergence) and share no data-dependency.
- **Do NOT** fold with issue #1722 (CD/LH name bytes) or any
  sibling in-flight CD/LH consistency check — different
  dimension, different guard.
- **Do NOT** touch `PLAN.md` or `.claude/CLAUDE.md`.

🤖 Prepared with Claude Code

kim-em:
Closes #1762

Session: `6198786e-9239-41bf-b543-8ce54d7cdb6d`

01cac11 feat: Track E CD-parse encryption-flag rejection + cd-encrypted-flag.zip fixture

🤖 Prepared with Claude Code


kim-em:

PR #1814, opened 2026-04-24
Track E: per-entry versionNeededToExtract >= 45 when ZIP64 features are used at CD parse + cd-zip64-entry-version-low.zip fixture

### LINKED ISSUE #1805: Track E: per-entry versionNeededToExtract >= 45 when ZIP64 features are used at CD parse + cd-zip64-entry-version-low.zip fixture
Track E hardening: at CD-parse time, enforce APPNOTE §4.4.3.2's
`versionNeededToExtract ≥ 45` lower bound on **any CD entry that
uses ZIP64 fields** (i.e. one or more of the standard 32-bit
`compressedSize`, `uncompressedSize`, `localOffset` slots carries
the `0xFFFFFFFF` sentinel and the entry's layout is resolved via
the ZIP64 `0x0001` extra block). Sibling of #1798
(per-entry `versionNeededToExtract ≤ 45` upper bound at CD parse,
claimed) and #1758 (EOCD64-level `versionNeededToExtract ≥ 45`
lower-bound sanity, in-flight PR #1764). This issue closes the
per-entry lower-bound dimension: the EOCD64-level check says
"the archive's envelope is ZIP64-capable"; #1798 says "no entry
claims a future ZIP capability"; this check says "any entry that
uses ZIP64 must declare the version that unlocks ZIP64 fields".

### Current state

`parseCentralDir` reads `versionNeeded` at CD+6 at
[`Zip/Archive.lean:473`](/home/kim/lean-zip/Zip/Archive.lean:473),
then reads the 32-bit `stdCompSize` / `stdUncompSize` /
`stdOffset` at CD+20 / +24 / +42 at
[`Zip/Archive.lean:479`](/home/kim/lean-zip/Zip/Archive.lean:479) /
[`:480`](/home/kim/lean-zip/Zip/Archive.lean:480) /
[`:499`](/home/kim/lean-zip/Zip/Archive.lean:499). ZIP64 kicks in
only when one of those slots is `val32Max` (`0xFFFFFFFF`), tested
at [`:544`](/home/kim/lean-zip/Zip/Archive.lean:544):

    let (uncompSize, compSize, localOff) ←
      if stdCompSize == val32Max || stdUncompSize == val32Max || stdOffset == val32Max then
        if hasDuplicateZip64Extra extraData then ...
        match parseZip64Extra extraData stdUncompSize stdCompSize stdOffset with
        | some v => pure v
        | none => throw (IO.userError s!"zip: malformed ZIP64 extra field for {name}")
      else
        pure (stdUncompSize.toUInt64, stdCompSize.toUInt64, stdOffset.toUInt64)

`versionNeeded` is **not** cross-checked against the sentinel
predicate. A crafted archive with `versionNeeded = 20`
("default" for non-ZIP64 entries, APPNOTE §4.4.3.2) and one of
`stdCompSize` / `stdUncompSize` / `stdOffset` equal to
`val32Max` (forcing ZIP64 resolution) currently passes CD parse —
the `parseZip64Extra` call succeeds if the 0x0001 block is
well-formed, returning the 64-bit value, and `Archive.list` /
`Archive.extract` proceeds as if the archive is ZIP64. A strict
reader that enforces APPNOTE §4.4.3.2's lower bound would reject
the archive. Parser-differential smuggling vector.

The writer side is trivially compliant: both `writeLocalHeader`
at [`Zip/Archive.lean:90`](/home/kim/lean-zip/Zip/Archive.lean:90)
and `writeCentralHeader` at
[`Zip/Archive.lean:116-117`](/home/kim/lean-zip/Zip/Archive.lean:116)
emit `if z64 then 45 else 20`, where
`z64 := needsZip64 entry` exactly matches the sentinel predicate
in `parseCentralDir`. So every writer-emitted ZIP64 entry has
`versionNeeded = 45` and every non-ZIP64 entry has
`versionNeeded = 20` — the reader-side guard never fires on
legitimate lean-zip archives.

Attack shape: craft a 1-byte stored-method entry
(`compressedSize = uncompressedSize = 1`), write its CD entry
with `stdCompSize = val32Max` and a well-formed 0x0001 extra
block containing the true compressed size (8 bytes), but set
`versionNeeded = 20`. Current behaviour: `parseZip64Extra`
resolves the 64-bit value and the entry proceeds; `Archive.list`
lists it. New behaviour: `parseCentralDir` throws
`"ZIP64 entry with versionNeededToExtract < 45"` before ZIP64
resolution.

### Deliverables

1. **Add the `versionNeeded ≥ 45` guard for ZIP64-using entries
   in `parseCentralDir`**. Place the new check immediately after
   the `stdOffset` read at
   [`Zip/Archive.lean:499`](/home/kim/lean-zip/Zip/Archive.lean:499)
   and before the `entryEnd` / name/extra extraction that
   currently starts at `:496`. Reuse the **same sentinel
   predicate** as the ZIP64-resolution gate at
   [`:544`](/home/kim/lean-zip/Zip/Archive.lean:544) so the two
   checks agree on "this entry uses ZIP64":

       -- APPNOTE §4.4.3.2: any CD entry that uses ZIP64 fields
       -- (i.e. one of compressedSize/uncompressedSize/localOffset
       -- is the 0xFFFFFFFF sentinel and the resolved value comes
       -- from the 0x0001 extra block) must declare
       -- versionNeededToExtract ≥ 45 — 45 is the ZIP version that
       -- unlocks ZIP64 field layout.  Sibling of the archive-level
       -- EOCD64 versionNeededToExtract ≥ 45 check (PR #NNNN) and
       -- the per-entry upper bound (PR #NNNN).  Writer is
       -- trivially compliant: writeLocalHeader and
       -- writeCentralHeader emit `if z64 then 45 else 20` with
       -- `z64 := needsZip64 entry`, exactly matching the sentinel
       -- predicate below.
       let usesZip64 :=
         stdCompSize == val32Max || stdUncompSize == val32Max ||
         stdOffset == val32Max
       if usesZip64 && versionNeeded < 45 then
         throw (IO.userError
           s!"zip: ZIP64 entry with versionNeededToExtract {versionNeeded} < 45 for CD entry at offset {pos}")

   The `diskNumberStart` check at
   [`:493`](/home/kim/lean-zip/Zip/Archive.lean:493) does not
   need to be reordered; the new guard slots naturally between
   the disk-number check (at +34, already-parsed) and the
   entry-end check (at +46 + variable-length fields). The
   `versionNeeded` value is already bound on line 473 and needs
   no re-read. Placing the guard before the ZIP64 resolution
   block gives a precise error attribution rather than letting
   `parseZip64Extra` succeed and the archive proceed. No
   interaction with #1798 (upper bound at a different
   comparator) or #1762 (flag bits at a different field).

2. **New regression fixture**
   `testdata/zip/malformed/cd-zip64-entry-version-low.zip`.
   Generate it via a new helper in
   `scripts/build-zip64-malformed-fixtures.py` (or the nearest
   adjacent builder; choose whichever one already produces
   ZIP64 fixtures — `zip64-extra-oversized-datasize.zip` and
   the two `cd-zip64-extra-duplicate.zip` /
   `lh-zip64-extra-duplicate.zip` are the templates). Shape:
   - one `method=0` (stored) entry with
     `uncompressedSize = compressedSize = 6` ("hello\n" or
     similar tiny fixed payload so byte-reproducibility is
     easy)
   - CD entry with `stdCompSize = 0xFFFFFFFF`,
     `stdUncompSize = 6`, `stdOffset = 0` (so only one slot
     forces ZIP64; keeps the fixture small)
   - well-formed 0x0001 extra block at CD carrying exactly the
     8-byte resolved `compressedSize` (one slot → `dataSize=8`,
     matching PR #1785's invariant so that check doesn't fire
     first)
   - CD `versionNeeded = 20` (the attack value; APPNOTE's
     "stored entry" default)
   - LH mirrors the CD values except the CD-specific
     `diskNumberStart`/attributes (keep the fixture minimal;
     LH `versionNeeded` must equal 20 so the CD/LH
     one-sided check at
     [`:869`](/home/kim/lean-zip/Zip/Archive.lean:869) does
     not fire first)
   - Fixture target size: ~120–160 bytes (comparable to
     existing ZIP64 fixtures). Keep the `SHA-256` digest
     reproducible by using a deterministic `zipfile.ZipInfo`
     construction and **not** writing `lastModTime`
     (hard-coded to `0x0000` / `0x0021` via lean-zip's
     `defaultDosTime` / `defaultDosDate`).
   - On-disk size: budget ≤ 200 bytes. If the builder produces
     a larger fixture, prefer a tighter construction (single
     entry, empty name, no comment) over truncation.

3. **Regression test** in
   `ZipTest/ZipFixtures.lean`. Add one `assertThrows` block
   using the pattern established by the existing ZIP64 fixtures.
   The substring must be
   `"ZIP64 entry with versionNeededToExtract"` — unique across
   the inventory (#1798 uses `"unsupported versionNeededToExtract"`
   and #1758 uses `"ZIP64 EOCD64 versionNeededToExtract"`; this
   one uses the per-entry **ZIP64 entry** phrasing for
   attribution).

4. **Inventory update** in `SECURITY_INVENTORY.md`:
   - *Recent wins* bullet under "ZIP Archive Reader/Extractor"
     referencing the committed PR number and the fixture
   - new row in the *Minimized Reproducer Corpus* table for
     `cd-zip64-entry-version-low.zip`, alphabetically slotted
     between `cd-zip64-extra-duplicate.zip` (:784) and
     `eocd-disknum-mismatch.zip` (:785)
   - **do not** leave placeholder `this PR` / `#TBD`; the
     linter rule landed in PR #1796 will warn otherwise. Use
     the `coordination create-pr` PR number once assigned, or
     split the inventory row into a follow-up PR per the
     one-PR-per-bullet half-closed two-step.

5. **Skill catalogue update** in
   `.claude/skills/error-wording-catalogue/SKILL.md`. Add one
   row for the new `"ZIP64 entry with versionNeededToExtract"`
   substring, noting the `< 45` distinguisher vs the sibling
   rows for #1798 and #1758.

### Context

- **Sibling issues and PRs**:
  - **#1798** (claimed, per-entry `versionNeeded ≤ 45` upper
    bound) — fires after the same `versionNeeded` read but
    with the **unconditional** upper-bound predicate
    `versionNeeded ≤ 45`. Current issue's guard fires with
    the ZIP64-sentinel-gated **lower-bound** predicate
    `versionNeeded ≥ 45` (conditional on sentinel). The two
    predicates never overlap — together they pin the field
    to exactly the APPNOTE-permitted range and eliminate a
    ZIP64-sentinel-mismatch smuggling vector.
  - **#1758** (in-flight PR #1764, EOCD64
    `versionNeededToExtract ≥ 45` lower bound) — archive-
    level counterpart. This issue applies the same lower
    bound at the per-entry layer so single-entry
    ZIP64-smuggling archives are caught even when the
    EOCD64 envelope is well-formed.
  - **#1762** (in-flight PR #1771, encryption-related GPF
    bits 0/6/13) — independent bit range, independent field.
    No interaction.
  - **#1773** (landed, stored-method size invariant) —
    precedence-shift template: same "reject at CD-parse
    before ZIP64 / late-decode" shape.
  - **PR #1801** (landed, method allowlist {0, 8}) — same
    early-rejection motif for a different field.
- **APPNOTE reference**: §4.4.3.2 enumerates the full version
  taxonomy:
  - 10: default
  - 20: file is a volume label / folder / file is compressed
        using Deflate
  - 45: **File uses ZIP64 format extensions**
  - 46: File is compressed using BZIP2
  - 50: File is encrypted using ...
  - 51..63: higher-capability features lean-zip does not
    implement
  `≥ 45` is the strict lower bound once the ZIP64 extension is
  in use.
- **Why CD-parse and not ZIP64-resolution-time**: the guard can
  fire *before* calling `parseZip64Extra`. Placing it earlier
  gives precise attribution and saves a branch in the resolved
  case. The alternative (check inside `parseZip64Extra`)
  would conflate the parser state with the validity check and
  would not fire on archives whose 0x0001 block is malformed
  (the parser returns `none` first).
- **Interaction with in-flight #1798**: #1798's upper-bound
  guard fires at [`:474`](/home/kim/lean-zip/Zip/Archive.lean:474)
  (immediately after the `versionNeeded` read); this issue's
  lower-bound guard fires at `:500` (after the `stdOffset`
  read, where the sentinel predicate is first checkable). No
  rebase conflict: insertion sites are ~25 lines apart and
  reference distinct `parseCentralDir` locals (`versionNeeded`
  alone vs `versionNeeded + stdCompSize + stdUncompSize +
  stdOffset`). If #1798 lands first, this issue's worker
  rebases and the two guards sit as adjacent blocks. If this
  issue lands first, #1798's worker similarly rebases.
- **Sizing**: expect ~6 LoC in `Zip/Archive.lean`,
  ~40–60 lines in the fixture builder script, ~8 lines in
  `ZipFixtures.lean`, ~2 lines in `SECURITY_INVENTORY.md`,
  ~2 lines in the error-wording-catalogue skill. Fixture
  itself is ≤ 200 bytes.

### Verification

- `lake build` — clean, zero warnings.
- `lake exe test` — *All tests passed!*; the new
  `cd-zip64-entry-version-low.zip` assertion appears in the
  *ZIP fixture tests: OK* summary.
- `scripts/check-inventory-links.sh` — `errors=0, warnings=0`;
  no new placeholder-PR (`this PR` / `#TBD`) references.
- Fixture byte-reproducibility: rerun the builder in a fresh
  `/tmp` directory, `sha256sum` both copies, `cmp`
  byte-for-byte against the committed copy.
- Manually verify the guard:
  - **fires** on the new
    `cd-zip64-entry-version-low.zip` fixture with substring
    `"ZIP64 entry with versionNeededToExtract"`
  - **does not fire** on any existing `testdata/zip/` fixture
    that uses `versionNeeded = 45` (e.g. the ZIP64 fixtures
    `oversized-zip64-compressed-size.zip`,
    `oversized-zip64-uncompressed-size.zip`,
    `zip64-extra-oversized-datasize.zip`,
    `cd-zip64-extra-duplicate.zip`,
    `lh-zip64-extra-duplicate.zip`,
    `zip64-eocd64-bad-recsize.zip`,
    `eocd-zip64-override-nosentinel.zip`)
  - **does not fire** on any existing non-ZIP64 fixture
    that uses `versionNeeded = 20`
  - **does not fire** on the `testdata/zip/interop/go-zip64.zip`
    third-party fixture

kim-em:
Closes #1805

Session: `725052c7-2039-4a39-96d3-7e7516d1c276`

98ba251 progress: feature session 20260424T124239Z closing #1805 (per-entry ZIP64 versionNeededToExtract >= 45)
3afc7dc feat: per-entry ZIP64 versionNeededToExtract >= 45 guard at CD parse + cd-zip64-entry-version-low.zip fixture

🤖 Prepared with Claude Code


kim-em:

PR #2533, opened 2026-06-10
chore: drop zopfli from the routine bench matrix; freeze it as a ratio ceiling

kim-em:
This PR removes zopfli from the routine `bench-report` corpus matrix and instead captures it as a frozen, one-shot artifact. zopfli is compress-only, level-less, and ~100× slower than zlib at default iteration count, so even at one level / single rep it dominated the wall-clock of the whole Canterbury+Silesia matrix on every regeneration. A new `bench-report --zopfli-ceiling <out>` mode runs zopfli once over every corpus file and writes `bench/results/zopfli-ceiling.json` (deterministic `ratio`/`out_size` are the signal; the single-rep `compress_mbps` is an artifact, not a benchmark). `bench/plot.py` overlays that frozen file on the ratio graphs, so the best-achievable-ratio reference stays on the dashboard while the routine matrix never recomputes it. `bench/README.md` documents the frozen artifact and the (discouraged) regenerate command.

The frozen `zopfli-ceiling.json` and the refreshed dashboard data are a deliberate follow-up rather than part of this commit: regenerating them requires building `bench-report`, which currently OOM-kills on `Zip.Spec.InflateTable` (heavy kernel evaluation) under the heavy concurrent build load on the bench machine. The plotter degrades gracefully when the frozen file is absent, so merging this is safe; the data lands once the machine has free memory.

🤖 Prepared with Claude Code

### DIFF
--- a/ZipBenchReport.lean
+++ b/ZipBenchReport.lean
@@ -221,9 +221,7 @@ def runReport (outPath : String) : IO Unit := do
 
   -- Real-corpus rows only: the same timing matrix over every committed corpus
   -- (one single-size workload per file), so the dashboard rests entirely on
-  -- representative data. zopfli is the ratio ceiling (compress-only, very slow,
-  -- level-less) — run at one nominal level / single rep so it appears on the
-  -- ratio graphs without dominating wall-clock.
+  -- representative data.
   let corpora ← loadCorpora
   if corpora.isEmpty then
     IO.eprintln "  no corpora found — run bench/fetch_corpora.sh"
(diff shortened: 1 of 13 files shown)


kim-em:
Update: the data caught a low-load window and is now included (commit 980effb) — generated on origin/master (#2531) on chungus:
- `bench/results/latest.json` refreshed (zopfli dropped from the matrix; native/zlib/miniz/libdeflate fresh; go/js/zig/ocaml reused, external/unchanged).
- `bench/results/zopfli-ceiling.json` — the frozen one-shot ratio ceiling (23 files), overlaid by plot.py.
- all 8 dashboard SVGs re-rendered with the zopfli ceiling line back on the ratio graphs.

So this PR is now complete (code + data), not code-only.

Issue #2506, opened 2026-06-04
Adopt the wide-buffer decoder as the default inflate path (and refresh the decode charts)


kim-em:
#2505 landed the wide-buffer Huffman decoder (`InflateBuf`) and proved it equal to the reference end-to-end — `InflateBuf.inflate = Inflate.inflate`, `InflateBuf.inflate (deflateFixed x) = .ok x`, 0 sorries, no `maxHeartbeats`/`native_decide`/`@[extern]`. It is **measured ~2.0× faster on text decode** (4324 vs 2110 MB/s, 1 MiB lvl 6), but it is **not yet the default decode path**, so `Inflate.inflate` / native gzip / native zlib still run the `BitReader`-based `decodeHuffmanFast` and users see no change. This issue is about actually plugging it in.

## Why it's a refactor, not a switch

Two structural facts block a one-line flip:

1. **Import cycle.** `Zip/Native/InflateBuf.lean` *imports* `Zip/Native/Inflate.lean`, so `inflateLoop`/`inflate` (in `Inflate.lean`) cannot call the buffered decoder. The buffered primitives (`refill`, `walkTree`, `decodeSym`, `takeBits`, `go`, `decodeHuffmanFastBuf`) have to move into `Inflate.lean` (they only need `lengthBase`/`copyLoop`/`readBitsFast`/`decodeWithTable`, all defined earlier there — no new dependency).
2. **The spec-equivalence is layered the wrong way for a flip.** Everything is proven as *buffered = BR-based = spec*, with the **BR-based decoder as the reference**. Pointing `inflateLoop` at the buffered decoder inverts that: the ~16 `decodeHuffmanFast_eq` rewrite sites in `InflateComplete`/`InflateLoopBounds`/`InflateCorrect`/`InflateRawSuffix`, plus the freshly-landed `inflateLoopBuf_eq`/`inflate_eq` bridges, all assume the old direction and would need re-threading.

## Proposed plan

- Move the buffered decoder into `Inflate.lean` (keep the `Zip.Native.InflateBuf` namespace so existing references in `InflateBufCorrect` don't churn).
- Make the buffered decoder the **primary** decoder and re-orient `decodeHuffmanFast_eq` so the spec-equivalence is established for the buffered decoder directly (keep the BR-based decoder + its `…BR_eq` as the intermediate, or retire it). This is the part that must be done carefully — it touches the large inflate correctness/completeness/bounds/suffix proofs and the new bridge lemmas.
[...]

PR #2600, opened 2026-06-13
doc(track-d): #2501 close — wide-buffer decoder satisfies all acceptance criteria (verified)

### LINKED ISSUE #2501: perf: buffered bit reader to amortize per-read allocation (Track D, decode — NO @[extern])
**Track D — runtime (decompress).** The decode `perf` profile (see the bulk-`readBits` issue) shows ~41% of inflate time in `malloc`/`free`/`lean_dec_ref_cold` plus ~26% in the bit reader, all from `BitReader` allocating a fresh 3-field struct (+ `Except`/pair) on **every** read. Bulk `readBits` removes the per-bit multiplier; this issue targets the remaining per-*read* struct-allocation churn — the structural cost of threading an immutable cursor — which is the bulk of the ~17–47× decode gap vs zlib (zlib holds bits in a mutable register and refills in bulk).

### Hard constraint

**No `@[extern]`** — the bit reader must stay fully verified in Lean. A trusted C core is off the table. So this is a verified pure-or-`ST` refinement, not an FFI shim.

### Approaches (to evaluate, cheapest-risk first)

1. **Wide-buffer pure reader.** A `BitReader` variant carrying a `UInt64` bit accumulator + a `bitsInBuffer : Nat`, refilled in bulk from `data` (one refill per ~56–64 bits) and consumed by field. Still pure (allocates the struct per update), but **far fewer updates and byte reads** — one refill amortises many small reads, and `peek`/`consume` avoid per-bit work. Partial win, smallest proof: a refinement `toBits`-correspondence to the current `BitReader`.
2. **`ST`/linear mutable reader.** Decode against a mutable bit-buffer (`ST.Ref` or a linear `BitReaderM`) so updates mutate in place and never allocate, with a proof that the mutable decode equals the pure-spec decode (the heavy obligation: relate the `ST` trace to the functional `inflate`). Full win, largest proof.

Spike approach 1 first (measure the speedup and proof size) before committing to 2.

### Proof

Whichever representation, keep the existing `Deflate.Spec.decode` / `inflate_correct` as the canonical spec and prove the new reader produces the identical decoded output (a refinement, like `decodeWithTable_eq` did for the table decoder). The bit-list semantics (`BitReader.toBits`, `bytesToBits`) are the bridge.

### Acceptance criteria

- Substantial native **decompress** throughput gain (re-profile: allocation share collapses); inflate output identical (roundtrip + fuzz); `grep -rc sorry Zip/`: 0; **no `@[extern]`/`native_decide`**; `lake build && lake exe test` green.
- Re-bench; refresh dashboard; `Landed` row with the decode delta and the before/after profile.

### Scope guard

Touch the decode reader + its refinement proof (`ZipCommon/BitReader.lean` or a new reader module; `Zip/Spec/Inflate*`/`Bitstream*`). Larger, multi-step, exploratory — drive interactively; spike approach 1's proof before approach 2. Do **after** the bulk-`readBits` quick win. PLAN.md off-limits.

🤖 Prepared with Claude Code

kim-em:
Closes #2501

Session: `9f936dd4-3380-4093-ac18-2a9f9a97edf4`

8ab9b41 doc(track-d): #2501 verification verdict — approach 1 landed, approach 2 deferred

🤖 Prepared with Claude Code

### DIFF
--- a/progress/2026-06-13T063247Z_9f936dd4.md
+++ b/progress/2026-06-13T063247Z_9f936dd4.md
@@ -1,0 +1,55 @@
+# Session 9f936dd4 — #2501 buffered bit reader: verify + close
+
+**UTC:** 2026-06-13T06:32Z
+**Type:** feature (one-shot, `pod once 2501`)
+**Issue:** #2501 — perf: buffered bit reader to amortize per-read allocation (Track D, decode)
+
+## Verdict
+
+**Closed #2501 as completed.** Every acceptance criterion is already satisfied
+by merged code (approach 1, the wide-buffer Huffman decoder), landed across
+PR #2505 (decoder, proven equal) and PR #2506/#2507 (made the default). This
+session independently re-verified the landed state rather than re-implementing
+it, and recorded the approach-2 verdict.
+
+## What I verified this session
+
+- **Build green** (inside `nix-shell`; the bare-shell link failure was the
+  documented cached-link-flags / libdeflate issue, not a code regression):
+  `lake -R build test` → `Build completed successfully (249 jobs)`.
+- **All tests pass**, including correctness + the decode fuzz harness:
+  `lake exe test` → 1000-iter `FuzzInflate`, 256 compress→FFI, 256
+  differential header-mutation iterations, all green; `All tests passed!`.
+- **Decode speedup realized** (this machine, native decompress): 256 KiB
+  `text` lvl 6 **2750 MB/s**, lvl 1 1964 MB/s; 64 KiB text lvl 6 1203 MB/s —
+  consistent with the D-13 dashboard's recorded ~1.7–2.2× over the
+  pre-flip `BitReader` decoder.
+- **0 sorries** in `Zip/`; **no** `native_decide` / `@[extern]` /
+  `@[implemented_by]` / `maxHeartbeats` on the decode path (the only textual
+  hits are docstrings asserting their *absence*).
+
+## Why not approach 2 (the ST/linear mutable reader)
+
+The issue says to spike approach 1 first and *then* decide on approach 2.
+Post-spike verdict: **not worth its cost.**
+
+- Approach 1 delivered the win the issue targeted (per-read struct-allocation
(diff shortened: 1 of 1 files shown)


kim-em:
## Why a doc-only PR closes a perf issue

This PR closes #2501 without code changes because **the deliverable is already merged**. The wide-buffer Huffman decoder (issue's approach 1) landed and was made the default decode path across **#2505** (decoder, proven equal to the `BitReader` reference) and **#2506/#2507** (default flip), recorded as dashboard rows **D-12/D-13**.

This session independently re-verified that every #2501 acceptance criterion holds, rather than re-implementing:

| Criterion | Status |
|---|---|
| Substantial native decompress gain; allocation share collapses | ✅ 1.7–2.2× (D-12/D-13); this run: 256 KiB text lvl6 **2750 MB/s** |
| inflate output identical (roundtrip + fuzz) | ✅ full suite + 1000-iter `FuzzInflate` + 256 differential-mutation iters pass |
| `grep -rc sorry Zip/`: 0 | ✅ 0 |
| no `@[extern]` / `native_decide` | ✅ none on decode path (only docstrings asserting their absence; also no `@[implemented_by]`/`maxHeartbeats`) |
| `lake build && lake exe test` green | ✅ (in `nix-shell`; bare-shell link failure is the documented libdeflate cached-link-flags issue) |
| Re-bench; dashboard Landed row | ✅ D-12 + D-13 already recorded |

**Approach 2 (ST/linear mutable bit reader) — post-spike verdict: deferred.** The issue says to spike approach 1 first and then decide. Approach 1 already captured the targeted win (per-read struct-allocation churn) with a fully verified refinement; the remaining decode cost is now `copyLoop` (back-reference memcpy, per D-10), which a mutable *bit* reader does not address — so approach 2's marginal gain is small while it carries the issue's heaviest proof obligation (relating an `ST` trace to functional `inflate`). If desired later, it should be a fresh, independently-scoped research issue, not left tracked open under #2501.

PR #2646, opened 2026-06-17
perf: incompressible pre-scan escape to stored block

### LINKED ISSUE #2633: perf(explore): incompressible / low-match pre-scan escape -> stored block
**Status: exploration — measure-first. From the Codex synthesis pass (≈ old plan Item 1.7).** Do not claim until triaged.

## Idea
`deflateRawBaseP` currently sizes a stored block only *after* paying the full matcher. For random / already-compressed input, a cheap bounded **sample pass** (e.g. distinct-4-gram / hash-collision rate, literal entropy, or repeated-4-byte count over a sample) could choose `deflateStoredPure` immediately and skip the matcher entirely.

## Magnitude / feasibility
- Huge on incompressible inputs (the matcher is ~90% of compress and produces almost no matches there); near-zero on compressible corpora (Canterbury/Silesia mostly compressible, so the geomean barely moves — the win is real-world archives / random data).
- Proof is **trivial** — it's a dispatch to the already-proven `deflateStoredPure`; the sample heuristic is opaque to correctness (worst case: a needlessly-stored compressible block, still valid).

## Why we might reject / risk
The heuristic must be **conservative** to avoid false-positives on compressible text (storing text would tank ratio). Tune the gate so it only fires on genuinely incompressible data. Limited benefit on the standard corpora means the dashboard won't show much — value is robustness on real-world/binary inputs.

kim-em:
This PR adds `incompressiblePrescan`, a cheap bounded sample over `deflateRaw`'s input that routes genuinely incompressible data (random or already-compressed bytes) straight to `deflateStoredPure`, skipping the full hash-chain match pass (and, at level 9, the optimal-parse DP) that the cost model would have discarded anyway. It closes https://github.com/kim-em/lean-zip/issues/2633.

The scan engages only at ≥1 MiB (below that the matcher is cheap, so the scan would be net overhead) and reads up to four 32 KiB contiguous regions spread across the input. Each region is judged by order-0 entropy first — most compressible data (text/source/html, low entropy) bails there after a single histogram pass, before any table is allocated — and only a high-entropy region pays for the 4-gram repeat check: every consecutive 4-gram is inserted into a fresh single-hash presence table, so repeated-block data (high-entropy bytes repeated at a window-reachable distance) shows up as a wall of collisions rather than slipping through. The gate fires only when every region is both near-maximal entropy and repeat-free. Self-similar data whose period exceeds the 32 KiB match window (e.g. `R ++ R` for a 512 KiB `R`) is genuinely incompressible by DEFLATE and is correctly stored.

The escape is opaque to correctness: it only ever selects the already-proven stored block, so the roundtrip and padding proofs (`inflate_deflateRaw`, `deflateRaw_pad`, `deflateRaw_goR_pad`) each extend with one extra stored case. New tests in `ZipTest/NativeDeflate.lean` pin the classifier (random fires; text, base64, repeated 8/16 KiB blocks, half-compressible, self-similar, and below-gate inputs do not) and roundtrip incompressible input through `deflateRaw` at every level via both native and FFI inflate.

## Before / after

**Dashboard (Canterbury + Silesia, `native` compressor, same machine, identical inputs).** Every native compression ratio is **byte-identical to master across all 57 corpus rows** (`max |Δratio| = 0.000000`) — the gate never fires on a compressible corpus file, so the speed-vs-ratio pareto curves are unchanged. No corpus file is incompressible enough to trip it, exactly as the issue predicted ("Canterbury/Silesia mostly compressible, so the geomean barely moves").

**Compress throughput (`bench compress-pareto`, 4 MiB, back-to-back vs master binary):**

| input | level | before | after | |
|-------|------:|-------:|------:|--|
| incompressible (PRNG) | 6 | 50 MB/s | 674 MB/s | 13.6× faster, identical output |
| incompressible (PRNG) | 9 | 2.6 MB/s | 660 MB/s | 259× faster, identical output |
| compressible (text) | 6 | 181 MB/s | 180 MB/s | within noise (entropy early-exit) |

The win is on real-world incompressible inputs (already-compressed archives, JPEG/PDF/encrypted blobs, random data), not the standard text corpora — measured here on PRNG bytes, which share the high-entropy / no-repeat profile of compressed data.

🤖 Prepared with Claude Code

### DIFF
--- a/Zip/Native/DeflateDynamic.lean
+++ b/Zip/Native/DeflateDynamic.lean
@@ -955,6 +955,141 @@ def deflateDynamicBlocksOptimal (data : ByteArray) (tokChunk : Nat) : ByteArray 
   else
     (emitSharedBlocks data (lz77OptimalIter data) tokChunk 0 BitWriter.empty).flush
 
+/-! ## Incompressible pre-scan
+
+`deflateRaw` already falls back to a stored block whenever every compressed
+candidate is larger — but it only learns this *after* paying the full hash-chain
+match pass (and, at level 9, the optimal-parse DP), which is essentially the
+whole compress cost and finds almost nothing on incompressible input. Measured
+(`bench compress-pareto`, 4 MiB PRNG): level 9 runs at ≈2.5 MB/s, all of it
+wasted before the stored fallback. The pre-scan reads up to ≈128 KiB of bounded
+sample and, when the input is unambiguously incompressible, routes straight to
+`deflateStoredPure` so the matcher never runs; a compressible file fails the
+first sampled region and falls through to the normal path almost immediately. -/
+
+/-- Minimum input size for the incompressible pre-scan to engage. The pre-scan's
+    fixed cost (one presence-table allocation + a region scan) is only worth paying
+    when the matcher it might skip is the dominant cost, i.e. on large inputs; below
+    1 MiB the matcher is cheap enough that the scan would be net overhead on the
+    common compressible file, so small inputs always take the normal path. This is
+    why the gate leaves the Canterbury/Silesia dashboard untouched — every corpus
+    file is either compressible or under the size gate. -/
+def prescanMinSize : Nat := 1048576
+
+/-- Number of contiguous sample regions the pre-scan reads, spread end to end
+    across the input (first at offset 0, last ending at `n`). A region that looks
+    even slightly compressible short-circuits the whole scan, so a normal
+    compressible file pays for at most the first region. -/
+def prescanRegions : Nat := 4
+
+/-- Bytes per sample region. Each region is scanned *contiguously* and every
+    consecutive 4-gram is inserted, so repetition is detected densely the way the
+    matcher itself finds matches. At 32 KiB a region spans two of the common
+    page/block sizes (4/8/16 KiB), so a repeated block surfaces as a wall of
+    4-gram collisions. ≤`prescanRegions · prescanRegionBytes` (128 KiB) bytes are
+    scanned, independent of input size. -/
+def prescanRegionBytes : Nat := 32768
+
+/-- log2 of the per-region 4-gram presence-table size (2^20 slots). Sized so that,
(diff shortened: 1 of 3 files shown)


kim-em:
## Before/after on the standard corpora (native compress) — no regression

Generated with the new `perf-pr-graphs` skill (#2647). **Rendered graphs (native before/after overlaid on the language-peer curves), Canterbury + Silesia:** https://gist.github.com/kim-em/99e7fa8ed4c0189bd19a5a24871e7abf — report artifact, intentionally not committed.

**Provenance** (measured same-machine; peers reused from the committed dashboard snapshot):
- BEFORE: `391eba0b` — this branch's merge-base with master (built in a throwaway worktree, so the baseline isolates exactly this PR's commit)
- AFTER: `c550b25b` — this PR
- coverage: before/after cover identical `(pattern, level)` rows; `max |Δratio| = 0.000000` (byte-identical output everywhere — the escape never fires on a compressible corpus file)

**Compression speed, geomean by level (after / before):**

| level | Canterbury (median-of-5) | Silesia (single-pass) |
|------:|:------------------------:|:---------------------:|
| 1 | 0.99× | 1.02× |
| 6 | 1.01× | 1.01× |
| 9 | 1.00× | 1.06× |

Canterbury (the row set to trust for a speed verdict) is flat at 0.99–1.01×. Silesia's small wiggles (e.g. L8 1.10×) are `reps=1` single-pass noise, not signal — its ratio is what matters there and it is identical. The native before/after curves are coincident on both corpora.

**Bottom line:** on the standard corpora the pre-scan is invisible — ratio byte-identical, speed unchanged — exactly as the issue predicted (Canterbury/Silesia are compressible, so no file clears the entropy gate). The win is on genuinely incompressible input the corpora don't contain: 4 MiB PRNG compress goes 50→674 MB/s (L6) and 2.6→660 MB/s (L9) with byte-identical output (per the PR description).

Issue #2650, opened 2026-06-18
perf(decode): de-box the Huffman symbol table (Array (UInt16 × UInt8) -> packed word)


kim-em:
**The lowest-proof-risk decode lever; a good warm-up and the foundation for multi-symbol decode.**

## Finding
See the fresh post-#2645 decode profile in https://github.com/kim-em/lean-zip/issues/2628#issuecomment-4737388343 (perf on `Inflate.inflate`, dickens L6). The 9-bit decode table is `Array (UInt16 × UInt8)` (`buildTable`/`tableEntry`, `Inflate.lean:145,160`). Each per-symbol read in `decodeSym` is `table[idx]!` → a **boxed heap pair**: the generated C of `goFused` shows `lean_array_fget_borrowed` then `lean_ctor_get_uint16` (sym) + `lean_ctor_get` (len) — a pointer-chase per symbol, executed once per literal and twice per match. `walkTree` (4.8%, the long-code fallback) reads the same boxed representation.

## Idea
Pack each entry into a single scalar — `sym (16) | len (8)` — and read it with one load + two masks, no pair object. **Generated C is the arbiter**: `Array UInt32` is *not* inline-scalar in this runtime (it is tagged objects → `fget` + `unbox_uint32`, which removes the pair box but keeps a tagged read); a true de-box wants a `ByteArray`-backed word via `getUInt32LE`. Spike both, read the C, keep whichever actually drops the `ctor_get` chase.

## Touch points
`Zip/Native/Inflate.lean`: `tableEntry` (145), `buildTable` (160), `decodeSym` (583), `walkTree` (567). Proofs: the `buildTable`/`decodeWithTable` correspondence in `Zip/Spec/` decode specs (representation-change equality transfer; statements unchanged).

## Magnitude / risk
- Bounded — the table read is ~4 of `goFused`'s `ctor_get` + the `walkTree` 4.8%. Estimate **~5-8% decode**. Smaller than the tuple-de-alloc, but the **lowest proof risk** (localized representation change) and it sets up a wider multi-symbol table later.

## Why we might reject
The boxed read may be a small slice of `goFused` once the tuple-de-alloc lands; re-measure ordering matters. Confirm the pointer-chase actually disappears in the C.

## Ordering
Collides with the tuple-de-alloc (#2649) and all other `goFused` levers — same function/file. Sequence, do not parallelize. Reasonable to do **first** as a low-risk foundation, or after #2649.

kim-em:
[...]

PR #2657, opened 2026-06-18
perf(decode): de-box the Huffman fast-decode table into parallel scalar arrays

kim-em:
This PR de-boxes the 9-bit Huffman fast-decode table that drives `Inflate.inflate`, closing https://github.com/kim-em/lean-zip/issues/2650 — "perf(decode): de-box the Huffman symbol table (Array (UInt16 × UInt8) -> packed word)". The table was `Array (UInt16 × UInt8)`, which stores each slot as a boxed heap pair, so every per-symbol read in the `goFused` hot loop compiled to `lean_array_get` followed by two `lean_ctor_get` calls pointer-chasing into a per-element pair — once per literal, twice per match. This replaces it with a `DecodeTable` structure holding two parallel scalar arrays (`syms : Array UInt16`, `lens : Array UInt8`), so each read is `lean_array_fget` + `unbox` with no chase. The generated C confirms the per-element `ctor_get` pointer-chase is gone; the only remaining `ctor_get(table, 0/1)` are loop-invariant struct-field reads on the same hot table object.

The issue suggested packing each entry into a single `UInt32` word read with masks. That approach does not survive elaboration here: extracting the high field of a packed word needs a shift, and the well-founded recursions `go`/`goFused` force that shift expression to weak head normal form during equation compilation, looping on `Nat.div` (deterministic timeout at whnf, even at 1M heartbeats; `@[irreducible]` on the accessors does not help, because the WF equation compiler's whnf unfolds through it). Plain array reads stay stuck under whnf, so the two-array split is what keeps both the de-box and the proofs. The rationale is documented on `DecodeTable`.

The correctness proofs transfer as a localized representation change with the load-bearing statements unchanged (`decodeWithTable_eq`, `goFused_eq_go`, and hence `decodeHuffmanFast = decodeHuffman`): `buildTable_syms_getElem` / `buildTable_lens_getElem` replace the single `htab` projection lemma, and `decodeSym_corr` generalizes the two parallel reads instead of one pair. There is no new `sorry`, no `native_decide`, no trust gap. The full library builds and the test suite passes, including the InflateTable differential test (table decode vs canonical tree walk) over 72 stream batches across 9 trees.

🤖 Prepared with Claude Code

### DIFF
--- a/Zip/Native/Inflate.lean
+++ b/Zip/Native/Inflate.lean
@@ -154,11 +154,29 @@ where
       else if bits % 2 == 0 then go z (bits / 2) (depth + 1)
       else go o (bits / 2) (depth + 1)
 
+/-- A de-boxed fast-decode table: the `(symbol, codeLen)` of each slot split into
+    two parallel scalar arrays. Storing `Array (UInt16 × UInt8)` boxes every slot
+    as a heap pair, so each per-symbol read pointer-chases through `lean_ctor_get`
+    twice; the split keeps each field in a scalar array (`lean_array_fget` +
+    `unbox`, no chase), which is the de-box of #2650.
+
+    Two scalar arrays rather than one packed `UInt32` word: extracting the high
+    field of a packed word needs a shift, and the well-founded recursions `go` /
+    `goFused` would force that shift to weak head normal form during equation
+    compilation and loop on `Nat.div`. A plain array read stays stuck under
+    `whnf`, so the split keeps both the de-box and the proofs. -/
+structure DecodeTable where
+  syms : Array UInt16
+  lens : Array UInt8
+
 /-- Build the `2^fastBits`-entry decode table for `tree`: slot `i` holds the
-    `(symbol, codeLen)` reached by walking `tree` on the bits of `i`. Built once
-    per Huffman tree; cheap relative to the symbols decoded with it. -/
-def buildTable (tree : HuffTree) : Array (UInt16 × UInt8) :=
-  Array.ofFn (n := 2 ^ fastBits) (fun i : Fin (2 ^ fastBits) => tableEntry tree i.val)
+    `(symbol, codeLen)` reached by walking `tree` on the bits of `i`, with the two
+    fields split into the parallel scalar arrays of `DecodeTable` so each
+    per-symbol read is a de-boxed scalar load instead of a heap-pair pointer-chase.
+    Built once per Huffman tree; cheap relative to the symbols decoded with it. -/
+def buildTable (tree : HuffTree) : DecodeTable where
+  syms := Array.ofFn (n := 2 ^ fastBits) (fun i : Fin (2 ^ fastBits) => (tableEntry tree i.val).1)
+  lens := Array.ofFn (n := 2 ^ fastBits) (fun i : Fin (2 ^ fastBits) => (tableEntry tree i.val).2)
 
 /-- Bits remaining in the reader from its current `(pos, bitOff)`. -/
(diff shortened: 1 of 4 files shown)


kim-em:
## Before/after decode throughput (`decompress_mbps`, machine `chungus`)

**Output-neutral de-box.** `max |Δratio| = 0.000000` — output is bit-identical, as expected for a pure representation change. The win is structural and confirmed in the generated C: the per-element `lean_ctor_get` pointer-chase into the boxed `(UInt16 × UInt8)` pair is gone (reads are now `lean_array_fget` + `unbox` on the two scalar arrays).

**Wall-clock delta is within measurement noise here** — read it as decode-neutral, not as a measured speedup. BEFORE is the committed dashboard (`bench/results/latest.json`, median-of-5, commit `3a38cf0a`); AFTER is a fresh run (commit `01ce536d`), so this is a cross-session comparison, and the AFTER run shared the machine with two other Lean builds pegging cores. Canterbury (the median-of-5 set) reads ~0.97x; Silesia (single-pass, ±30% noise) reads ~1.01–1.03x. The signs disagree → the true effect is below this setup's noise floor.

canterbury decode
silesia decode

Canterbury `decompress_mbps` geomean by level (median-of-5; the reliable speed set):

|------:|------:|-------:|------:|--------:|
| 1 | 0.314 | 103.0 |  99.9 | 0.97x |
| 2 | 0.311 | 109.4 | 106.1 | 0.97x |
| 3 | 0.309 | 114.5 | 110.9 | 0.97x |
| 4 | 0.305 | 116.4 | 112.5 | 0.97x |
| 5 | 0.305 | 116.9 | 112.7 | 0.96x |
| 6 | 0.304 | 118.5 | 114.4 | 0.97x |
| 7 | 0.297 | 117.9 | 114.0 | 0.97x |
| 8 | 0.296 | 118.8 | 114.6 | 0.96x |

Silesia (single-pass — use for ratio/coverage, not a fine speed delta): ~1.01–1.03x across L1–L8.

kim-em:
## Controlled re-measurement — supersedes the cross-session numbers above

The first comment compared a fresh AFTER against the recorded `latest.json` (a *cross-session* baseline), whose few-percent noise floor couldn't settle a single-digit delta. I re-measured with **interleaved BEFORE/AFTER runs, same session, pinned to one core**, so common-mode load/turbo/thermal cancels and the paired ratio is the real signal. Output is still bit-identical (`max |Δratio| = 0`).

**Verdict: the two corpora genuinely diverge, and Silesia (large real-world files) — the corpus that matters — is a clear win at every level.**

| corpus | controlled after/before | 95% CI | n |
|---|---|---|---|
| Canterbury (small synthetic, ≤1 MB) | **−4.5%** | [−5.5%, −3.5%] | 594 |
| Silesia 1–8 (large real-world) | **+2.2%** | [+1.5%, +3.0%] | 576 |
| Silesia L9 | **+4.9%** | [+1.3%, +8.6%] | 24 |

Cause (from the generated C): the de-box trades the boxed pair's single pointer-chase for two scalar-array reads. On Silesia's large hot-loop files the compact scalar arrays (2 KB + 4 KB) win on cache behaviour vs scattered heap pairs; on tiny Canterbury files everything is in cache regardless, so the extra per-symbol read just shows as overhead. L9's match-heavy streams call `decodeSym` most, which is why its Silesia gain is largest.

silesia decode (controlled)
canterbury decode (controlled)

Net: a real, statistically-significant decode improvement on the representative corpus across all levels including L9, at the cost of a regression on small in-cache inputs. Output-neutral, fully proven. Merging on that basis.

Issue #2651, opened 2026-06-18
perf(decode): output literal-run fast path + pre-sized output buffer


kim-em:
**Targets the addressable part of the ~31% output-copy path.**

## Finding
See the fresh post-#2645 decode profile in https://github.com/kim-em/lean-zip/issues/2628#issuecomment-4737388343 (perf on `Inflate.inflate`, dickens L6). The output path is ~31% of decode: `lean_byte_array_copy_slice` 11.9% + `memmove` 7.5% + `Inflate.copyLoop` 6.5% + `lean_copy_sarray` 3.6% + `ByteArray.extract` 2.1%. The `memmove`/`copy_slice` core (~19%) is irreducible (the bytes must be copied). The **addressable ~12%** is: `copyLoop` (the overlap/short-distance copy), `copy_sarray` (the output buffer's doubling reallocs — `goFused` does `output.push sym.toUInt8` one byte per literal), and `extract`.

## Idea
1. **Literal-run fast path**: batch consecutive literals instead of one `output.push` + one decode-tuple per literal — decode a run, append in one shot.
2. **Pre-sized output buffer**: `goFused` has `maxOut` and (for most streams) a knowable decompressed size; pre-allocating the output removes the `copy_sarray` doubling reallocs.
3. Tighten the short-distance overlap copy in `copyLoop`.

## Touch points
`Zip/Native/Inflate.lean`: `goFused` literal branch (666+), `copyLoop` (the overlap copy), output buffer threading. Proofs: content-preservation / append correctness for the buffer (the `lean-content-preservation` skill covers this shape).

## Magnitude / risk
- ~12% addressable; realistic capture maybe half that. The memmove core is a hard floor.
- Pre-sizing needs a sound output-size bound to stay proven (cf. the BitWriter wide-store #2631 pre-size concern on the compress side).

## Ordering
Touches `goFused` → collides with #2649 / the de-box; sequence after them (the literal path is simplest to restructure once the per-symbol tuple is already scalar).

kim-em:
Closing — the addressable decode-output work this issue targeted is done and in production, though not by the route the title proposed.

[...]

PR #2660, opened 2026-06-18
perf(decode): pre-sized output buffer via threaded sizeHint

kim-em:
This PR adds an optional `sizeHint` parameter to the native DEFLATE decoders (`Inflate.inflate`/`inflateRaw` and `InflateBuf.inflate`) that pre-reserves the output buffer's capacity, removing the doubling reallocations (`lean_copy_sarray`) that a buffer grown one literal at a time otherwise pays. This targets the `copy_sarray` portion of the ~31% output-copy path identified in https://github.com/kim-em/lean-zip/issues/2651 "perf(decode): output literal-run fast path + pre-sized output buffer".

`0` (the default) reserves nothing, so existing callers and every correctness proof are unchanged: `ByteArray.emptyWithCapacity n` is definitionally `{ data := Array.empty }` for every `n`, so the initial buffer is still `.empty`. Two spec proofs that `rw` over the unfolded buffer pick up a one-token `emptyWithCapacity_eq_empty` normalization; the rest transfer untouched. The new `rfl` bridge lemmas `Inflate.inflateRaw_sizeHint_eq`, `Inflate.inflate_sizeHint_eq`, and `InflateBuf.inflate_sizeHint_eq` state that any `sizeHint` equals the verified `sizeHint := 0` form, so every existing correctness/suffix/bounds theorem transfers verbatim to a hinted call.

The ZIP entry decoder passes the entry's declared uncompressed size. Because `uncompressedSize` is untrusted metadata, the speculative reservation is bounded by three guards no malicious entry controls: a fixed absolute ceiling `nativePresizeCap` (64 MiB, the real safety property), the `maxEntrySize` output budget (so a `0` reject-all budget reserves nothing), and a heuristic `compData.size * 1100` ratio bound. The bench passes the known decompressed size.

Measured native decode speedup from pre-sizing alone (10 MiB, level 6): text 1.12x, prng 1.36x, cyclic 1.14x, constant 1.09x.

The output literal-run fast path (issue item 1) is deferred to a follow-up: it requires restructuring the `go_corr` symbol-by-symbol bisimulation and `goFused_eq_go`, and its marginal gain over pre-sizing — now that pushes no longer realloc — is unmeasured. This PR's pre-size half stands alone, fully proven and measured.

🤖 Prepared with Claude Code

### DIFF
--- a/Zip/Archive.lean
+++ b/Zip/Archive.lean
@@ -1212,7 +1212,17 @@ private def readEntryData (h : IO.FS.Handle) (entry : Entry) (label : String)
         -- output", unlike the FFI path where `0` means unlimited. Callers
         -- that opt into `maxEntrySize := 0` for unlimited FFI decompression
         -- therefore get an immediate rejection on the native backend.
-        match Zip.Native.Inflate.inflate compData maxEntrySize.toNat with
+        --
+        -- Pre-size the output buffer to the entry's declared uncompressed size,
+        -- removing the doubling reallocations of a buffer grown one literal at a
+        -- time. `uncompressedSize` is untrusted ZIP metadata, so cap the hint to a
+        -- multiple of the compressed bytes already in RAM: DEFLATE's maximum ratio
+        -- is ~1032:1, so this never truncates a hint for a valid stream, but stops
+        -- a tiny malicious entry from forcing a huge speculative allocation. The
+        -- hint is a capacity hint only — never affecting the bytes produced or the
+        -- `maxEntrySize` bomb guard.
+        let sizeHint := min entry.uncompressedSize.toNat (compData.size * 1100)
+        match Zip.Native.Inflate.inflate compData maxEntrySize.toNat (sizeHint := sizeHint) with
         | .ok data => pure data
         | .error msg => throw (IO.userError s!"zip: native inflate failed for {label}: {msg}")
       else RawDeflate.decompress compData maxEntrySize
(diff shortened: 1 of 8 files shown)


kim-em:
## Before/after `decompress_mbps` on the real corpora

Machine `Linux chungus x86_64` (same as the committed dashboard). Metric: native decode throughput vs ratio, overlaid on the reused other-language curves. Output-neutral: `max |Δratio| = 0.000000`.

### Silesia (the corpus that matters)
silesia decode

### Canterbury
canterbury decode

### Verdict — paired interleaved A/B (hint vs no-hint, same binary, `taskset`-pinned, common-mode noise cancelled)

This is the trustworthy delta. A single cross-session pass cannot resolve it (Silesia is single-pass, ±30% run-to-run), which is exactly why the overlay's before/after sit on top of each other.

| corpus | speedup (no-hint time / hint time) | 95% CI | verdict |
|---|---|---|---|
| **silesia** (n=144) | **1.024x** | [1.017, 1.030] | real, ~2.4% |
| canterbury (n=132) | 1.00x | [0.987, 1.010] | within noise |

The +2.4% on real Silesia is consistent with the `copy_sarray` doubling-realloc line being ~3.6% of decode in the profile — pre-sizing captures most of it. It is **far below** the synthetic-pattern numbers I first quoted (text 1.12x, prng 1.36x); those patterns (especially prng→stored blocks) are not representative of real data.

### Positioning (visual read)
Native lean-zip decode sits in the **slow cluster** (~100–150 MB/s) on both corpora — above OCaml, alongside Go, and well below the C/Rust/SIMD decoders (zlib ~430, miniz_oxide ~480, libdeflate ~850 MB/s). **This PR does not move that positioning**; before (grey) and after (red) overlap. The lone Canterbury L1 red dip is single-pass jitter on tiny files, not a regression (Δratio = 0, and the paired A/B shows no Canterbury change).

**Caveats:** AFTER measured at levels 1,6 only (decode speed is near-level-independent); the overlay used the level-{1,6} intersection with the dashboard. The headline number is the paired A/B, not the noisy cross-session geomean.

So: a small (+2.4% Silesia), no-downside, fully-proven free win. Happy to hold for your call before merging.

PR #2663, opened 2026-06-18
perf(decode): drop redundant per-symbol guards from the wide-buffer decoder

kim-em:
This PR removes the per-symbol overhead in the wide-buffer Huffman decode loop that exists purely as well-founded-termination scaffolding. The loop's two per-symbol `bitpos`-progress guards — needed only to satisfy `goFused`'s termination proof, never firing on valid input — sat on the critical path; this PR drops them, drops the redundant `bitpos` parameter (it equals `pos*8 − cnt`), and decodes the common literal inline (skipping `decodeSym`'s per-symbol result tuple). Output is byte-identical; the measured decode speedup is **≈ +3% Canterbury / +4% Silesia**, uniform across all levels.

The guard-light loop `goFusedP` (a real terminating `def`, measure `(data.size − pos) * 9 + cnt`) is proven equal to the production `goFused` by `goFusedP_eq`: a `goFusedP … = (goFused … bp …).map (drop bitpos)` bisimulation by functional induction, carrying `bitpos` as a variable under the buffer invariant `bp + cnt = pos*8` with `pos ≤ data.size`. The inline-literal fast path agrees with `goFused`'s `decodeSym` step via `decodeSym_inline`, the cnt-based no-progress guard corresponds to `goFused`'s bitpos-based one, and `goFused`'s out-of-range guard is shown dead. `goFusedP` is then wired into the production `decodeHuffmanFastBuf`, and `decodeHuffmanFastBuf_eq` bridges `goFusedP → goFused → go` so every downstream correctness proof transfers unchanged.

Builds clean, all tests pass, zero `sorry`. Before/after decode graphs and the in-PR dashboard refresh are included; the graphs are posted as a PR comment.

🤖 Prepared with Claude Code

### DIFF
--- a/Zip/Native/Inflate.lean
+++ b/Zip/Native/Inflate.lean
@@ -762,7 +762,7 @@ def goFusedP (litTable distTable : HuffTree.DecodeTable)
         (output.push litTable.syms[(bitBuf &&& 0x1FF).toNat]!.toUInt8)
   else
   let cnt0 := cnt
-  match hds0 : decodeSym litTree litTable bitBuf cnt with
+  match decodeSym litTree litTable bitBuf cnt with
   | .error e => .error e
   | .ok (sym, bitBuf, cnt, _used) =>
     if sym < 256 then
@@ -778,7 +778,7 @@ def goFusedP (litTable distTable : HuffTree.DecodeTable)
         let extra := Inflate.lengthExtra[idx]'(by simp [Inflate.lengthExtra_size, Inflate.lengthBase_size] at h ⊢; omega)
         let (extraBits, bitBuf, cnt) ← takeBits bitBuf cnt extra.toNat
         let length := base.toNat + extraBits
-        match hds1 : decodeSym distTree distTable bitBuf cnt with
+        match decodeSym distTree distTable bitBuf cnt with
         | .error e => .error e
         | .ok (distSym, bitBuf, cnt, _dused) =>
           let dIdx := distSym.toNat
@@ -817,9 +817,11 @@ def decodeHuffmanFastBuf (br : BitReader) (output : ByteArray)
   -- align: drop the bitOff bits already consumed in the first byte
   let bitBuf := bitBuf >>> br.bitOff.toUInt64
   let cnt := cnt - br.bitOff
-  let bitpos := br.pos * 8 + br.bitOff
-  let (out, pos', bitBuf', cnt', _bitpos') ←
-    goFused litTable distTable br.data litTree distTree maxOut br.data.size pos bitBuf cnt bitpos output
+  -- `goFusedP` is the guard-light loop: no threaded `bitpos`, no provably-dead
+  -- out-of-range guard, and the common literal decoded inline. Proven equal to
+  -- `goFused` (`Zip.Spec.InflateBufCorrect.goFusedP_eq`), hence to the reference.
+  let (out, pos', bitBuf', cnt') ←
+    goFusedP litTable distTable br.data litTree distTree maxOut pos bitBuf cnt output
   let _ := bitBuf'
   let endbit := pos' * 8 - cnt'
   .ok (out, { data := br.data, pos := endbit / 8, bitOff := endbit % 8 })
--- a/Zip/Native/Inflate.lean
+++ b/Zip/Native/Inflate.lean
@@ -733,6 +733,79 @@ decreasing_by
   · exact Prod.Lex.left _ _ (by omega)
   · exact Prod.Lex.left _ _ (by omega)
 
+/-! ## EXPERIMENT: provable-shape guard-light decoder (measure the landable win)
+
+`goFusedP` is the shape a *proven* version would have: no `bitpos` parameter
+(it equals `pos*8 - cnt`, so the termination measure is
+`((data.size - pos)*8 + cnt, data.size - pos)` over `pos`/`cnt` already threaded),
+no out-of-range guard (provably dead under `pos ≤ data.size`), and the common
+literal decoded inline (its `llen ≠ 0` check already guarantees ≥1 bit, so that
+path needs no progress guard either). Only the non-literal `decodeSym` path keeps
+a cheap `cnt`-based no-progress guard. `partial` here just to measure; the real
+PR proves termination from the measure above. -/
+def goFusedP (litTable distTable : HuffTree.DecodeTable)
+    (data : ByteArray) (litTree distTree : HuffTree) (maxOut : Nat)
+    (pos : Nat) (bitBuf : UInt64) (cnt : Nat) (output : ByteArray) :
+    Except String (ByteArray × Nat × UInt64 × Nat) := do
+  if hrc : cnt ≤ 56 ∧ pos < data.size then
+    goFusedP litTable distTable data litTree distTree maxOut
+      (pos + 1) (bitBuf ||| (data[pos]!.toUInt64 <<< cnt.toUInt64)) (cnt + 8) output
+  else
+  if hlit : litTable.lens[(bitBuf &&& 0x1FF).toNat]!.toNat ≠ 0
+      ∧ litTable.lens[(bitBuf &&& 0x1FF).toNat]!.toNat ≤ cnt
+      ∧ litTable.syms[(bitBuf &&& 0x1FF).toNat]! < 256 then
+    if output.size ≥ maxOut then throw "Inflate: output exceeds maximum size"
+    else
+      goFusedP litTable distTable data litTree distTree maxOut pos
+        (bitBuf >>> (litTable.lens[(bitBuf &&& 0x1FF).toNat]!.toNat).toUInt64)
+        (cnt - litTable.lens[(bitBuf &&& 0x1FF).toNat]!.toNat)
+        (output.push litTable.syms[(bitBuf &&& 0x1FF).toNat]!.toUInt8)
+  else
+  let cnt0 := cnt
+  match hds0 : decodeSym litTree litTable bitBuf cnt with
+  | .error e => .error e
+  | .ok (sym, bitBuf, cnt, _used) =>
+    if sym < 256 then
+      if output.size ≥ maxOut then throw "Inflate: output exceeds maximum size"
+      else if hnp : cnt0 ≤ cnt then throw "Inflate: no progress in Huffman decode"
+      else goFusedP litTable distTable data litTree distTree maxOut pos bitBuf cnt (output.push sym.toUInt8)
+    else if sym == 256 then .ok (output, pos, bitBuf, cnt)
+    else
+      let idx := sym.toNat - 257
+      if h : idx ≥ Inflate.lengthBase.size then throw s!"Inflate: invalid length code {sym}"
+      else
+        let base := Inflate.lengthBase[idx]
+        let extra := Inflate.lengthExtra[idx]'(by simp [Inflate.lengthExtra_size, Inflate.lengthBase_size] at h ⊢; omega)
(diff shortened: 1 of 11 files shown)


kim-em:
## Handoff for a fresh session

Everything needed to take over cold. The branch already has `goFusedP` (terminating `def`, in `Zip/Native/Inflate.lean`) and `decodeSym_inline` (helper, in `Zip/Spec/InflateBufCorrect.lean`), both committed and building. What's below is the remaining work + how to reproduce the measurement (the A/B harness was throwaway and is **not** on the branch).

### Remaining proof obligation — exact statement

Add to `Zip/Spec/InflateBufCorrect.lean` (after `decodeSym_inline`). The statement and `goFusedP.induct` are confirmed to elaborate (~12 cases):

```lean
set_option maxHeartbeats 2000000 in
set_option maxRecDepth 8000 in
theorem goFusedP_eq (litTable distTable : HuffTree.DecodeTable) (data : ByteArray)
    (litTree distTree : HuffTree) (maxOut : Nat) :
    ∀ (pos : Nat) (bitBuf : UInt64) (cnt : Nat) (output : ByteArray),
    pos ≤ data.size → cnt ≤ pos * 8 →
    (match goFusedP litTable distTable data litTree distTree maxOut pos bitBuf cnt output,
           goFused litTable distTable data litTree distTree maxOut data.size pos bitBuf cnt
             (pos * 8 - cnt) output with
     | .error e1, .error e2 => e1 = e2
     | .ok (o1, p1, b1, c1), .ok (o2, p2, b2, c2, _) => o1 = o2 ∧ p1 = p2 ∧ b1 = b2 ∧ c1 = c2
     | _, _ => False) := by
  intro pos bitBuf cnt output
  induction pos, bitBuf, cnt, output using goFusedP.induct
  -- ~12 cases; each: unfold goFusedP (case ctx picks the branch) + rw [goFused] one step, then align
  all_goals sorry
```

### Proof plan (per case)
- **Invariant carried:** `bitpos = pos*8 - cnt` (so `goFused`'s threaded `bitpos` stays `= pos*8 - cnt`; preserved by both refill and decode — verify with `omega`), plus `pos ≤ data.size`, `cnt ≤ pos*8`. Don't `intro hpos hcnt` before `induction` (they depend on the induction vars); intro them inside each case.
- **Out-of-range guard is dead:** `goFused`'s `dataSize*8 < bitpos + (cnt0 - cnt)` is `pos*8 - cnt_new ≤ dataSize*8`, true since `pos ≤ dataSize`. So `goFused`'s `if _h₂ : ...` always takes the `else`.
- **Inline literal = decodeSym:** in `goFusedP`'s inline branch, `decodeSym litTree litTable bitBuf cnt = .ok (syms[idx], bitBuf>>>len, cnt-len, len)` by `decodeSym_inline` — so it matches `goFused`'s `decodeSym` step. (For long literals `llen=0`, `goFusedP` falls through to the same `decodeSym` path as `goFused`.)
- **No-progress guards correspond:** both throw on `cnt0 ≤ cnt_new`; in the inline path `llen ≥ 1` so `goFused`'s guard there is also false.
- **bitpos rewrite at each recursive call:** `goFused`'s recursive `bitpos + (cnt0 - cnt')` `= pos'*8 - cnt'` (omega) → rewrite, then apply the IH (which states `goFused` at `pos'*8 - cnt'`).
- **Template:** mirror `go_corr` (same file, ~line 765) and its `repeat' (first | rfl | split)` for the back-reference guard cascade.

### Then wire it in
- `decodeHuffmanFastBufP` (copy of `decodeHuffmanFastBuf` calling `goFusedP`; endbit from `pos'*8 - cnt'`).
- `decodeHuffmanFastBufP_eq : decodeHuffmanFastBufP = decodeHuffmanFastBuf` (from `goFusedP_eq` + the `refill`/init setup; needs the initial `pos ≤ data.size`, `cnt ≤ pos*8` from `refill_corr`).
- Repoint `Inflate.decodeHuffmanFast` → `decodeHuffmanFastBufP`; compose `decodeHuffmanFastBufP_eq` into the existing `decodeHuffmanFast_eq` so all correctness transfers.
- Then in-PR perf graphs + dashboard refresh per the `perf-pr-graphs` skill.

### Reproduce the +3.9% (paste-back harness, NOT for merge)

Add these entry points to `Zip/Native/InflateBuf.lean`:

```lean
partial def inflateLoopBufP (br : BitReader) (output : ByteArray)
    (fixedLit fixedDist : HuffTree) (maxOut dataSize : Nat) : Except String (ByteArray × Nat) := do
  let (bfinal, br₁) ← br.readBits 1
  let (btype, br₂) ← br₁.readBits 2
  let (output', br') ← match btype with
    | 0 => Inflate.decodeStored br₂ output maxOut
    | 1 => decodeHuffmanFastBufP br₂ output fixedLit fixedDist maxOut
    | 2 => do let (lt, dt, br₃) ← Inflate.decodeDynamicTrees br₂
              decodeHuffmanFastBufP br₃ output lt dt maxOut
    | _ => throw s!"Inflate: reserved block type {btype}"
  if bfinal == 1 then return (output', br'.alignToByte.pos)
  else if br'.bitPos ≤ br.bitPos then throw "Inflate: no progress in inflate loop"
  else if dataSize * 8 < br'.bitPos then throw "Inflate: bit position out of range"
  else inflateLoopBufP br' output' fixedLit fixedDist maxOut dataSize

def inflateP (data : ByteArray) (maxOut : Nat := 1024*1024*1024) (sizeHint : Nat := 0) :
    Except String ByteArray := do
  let br : BitReader := { data, pos := 0, bitOff := 0 }
  let fixedLit ← HuffTree.fromLengths Inflate.fixedLitLengths
  let fixedDist ← HuffTree.fromLengths Inflate.fixedDistLengths
  let (output, _) ← inflateLoopBufP br (ByteArray.emptyWithCapacity sizeHint) fixedLit fixedDist maxOut data.size
  return output
```
where `decodeHuffmanFastBufP` is the obvious copy of `decodeHuffmanFastBuf` calling `goFusedP` (drop the 5th `bitpos` result component; `endbit := pos'*8 - cnt'`).

Then a paired interleaved A/B in `ZipBenchReport.lean` (compare `InflateBuf.inflate` vs `InflateBuf.inflateP`, both `sizeHint := data.size`), `@[noinline]` decode wrappers, `timePerOp 1` per side alternating order, geomean of log-ratios + 95% CI, run pinned: `taskset -c <free-core> lake env .lake/build/bin/bench-report --provable-ab 12`. Verify byte-identical output before timing. Needs the gitignored Silesia corpus (`bash bench/fetch_corpora.sh silesia`); run from the main worktree (Silesia is only fetched there). **Measured: +2.9% Canterbury / +3.9% Silesia, level 6, machine `chungus`.**

### Findings recap (why this and not a literal-run)
- Literal-run *batching* is **counterproductive** (−15%): the loop is already tight post-#2657/#2660, and batching adds tuple/peek overhead.
- The win is removing the per-symbol `bitpos`-progress guards (pure WF scaffolding, never fire on valid input) + not threading `bitpos` + inlining the literal decode. Isolation showed inline-with-guards is neutral; the guards are the cost. The simpler subset (drop bitpos + out-of-range guard, keep `decodeSym` for literals) is ~+1.7% with a smaller proof, if the full inline proof proves too costly.

kim-em:
### Proof-infra finding for the bisimulation (read before grinding cases)

Started the `goFusedP_eq` bisimulation; the statement and `goFusedP.induct` elaborate, and the **refill case proof pattern works** with one important reformulation, but there's a `goFusedP.induct` quirk to fix first.

**Reformulate the statement to carry `bitpos` as a variable** (avoids all rewriting under the `match` motive — the original `pos*8 - cnt` form fights `rw`):
```lean
∀ (pos : Nat) (bitBuf : UInt64) (cnt : Nat) (output : ByteArray) (bp : Nat),
  pos ≤ data.size → bp + cnt = pos * 8 →
  (match goFusedP … pos bitBuf cnt output, goFused … data.size pos bitBuf cnt bp output with …)
```
Then the refill case is just: `intro bp hpos hbp; rw [goFusedP, dif_pos hrc, goFused, dif_pos hrc]; exact ih bp (by omega) (by omega)` — recursive calls pass the exact `bp` to the IH, no match-motive rewrites. Decode cases pass `bp + (cnt0 - cnt')`; the out-of-range guard is killed by `dif_neg (by omega)` using `bp + cnt = pos*8` and `pos ≤ data.size`.

**The blocker:** `goFusedP.induct` generates the refill case hypothesis as `hrc : cnt ≤ 56 ∧ pos < ByteArray.size ?data` — a **metavariable** `?data` — so `dif_pos hrc` doesn't match the goal's `data.size`. Root cause: `goFusedP` inlines the `litTable.lens[(bitBuf &&& 0x1FF).toNat]!.toNat` lookups directly in the `if hlit :` condition (done to dodge a termination-`omega` atom mismatch from a `let`-bound `llen`). That inlining yields a malformed `.induct`.

**Fix to try first:** give `goFusedP` a tiny non-recursive helper for the literal peek so the branch condition is over a clean atom, restoring a well-formed `.induct` *and* keeping termination happy, e.g.
```lean
@[inline] def litPeek (table : HuffTree.DecodeTable) (bitBuf : UInt64) : UInt16 × Nat :=
  ((table.syms[(bitBuf &&& 0x1FF).toNat]!), (table.lens[(bitBuf &&& 0x1FF).toNat]!.toNat))
```
and condition on its components (or whichever shape makes both `termination_by … decreasing_by omega` and `goFusedP.induct` well-formed — verify the refill-case hypothesis has no metavariable before grinding the other 9 cases). Cases are `case1`…`case10`; `with` mis-parses the fixed params as alternatives, so use bare `induction … using goFusedP.induct` + `case caseN => …`.

kim-em:
## Before/after `decompress_mbps` (decode), machine `chungus`

Wiring `goFusedP` into the production decoder. Output is **byte-identical** (`max |Δratio| = 0.000000`); native rows refreshed in-PR.

**Silesia** (12 files, single pass):

silesia decode pareto

**Canterbury** (11 files, median-of-5):

canterbury decode pareto

Native decode geomean by level:

| level | Canterbury before→after | Silesia before→after |
|------:|:------------------------|:---------------------|
| 1 | 97.9 → 100.6 (1.03×) | 116.0 → 120.9 (1.04×) |
| 2 | 103.9 → 106.6 (1.03×) | 123.5 → 128.2 (1.04×) |
| 3 | 108.8 → 111.6 (1.03×) | 130.9 → 136.7 (1.04×) |
| 4 | 110.3 → 113.2 (1.03×) | 132.8 → 138.6 (1.04×) |
| 5 | 110.1 → 113.2 (1.03×) | 138.8 → 144.2 (1.04×) |
| 6 | 111.6 → 114.9 (1.03×) | 143.2 → 149.0 (1.04×) |
| 7 | 106.8 → 114.9 (1.08×) | 144.9 → 149.9 (1.03×) |
| 8 | 112.0 → 115.4 (1.03×) | 144.9 → 150.8 (1.04×) |

**≈ +3% Canterbury / +4% Silesia**, uniform across all 8 levels — matching this PR's original paired-interleaved figure (+2.9% / +3.9%). The dashboard refresh measured a +3.97% decode geomean over 207 rows with ratio unchanged.

Caveats: AFTER is a fresh run vs the committed base-branch `latest.json` BEFORE (same machine, cross-session); the per-level uniformity on both corpora — and the median-of-5 Canterbury agreeing with single-pass Silesia — makes a noise explanation unlikely. Decode-only metric (no ratio trade-off). lean-zip native stays in its cluster (above OCaml/Go on Silesia); this is an in-cluster speedup, not a leapfrog.

Issue #2668, opened 2026-06-20
perf(decode): adopt libdeflate's decoder architecture — tracking issue


kim-em:
**Umbrella / tracking issue: re-architect the native DEFLATE decoder along libdeflate's lines. Records what libdeflate actually does (correcting #2653's premise) and links the per-phase work, split by the lean4#14053 dependency.**

## Background: why #2653 didn't pan out, and what libdeflate actually does

#2653 assumed libdeflate decodes "1-2 symbols per table lookup". Both prototypes built on that premise were net-negative (closed on #2653): a lazy runtime second-lookup was −1.7%/−2.9% (canterbury/dickens), and an eager precomputed wide 2-symbol table was −31% to −53%, dominated by per-block table-build cost.

Reading libdeflate's source (`lib/deflate_decompress.c`, `lib/decompress_template.h`) shows the premise was wrong. libdeflate decodes **one symbol per table lookup**; its speed (multiple GB/s vs our ~143 MB/s) comes from a different, composable set of techniques:

1. **Bulk word refill** — a single unaligned 8-byte load OR'd into the bit buffer (`REFILL_BITS_BRANCHLESS`). We load one byte at a time in a ~7-iteration loop.
2. **Multiple symbols per refill** — the fastloop refills once (~56 bits) then decodes up to 3 literals back-to-back, no refill between them. We refill after every symbol.
3. **Packed single-word entry** — one `u32` per entry (literal/length-base + codeword length + extra-bit count + type flags); literal fast path is a single high-bit test. We use two arrays plus separate length-base/extra lookups.
4. **Canonical O(num_syms + 2^bits) build, no Huffman tree** — counting sort + table-doubling fill, with subtables (array lookups) for long codes. We build a `HuffTree` per block, walk it per slot to build the table, and tree-walk long codes.
5. **Wider root table** — `LITLEN_TABLEBITS = 11` vs our `fastBits = 9`.

## What our bottleneck actually is (corrected after the #2669 attempt)

[...]

PR #2673, opened 2026-06-20
perf(decode): packed single-word decode-table entry

### LINKED ISSUE #2670: perf(decode): packed single-word decode-table entry (Phase 3 of #2668)
**Phase 3 of #2668. Available now — does NOT need lean4#14053.**

Fold the decode table's parallel `syms : Array UInt16` + `lens : Array UInt8` (and the separate `lengthBase`/`lengthExtra` lookups on the length path) into one packed `Array UInt64`, libdeflate-style: each entry holds symbol / length-base + codeword length + extra-bit count + a type flag, so one array read + shifts yields everything and the literal fast path is a single high-bit test.

Caution: #2650's comment records why the table was split into scalar arrays in the first place — extracting a packed field needs a shift, which forced WHNF / `Nat.div` looping in the well-founded recursions' equation compiler. Either keep the packed table behind an equivalence to the split `DecodeTable` (prove `unpack (packed i) = (syms i, lens i)` and rewrite), or confirm the shift stays inert under the relevant `whnf`.

kim-em:
This PR folds the native DEFLATE decode table's two parallel scalar arrays (`syms : Array UInt16` + `lens : Array UInt8`) into one packed `Array UInt32`, libdeflate-style: each entry holds the codeword length in the low byte and the symbol in bits 8–23, so a per-symbol decode reads one tagged-scalar array slot and extracts both fields with register shifts (`lenAt` / `symAt`) instead of two reads into two arrays / two cache lines. The packed table is kept behind an equivalence to the split projections — `buildTable_lenAt` / `buildTable_symAt` recover `(tableEntry …).2` / `(tableEntry …).1` via `unpackLen`/`unpackSym` of `packEntry` — so every decode proof transfers with a mechanical `lens[i]!`→`lenAt i`, `syms[i]!`→`symAt i` rewrite and no new `sorry`.

Two design points the [#2650](https://github.com/kim-em/lean-zip/issues/2650) comment warned about:

- **The length is the low byte deliberately.** Only the length flows into the well-founded recursions' termination measure (`cnt - len`), and extracting it must avoid a shift: a `>>>` on the stuck array read `packed[idx]!` forces `Nat.shiftRight`/`Nat.div` during the equation compiler's `whnf`, which loops on the opaque value. The low byte extracts with `toUInt8` (a width truncation, no `Nat.div`), so `unpackLen` stays inert under `whnf`; the symbol uses a shift but never enters a measure.
- **`UInt32`, not `UInt64`.** On a 64-bit target `lean_box_uint32` is a tagged scalar stored inline in the array (no allocation), like the old `UInt8`/`UInt16` arrays; `lean_box_uint64` heap-allocates a boxed word per entry, so an `Array UInt64` would pointer-chase through `lean_ctor_get` on every read, re-introducing the boxing [#2650](https://github.com/kim-em/lean-zip/issues/2650) removed. A `UInt32` holds len (8) + symbol (16) with room to spare — this is also libdeflate's `u32` entry width.

The source reads each slot through `lenAt`/`symAt`, which name the same `packed[idx]!` element; the compiler CSEs the shared read, so the fastloop literal path lowers to one `lean_array_get` plus shifts (verified in the generated C). Measured ~+10% native decode throughput on the Canterbury corpus (median of 3): alice29 160→176, lcet10 142→156, plrabn12 142→156, ptt5 399→436, sum 96→109 MB/s.

Phase 3 of [#2668](https://github.com/kim-em/lean-zip/issues/2668). Closes [#2670](https://github.com/kim-em/lean-zip/issues/2670).

🤖 Prepared with Claude Code

### DIFF
--- a/Zip/Native/Inflate.lean
+++ b/Zip/Native/Inflate.lean
@@ -154,29 +154,90 @@ where
       else if bits % 2 == 0 then go z (bits / 2) (depth + 1)
       else go o (bits / 2) (depth + 1)
 
-/-- A de-boxed fast-decode table: the `(symbol, codeLen)` of each slot split into
-    two parallel scalar arrays. Storing `Array (UInt16 × UInt8)` boxes every slot
-    as a heap pair, so each per-symbol read pointer-chases through `lean_ctor_get`
-    twice; the split keeps each field in a scalar array (`lean_array_fget` +
-    `unbox`, no chase), which is the de-box of #2650.
+/-- Pack a slot's `(symbol, codeLen)` into one `UInt32` word: the codeword length
+    in the low byte (bits 0–7), the 16-bit symbol in bits 8–23. libdeflate-style
+    single-word entry (libdeflate also uses a `u32`). The source reads the slot
+    through `lenAt`/`symAt`, which name the same `packed[idx]!` element; the
+    compiler CSEs the shared read, so a per-symbol decode lowers to **one**
+    `lean_array_get` plus register shifts (verified in the generated C for the
+    fastloop literal path), rather than two reads into the parallel `syms`/`lens`
+    arrays of #2650 (which also touched two cache lines). Even without the CSE it
+    is two reads of one tagged-scalar array, never the old two-array / two-cache-
+    line shape.
 
-    Two scalar arrays rather than one packed `UInt32` word: extracting the high
-    field of a packed word needs a shift, and the well-founded recursions `go` /
-    `goFused` would force that shift to weak head normal form during equation
-    compilation and loop on `Nat.div`. A plain array read stays stuck under
-    `whnf`, so the split keeps both the de-box and the proofs. -/
+    **`UInt32`, not `UInt64`.** On a 64-bit target `lean_box_uint32` is a tagged
+    scalar stored inline in the array (no allocation), exactly like the old
+    `Array UInt8`/`Array UInt16`; `lean_box_uint64` instead heap-allocates a boxed
+    word per entry, so an `Array UInt64` would pointer-chase through `lean_ctor_get`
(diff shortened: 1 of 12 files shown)


kim-em:
Before/after native **decode throughput** (`decompress_mbps`; machine `Linux chungus x86_64`, same as the dashboard; decode-only, levels 1–8 for the overlay; output-neutral, `max |Δratio| = 0.000000`).

The packed single-word entry lifts decode throughput consistently on both corpora, with no ratio change. AFTER (red) sits cleanly above BEFORE (grey dashed) at every level; native moves up within the language-native cluster (toward Go on Silesia, level with JS/OCaml on Canterbury).

**Silesia** (the corpus that matters most — large real-world files, single-pass but consistent across all 8 levels):

silesia decode

|--:|--:|--:|--:|--:|
| 1 | 0.347 | 123.0 | 135.6 | 1.10x |
| 4 | 0.334 | 141.9 | 156.6 | 1.10x |
| 6 | 0.332 | 152.6 | 168.7 | 1.11x |
| 8 | 0.323 | 154.9 | 170.2 | 1.10x |

**Canterbury** (median-of-5):

canterbury decode

|--:|--:|--:|--:|--:|
| 1 | 0.314 | 102.3 | 118.4 | 1.16x |
| 4 | 0.305 | 116.0 | 133.8 | 1.15x |
| 6 | 0.304 | 117.5 | 136.2 | 1.16x |
| 8 | 0.296 | 117.7 | 136.7 | 1.16x |

Geomean across all 9 levels / both corpora: **+13% decode**, ratio unchanged. The dashboard (`bench/results/latest.json` + `bench/graphs/*.svg`) is refreshed in this PR's own diff.

Issue #2675, opened 2026-06-21
perf(decode): single-pass back-reference copy via a temporary @[extern] copyWithin (lean#14158), ≈+35% decode


kim-em:
**Output back-reference copy investigation — surfaced by the post-#2670 decode profile (recorded on #2668). Separate from the #2668 symbol-decode program; this is the output-copy side.**

## Finding
On the post-#2670 dickens decode profile, the output-copy path is ~28% of decode, of which the addressable, *our-code* part is:
- `Inflate.copyLoop` — **7.4%** self-time
- `lean_byte_array_copy_slice` — **11.7%** (the byte movement copyLoop drives)

The raw `memmove`/`copy_slice` of bytes is irreducible (a back-reference must copy its bytes), but `copyLoop`'s shape looks like it copies **twice**.

## Suspected inefficiency: extract-then-append double copy
`copyLoop` (Zip/Native/Inflate.lean) for the common non-overlapping back-reference (`length ≤ distance`) is:

    buf ++ buf.extract start (start + length)

`buf.extract start (start+length)` allocates a fresh `ByteArray` and copies `length` bytes into it; then `++` appends that, copying `length` bytes again — ~**2×length** copied (plus an allocation) for a back-reference that fundamentally needs one `length`-byte copy. The overlapping/RLE branch (`buf ++ (fillDouble (buf.extract ...) length).extract 0 length`) materializes even more intermediates.

libdeflate/zlib copy a back-reference **directly into the output buffer in one pass** (a single `memcpy`/byte loop from `out[start..]` to `out[size..]`).

## What to investigate / do
1. Confirm the double-copy in the generated C (is `buf ++ buf.extract …` really two copies + an alloc, or does the compiler fuse it?).
2. For the non-overlapping case (`length ≤ distance`, the common one — source `[start, start+length)` doesn't overlap the destination `[size, size+length)` since `start+length ≤ size`): replace with a single in-place forward copy that appends `length` bytes directly from `buf[start..]`. Likely needs a `ByteArray.copyWithin`-style primitive (or careful use of an existing one) that copies from the array into its own tail; check what's available before adding anything.
3. For the overlapping/RLE case (`length > distance`): a single periodic fill rather than `extract` + `fillDouble` + `extract` + `++`.
4. Prove the new copy equal to `copyLoop` (which is already proven equal to the reference `copyLoopGo` via `copyLoop_eq_ofFn`), so the decode-correctness chain transfers.

## Payoff
[...]

PR #2676, opened 2026-06-21
perf(decode): fuse overlapping back-ref copy into one copySlice

kim-em:
This PR fuses the redundant trailing `extract` + `append` in `copyLoop`'s overlapping (RLE) branch into a single `copySlice`. The overlapping back-reference copy was `buf ++ (fillDouble seed length).extract 0 length`, which materializes a `length`-byte slice (an allocation plus a memcpy) and then appends it (a second memcpy). Because the `fillDouble` result is a distinct array from `buf`, its `length`-byte prefix copies straight into `buf`'s tail with one `copySlice` (one memcpy, no intermediate) — `fill` and `buf` are different objects, so there is no self-aliasing. The new `copySlice_prefix_eq_append` helper proves the two forms equal, so `copyLoop_eq_ofFn` and the whole decode-correctness chain carry over with a one-line proof change. Measured ~8-17% native decode speedup on RLE-heavy inputs (period 1-64, 8 MB each), correctness preserved.

The hot non-overlapping case (`length <= distance`) is deliberately left unchanged. A single-pass self-append there would need `buf.copySlice start buf buf.size length false`, with `buf` as both the borrowed source and the owned destination. That is **safe** (no use-after-free — valgrind clean) but **quadratic**: because `buf` is passed both borrowed and owned, the compiler inserts a protective `inc`, so the destination is non-exclusive inside the call and `lean_copy_sarray` duplicates the whole buffer every call (O(size) per back-reference). An efficient single pass needs a new `ByteArray.copyWithin`-style primitive in Lean core that takes the buffer as a single owned argument (a prototype measured ~28-38% decode speedup); that cannot be done in pure Lean and is out of scope here — see issue #2675 for the upstream writeup.

Refs #2675.

🤖 Prepared with Claude Code

### DIFF
--- a/Zip/Native/Inflate.lean
+++ b/Zip/Native/Inflate.lean
@@ -354,17 +354,20 @@ decreasing_by
     (a `memcpy`) instead of `length` per-byte `push`es / bounds-checks / modular
     indices. For an **overlapping** back-reference (`k = 0 ∧ length > distance`,
     the RLE case) the copy is the periodic extension of the `distance`-byte window,
-    built by `fillDouble` (a handful of memcpys) instead of `length` per-byte
-    `push`es. A partial copy (`k ≠ 0`, never produced by the decoders) falls back
-    to the per-byte `copyLoopGo`. All three are proven equal to `copyLoopGo` via
-    `copyLoop_eq_ofFn`, so every decode correctness proof is unaffected. -/
+    built by `fillDouble` (a handful of memcpys). The `length` bytes of that fill
+    are copied straight into `buf`'s tail with a single `copySlice` (one `memcpy`,
+    no trailing `extract` + `append`), since the fill is a distinct array so there
+    is no self-aliasing. A partial copy (`k ≠ 0`, never produced by the decoders)
+    falls back to the per-byte `copyLoopGo`. All three are proven equal to
+    `copyLoopGo` via `copyLoop_eq_ofFn`, so every decode correctness proof is
+    unaffected. -/
 def copyLoop (buf : ByteArray) (start distance : Nat)
     (k length : Nat)
     (hd_pos : distance > 0 := by omega) (hsd : start + distance ≤ buf.size := by omega) : ByteArray :=
   if k = 0 ∧ length ≤ distance then
     buf ++ buf.extract start (start + length)
   else if k = 0 then
-    buf ++ (fillDouble (buf.extract start (start + distance)) length).extract 0 length
+    (fillDouble (buf.extract start (start + distance)) length).copySlice 0 buf buf.size length false
   else
     copyLoopGo buf start distance k length hd_pos hsd
 
--- a/Zip/Spec/DecodeCorrect.lean
+++ b/Zip/Spec/DecodeCorrect.lean
@@ -599,6 +599,17 @@ private theorem fillDouble_get (seed : ByteArray) (length : Nat) :
         append_self_getElem! seed (i % (seed.size + seed.size)) hpos (Nat.mod_lt i (by omega)),
         Nat.mod_mod_of_dvd i ⟨2, by omega⟩]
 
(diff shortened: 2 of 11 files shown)


kim-em:
## Before/after `decompress_mbps` (machine `Linux chungus x86_64`)

**Output-neutral** (`max |Δratio| = 0.000000`) and **flat on these corpora** — this PR only touches the cold overlapping/RLE copy path, which canterbury/silesia barely exercise. The path-specific win (~8–17% native decode) is on RLE-heavy data and is not visible on the standard corpora; see the PR description and #2675 for that measurement. Before/after native curves sit exactly on top of each other; lean-zip native decode is in the slower cluster (≈ Go/JS tier on these corpora, well below the C/Rust/SIMD engines).

**Caveats:** BEFORE is a fresh merge-base measurement (the committed dashboard snapshot `a19271bf` predated #2673, so the freshness guard required re-measuring), levels 1–8 (L9 optimal-parse compress skipped for the decode overlay); Silesia is single-pass (reps=1, ±noise), Canterbury median-of-5. Same machine for before/after. The in-PR dashboard refresh (this PR's second commit) brings `latest.json` current past #2673.

silesia decode before/after

canterbury decode before/after

### silesia — native `decompress_mbps` geomean by level
|--:|--:|--:|--:|--:|
| 1 | 0.347 | 135.4 | 135.7 | 1.00x |
| 2 | 0.342 | 144.3 | 144.3 | 1.00x |
| 3 | 0.338 | 153.7 | 153.2 | 0.99x |
| 4 | 0.334 | 156.7 | 156.4 | 1.00x |
| 5 | 0.333 | 163.0 | 163.3 | 1.00x |
| 6 | 0.332 | 168.1 | 164.4 | 0.98x |
| 7 | 0.323 | 168.5 | 170.6 | 1.01x |
| 8 | 0.323 | 169.7 | 171.1 | 1.01x |

### canterbury — native `decompress_mbps` geomean by level
|--:|--:|--:|--:|--:|
| 1 | 0.314 | 118.7 | 118.3 | 1.00x |
| 2 | 0.311 | 125.8 | 125.5 | 1.00x |
| 3 | 0.309 | 132.2 | 131.0 | 0.99x |
| 4 | 0.305 | 134.5 | 133.9 | 1.00x |
| 5 | 0.305 | 134.5 | 134.3 | 1.00x |
| 6 | 0.304 | 136.5 | 136.2 | 1.00x |
| 7 | 0.297 | 135.4 | 136.0 | 1.00x |
| 8 | 0.296 | 137.1 | 137.0 | 1.00x |

The ±1–2% wobble is cross-session/single-pass noise, not the change (provably byte-identical). Verdict: **no regression**; this is a win only on the RLE path it targets.

kim-em:
Closing — abandoned after careful profiling turned up a hot-path regression.

Interleaved/paired measurement (old vs new in alternation, pinned to one core, median-of-21 per sample) shows the fused overlapping `copySlice` is **~0.6% slower on the hot non-overlapping path**, slower in 88/100 pairs (sign-test p ≈ 0), consistent across dickens/webster/samba/nci. The generated C shows why: changing `copyLoop`'s overlapping branch perturbs `length`'s refcount codegen and adds a `lean_dec(length)` to the hot *non-overlapping* branch — one extra op per back-reference. Hoisting the overlapping branch into a `@[noinline]` worker was worse (~1.1% slower, 38/40), and the extra `dec` persisted; any change to `copyLoop` seems to perturb the hot path.

So the trade is ~0.6% slower on essentially all real data (non-overlapping dominates) for +8–17% on RLE-heavy data only — a net regression for typical workloads. And `copyLoop` will be rewritten wholesale once the upstream `ByteArray.copyWithin` primitive lands (+~35%, hot path included), which supersedes this. Not worth shipping an interim regression.

Investigation and the upstream-primitive writeup remain on #2675.

PR #2683, opened 2026-06-23
perf(decode): ship verified tree-free decoder on the production path

### LINKED ISSUE #2682: perf(decode): switch production inflate onto the verified tree-free decoder (close the code-length validation gap)
**Ship the verified tree-free decoder on the production path — the actual ~14.5% decode win. Successor to #2678 (whose original framing is now stale). Part of #2668.**

## Current state (already on master)
- #2679 proved `buildTableCanonical_eq`.
- #2681 landed `Inflate.inflateTreeFree`: a **verified, well-founded (no `partial`)** canonical Huffman decoder that builds no runtime Huffman tree, with a full correctness proof in `Zip/Spec/InflateTreeFreeCorrect.lean` (no `sorry`). Measured **~1.145x faster** (canterbury 1.18x, silesia 1.11x, geomean over 23 files), byte-identical to `Inflate.inflate` on every corpus file.

But `inflateTreeFree` is a **separate entry point** — the production `Inflate.inflate` / `inflateRaw` / `inflateLoop` (and the gzip/zlib/ZIP callers) still use the tree-built decoder. This issue switches them over, so the win actually ships.

## The blocker that makes this more than a one-line swap: a correctness/security gap
#2681 proved only the **forward** direction (`inflateTreeFree_of_inflate`: `inflate = .ok out ⇒ inflateTreeFree = .ok out`). It explicitly notes the full `iff` does **not** hold:

> for invalid dynamic code lengths the tree path rejects where the tree-free path may accept

i.e. `inflateTreeFree`'s accept-set is currently a **superset** of `inflate`'s — it accepts some malformed streams `inflate` rejects. Switching production onto it as-is would make the native decoder **accept inputs it currently rejects**, breaking the `native ⊆ FFI` property the fuzz census enforces (`ZipTest/FuzzInflate.lean`) and regressing the malformed-input rejection posture (Huffman-guard suites, malformed-archive fixtures, `maxOutputSize` bomb tests). **That is a security regression and a hard blocker — it must be closed first.**

The gap is localized: the loop-level correspondence is already an `iff` (`goTreeFree_ok_iff_goFusedP`), so the only accept-set difference is in the **dynamic-tree code-length validation** — `decodeDynamicTrees` builds trees via `HuffTree.fromLengths`, which rejects (a) any length `> maxBits` ("Inflate: code length exceeds maximum") and (b) oversubscribed/over-full codes via the Kraft check ("Inflate: oversubscribed Huffman code"); the tree-free `decodeDynamicTrees_extract` extracts the length vectors but skips those rejections. (Fixed blocks use `fixedLit/DistLengths`, always valid — no gap there.) **Audit to confirm** the code-length validation is the *only* difference; do not assume.

## Deliverables

1. **Close the validation gap.** Add the same `maxBits`/Kraft code-length validity check `fromLengths` performs to the tree-free dynamic-tree setup (`decodeDynamicTrees_extract` / `inflateTreeFree`), returning the same error messages. It is computable cheaply from the per-length `count` arrays the canonical build already needs — no tree, no extra pass. Match the existing wording (see the `error-wording-catalogue` skill).
2. **Upgrade the theorem to full equality.** Prove `inflateTreeFree data = inflate data` (both `.ok` and `.error`), i.e. extend the forward `inflateLoopTreeFree_of_inflateLoop` / `inflateTreeFree_of_inflate` with the error-path correspondence the validation enables. (If a clean two-sided message-exact equality is impractical, the minimum bar is the containment that guarantees `native ⊆ FFI`: `inflateTreeFree` rejects whenever `inflate` does — but full functional equality is the goal and makes step 3 a mechanical rewrite.)
3. **Switch the production path.** Route `inflate`/`inflateRaw`/`inflateLoop` (and re-check the gzip/zlib wrappers + the ZIP entry decoder that call `inflate`) onto the tree-free decoder, rewriting the existing correctness chain (`InflateCorrect`, `InflateComplete`, `InflateRawSuffix`, `InflateLoopBounds`, `InflateBufCorrect`) through the step-2 equality so every downstream theorem transfers. No `@[implemented_by]` / no trust gap — the production def *is* the tree-free decoder, proven equal to the old spec.
4. **Perf graphs.** This is a real decode throughput change — run the `perf-pr-graphs` workflow (before/after, decode) and show Kim before merge.

## Verification
`nix-shell --run "lake build && lake exe test"` green; **`FuzzInflate` native⊆FFI census and every malformed-input/Huffman-guard suite stay green** (this is the whole point of step 1); the `InflateTable` canonical conformance test stays green; no `sorry`; decode correctness chain intact after rewiring.

## References
Forward proof + building blocks: `InflateTreeFreeCorrect.lean` (`inflateTreeFree_of_inflate`, `inflateLoopTreeFree_of_inflate Loop`, `decodeDynamicTrees_extract`, `decodeHuffmanFastBufTreeFree`, `goTreeFree_ok_iff_goFusedP`, `goTreeFreeU_eq`, `decodeSymCanon_ok_iff_decodeSym`). Validation to mirror: `HuffTree.fromLengths` in `Zip/Native/Inflate.lean`. Supersedes #2678. Part of #2668.

kim-em:
This PR routes the production decompression entry points onto the verified tree-free canonical Huffman decoder from #2681, after first closing the code-length validation gap that made its accept-set a superset of `Inflate.inflate`, so the ~14.5% decode win ships without weakening malformed-input rejection. Closes #2682. Part of #2668.

The gap (per #2681): on malformed dynamic code lengths the tree-free path skipped the `maxBits`/Kraft validation `HuffTree.fromLengths` performs, so it accepted some streams `inflate` rejects — switching production onto it as-is would widen the native accept-set versus the zlib FFI, regressing the `native ⊆ FFI` posture the fuzz census enforces. `HuffTree.validateLengths` factors that exact check (same `"Inflate: code length exceeds maximum"` / `"Inflate: oversubscribed Huffman code"` messages, computed from the lengths alone, no tree), and `decodeDynamicLengthsOnly` now runs it on the lit/dist length vectors, rejecting exactly what `decodeDynamicTrees` rejects.

With the gap closed the forward-only correspondence of #2681 becomes a two-sided accept-set equality: `inflateTreeFree data = .ok out ↔ inflate data = .ok out` (`inflateTreeFree_ok_iff`) and the start-position form `inflateRawTreeFree_ok_iff`, built from the existing forward direction and the new backward loop lemma `inflateLoop_of_inflateLoopTreeFree` (mirroring `inflateLoopTreeFree_of_inflateLoop`), itself resting on `decodeDynamicTrees_of_lengthsOnly` (the converse of `decodeDynamicTrees_extract`) and the factored `validateLengths` ↔ `fromLengths` lemmas. No `sorry`.

The production decoders `GzipDecode.decompress`, `ZlibDecode.decompress`, `decompressAuto`, and the ZIP entry decoder `readEntryData` (plus the `FuzzInflate` native census) now decode through `inflateRawTreeFree` / `inflateTreeFree`; the accept-set equality proof guarantees the census stays identical to the old decoder's — verified by the run below (native-accept/FFI-reject unchanged at 1, zero silent corruption).

Note on architecture: `Inflate.inflate` / `inflateRaw` / `inflateLoop` are kept as the verified reference, and the production callers are switched onto the tree-free decoder proven accept-set-equal to it, rather than redefining `inflate` in place. Redefining in place would break `inflate_eq` (`InflateBuf.inflate = Inflate.inflate`, a full equality that fails once the two paths differ on malformed-data error strings) and force a churn of the whole correctness chain for no verification gain; the ok-iff bridge transfers every `.ok`-based downstream theorem to the production path. (An audit confirmed no theorem references the production decoder functions directly.)

Verification: `nix-shell --run "lake build && lake exe test"` green; no `sorry`; `FuzzInflate` native⊆FFI census, malformed/Huffman-guard suites, and the `InflateTable` conformance test all stay green.

🤖 Prepared with Claude Code

### DIFF
--- a/Zip/Archive.lean
+++ b/Zip/Archive.lean
@@ -1217,8 +1217,8 @@ private def readEntryData (h : IO.FS.Handle) (entry : Entry) (label : String)
     if entry.method == 0 then pure compData
     else if entry.method == 8 then
       if useNative then
-        -- `Zip.Native.Inflate.inflate` treats `0` as "reject any non-empty
-        -- output", unlike the FFI path where `0` means unlimited. Callers
+        -- `Zip.Native.Inflate.inflateRawTreeFree` treats `0` as "reject any
+        -- non-empty output", unlike the FFI path where `0` means unlimited. Callers
         -- that opt into `maxEntrySize := 0` for unlimited FFI decompression
         -- therefore get an immediate rejection on the native backend.
         --
--- a/Zip/Archive.lean
+++ b/Zip/Archive.lean
@@ -3,6 +3,7 @@ import Zip.Checksum
 import ZipCommon.Handle
 import Zip.RawDeflate
 import Zip.Native.Inflate
+import Zip.Native.InflateTreeFree
 import Zip.Native.Crc32
 
 /-! ZIP archive construction and extraction: entry metadata, local/central headers,
@@ -1236,7 +1237,7 @@ private def readEntryData (h : IO.FS.Handle) (entry : Entry) (label : String)
         -- above `nativePresizeCap` simply keep a few of their doublings.
         let sizeHint := min (min entry.uncompressedSize.toNat maxEntrySize.toNat)
           (min (compData.size * 1100) nativePresizeCap)
-        match Zip.Native.Inflate.inflate compData maxEntrySize.toNat (sizeHint := sizeHint) with
+        match (Zip.Native.Inflate.inflateRawTreeFree compData 0 maxEntrySize.toNat sizeHint).map Prod.fst with
         | .ok data => pure data
         | .error msg => throw (IO.userError s!"zip: native inflate failed for {label}: {msg}")
       else RawDeflate.decompress compData maxEntrySize
--- a/Zip/Native/Gzip.lean
+++ b/Zip/Native/Gzip.lean
@@ -35,8 +35,8 @@ termination_by data.size - pos
     non-empty output (the inner inflate guards compare
     `output.size + len > maxOutputSize`). The outer-loop guard raises an
     `Except` error containing `"Gzip: total output exceeds maximum size"`;
-    the inner per-member `Inflate.inflateRaw` call also enforces the bound
-    and may surface `"Inflate: output exceeds maximum size"` first.
+    the inner per-member `Inflate.inflateRawTreeFree` call also enforces the
+    bound and may surface `"Inflate: output exceeds maximum size"` first.
     See `SECURITY_INVENTORY.md` *Decompression Limit Inventory*. -/
 def decompress (data : ByteArray) (maxOutputSize : Nat := 1024 * 1024 * 1024) :
     Except String ByteArray := do
@@ -132,7 +132,7 @@ namespace ZlibDecode
     Returns the decompressed data.
 
     `maxOutputSize` (default 1 GiB) is forwarded to the inner
-    `Inflate.inflateRaw`; this layer adds no separate guard. Unlike the
+    `Inflate.inflateRawTreeFree`; this layer adds no separate guard. Unlike the
     FFI path, where `maxDecompressedSize := 0` means unlimited, here `0`
     rejects any non-empty output (the inflate guards compare
     `output.size + len > maxOutputSize`). Overflow raises an `Except`
@@ -230,7 +230,7 @@ def detectFormat (data : ByteArray) : CompressFormat :=
 /-- Decompress data by auto-detecting the format (gzip, zlib, or raw deflate).
 
     `maxOutputSize` (default 1 GiB) is forwarded to whichever of
-    `GzipDecode.decompress`, `ZlibDecode.decompress`, or `Inflate.inflate`
+    `GzipDecode.decompress`, `ZlibDecode.decompress`, or `Inflate.inflateTreeFree`
     the dispatch picks based on `detectFormat`. The surfaced error
     substring depends on the dispatch: `"Gzip: total output exceeds
     maximum size"` (gzip outer guard), or `"Inflate: output exceeds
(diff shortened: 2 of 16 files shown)


kim-em:
## Decode throughput — before/after (`decompress_mbps`)

Machine `Linux chungus x86_64`. **before** = merge-base `eae020e9` binary (production decode = `Inflate.inflate`, tree-built canonical table); **after** = this branch (production decode = `inflateRawTreeFree`). Both run from the main worktree (so both see Silesia), pinned to one core, levels 1–8 (L9 optimal-parse compress skipped — decode is level-independent here). `max |Δratio| = 0.000000` — the decode is **byte-identical**, as the `inflateTreeFree_ok_iff` accept-set-equality proof guarantees.

**Canterbury** (geomean over 11 files, median-of-5):

| 1 | 0.314 | 119.8 | 127.8 | 1.07x |
| 4 | 0.305 | 136.0 | 144.6 | 1.06x |
| 6 | 0.304 | 137.9 | 148.0 | 1.07x |
| 8 | 0.296 | 138.6 | 149.3 | 1.08x |

**Silesia** (geomean over 12 files, single-pass):

| 1 | 0.347 | 137.1 | 146.7 | 1.07x |
| 4 | 0.334 | 158.2 | 168.9 | 1.07x |
| 6 | 0.332 | 170.8 | 183.3 | 1.07x |
| 8 | 0.323 | 173.4 | 185.0 | 1.07x |

A consistent **~1.07x** decode speedup across every level on both corpora (red AFTER sits clearly above gray BEFORE), output-neutral. The win is the tree-free decode path (no `fromLengths`/tree build, canonical `walkCanonical` fallback) now shipping on production. Visually, `lean-zip native` stays in the mid-pack — above OCaml/Go on Silesia, below zlib/miniz_oxide/libdeflate.

silesia decode

canterbury decode

Caveat: this is a cross-session before/after (separate binaries, same machine); Silesia is single-pass (±noise). The ~7% delta is uniform across all 16 (corpus, level) cells, well above the noise floor. @kim-em — graphs for your review before merge.

kim-em:
Superseded by #2685, which consolidates this (the gap-fix + bridge + production switch) with the follow-up rename into a single PR against master. Before/after perf graphs remain at the comment above.

PR #2684, opened 2026-06-23
refactor(decode): make Inflate.inflate itself the tree-free decoder

kim-em:
This PR makes `Inflate.inflate` / `Inflate.inflateRaw` *themselves* the tree-free production decoder (the literal API form Kim asked for as a follow-up to #2683), rather than routing each caller onto a separately-named tree-free entry point. Stacked on `issue-2682` (#2683); when that merges this rebases onto master.

The tree-free canonical decoder takes over the canonical names (defined in `Zip.Native.InflateTreeFree`), and the old tree-building implementation is kept — renamed — as the verified reference `Inflate.inflateReference` / `inflateRawReference` (`inflateLoop` and its theorems are untouched, still the reference loop). Because the canonical names now resolve to the tree-free path, every direct user, benchmark, and test of `Inflate.inflate` gets the decode win with no per-caller switch, so this reverts #2683's caller-by-caller routing (gzip/zlib/ZIP/`decompressAuto` go back to plain `inflate`/`inflateRaw`).

The verified spec tree is restated about the reference via a mechanical `Inflate.inflate`/`inflateRaw` → `inflateReference`/`inflateRawReference` rename across the proof files — each file stays self-contained, so there are no new cross-imports and no import-DAG cycles (the equivalence bridge sits at the top of the DAG and would otherwise be unusable by the early roundtrip-proof files). `Zip.Spec.InflateTreeFreeCorrect` proves the standalone bridge `inflate_ok_iff_reference` (`inflate data = .ok out ↔ inflateReference data = .ok out`) and `inflateRaw_ok_iff_reference`, so every decode-correctness / completeness / roundtrip theorem about the reference transfers to the production decoder, and the `native ⊆ FFI` posture is preserved.

`nix-shell --run "lake build && lake exe test"` green; no `sorry`; the FuzzInflate native census (now exercising the tree-free `Inflate.inflate` directly) is unchanged — native-accept/FFI-reject 1, zero silent corruption. Decode throughput is byte-for-byte the same tree-free loop as #2683 (only renamed), so the ~7% win and the dashboard refresh carry over unchanged; no separate perf graphs needed.

🤖 Prepared with Claude Code

### DIFF
--- a/Zip/Archive.lean
+++ b/Zip/Archive.lean
@@ -1237,7 +1237,7 @@ private def readEntryData (h : IO.FS.Handle) (entry : Entry) (label : String)
         -- above `nativePresizeCap` simply keep a few of their doublings.
         let sizeHint := min (min entry.uncompressedSize.toNat maxEntrySize.toNat)
           (min (compData.size * 1100) nativePresizeCap)
-        match (Zip.Native.Inflate.inflateRawTreeFree compData 0 maxEntrySize.toNat sizeHint).map Prod.fst with
+        match Zip.Native.Inflate.inflate compData maxEntrySize.toNat (sizeHint := sizeHint) with
         | .ok data => pure data
         | .error msg => throw (IO.userError s!"zip: native inflate failed for {label}: {msg}")
       else RawDeflate.decompress compData maxEntrySize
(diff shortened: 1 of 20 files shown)


kim-em:
Superseded by #2685, which consolidates the gap-fix, the bridge, and the `Inflate.inflate`-is-tree-free rename into a single PR against master.

PR #2689, opened 2026-06-23
perf(decode): compute validateLengths Kraft sum from the count array, not a per-block List

### LINKED ISSUE #2686: perf(decode): compute validateLengths Kraft sum from the count array, not a per-block List
**A self-contained decode-path allocation fix, introduced in #2685.** Landed-as-pragmatic, flagged for cleanup.

## Finding
`HuffTree.validateLengths` (`Zip/Native/Inflate.lean`) — the `maxBits`/Kraft code-length check the tree-free dynamic-block setup runs (`decodeDynamicLengthsOnly`) to keep `native ⊆ FFI` — computes the Kraft sum as:

```
let lsList := lengths.toList.map UInt8.toNat
let kraft := (lsList.filter (· != 0)).foldl (fun acc l => acc + 2 ^ (maxBits - l)) 0
```

That allocates a `List` of up to 320 `Nat`s (`toList.map`) plus a filtered list, **per dynamic Huffman block**. It was implemented this way (mirroring `fromLengths`) purely for proof tractability — but #2682 explicitly called for it to be "computable cheaply from the per-length `count` arrays the canonical build already needs — no extra pass," and this misses that.

## Idea
The Kraft sum is `Σ_{len=1}^{maxBits} count[len] · 2^(maxBits − len)`, and `countLengthsFast lengths maxBits` (the per-length histogram) is **already computed** by `buildLongDecode` / `buildTableCanonicalFast` for the very same length vector. So `validateLengths` can be an `O(maxBits)` `Array.foldl` over the count array with **zero allocation** (the `maxBits` over-bound check is already `Array.any`, no list needed). Ideally share the one `countLengthsFast` result across validate + table build rather than recomputing.

## Magnitude / risk
- Per-dynamic-block, on the header-setup path (not the inner symbol loop). Impact scales with block count; modest but pure overhead removal, and it removes a regression vs the issue's stated intent.
- Low risk. The proof surface is local: `fromLengths_eq_validate` / `validateLengths_ok_iff_fromLengths` / `validateLengths_valid` in `Zip/Spec/InflateTreeFreeCorrect.lean` would re-prove against the count-array form (the Kraft-sum-over-count = Kraft-sum-over-filtered-list identity is the one new lemma; `Huffman.Spec.countLengths` already relates the two).

## Touch points
`Zip/Native/Inflate.lean` (`validateLengths`, `countLengthsFast`); `Zip/Spec/InflateTreeFreeCorrect.lean` (the `validateLengths` ↔ `fromLengths` correspondence lemmas). Standalone — no dependency on the #2668 architecture phases.

🤖 Prepared with Claude Code

kim-em:
This PR computes the dynamic-block code-length Kraft check (`HuffTree.validateLengths`, run per dynamic Huffman block by the tree-free `decodeDynamicLengthsOnly`) from the per-length `count` histogram (`countLengthsFast` — the same array the canonical table build already needs) via a new allocation-free `O(maxBits)` fold `kraftSumFast`, replacing the per-block `lengths.toList.map`/`filter` `List` of up to 320 boxed `Nat`s that `fromLengths` allocates. The over-bound check stays a plain `Array.any`; no tree is built. This removes a per-dynamic-block allocation regression versus the intent of https://github.com/kim-em/lean-zip/issues/2682, on the header-setup path.

The `validateLengths` ↔ `fromLengths` correspondence is re-proven against the count-array form without weakening any statement: `validate_kraft_eq` certifies that the count-histogram Kraft sum equals the filtered-list Kraft sum on every length set with no over-bound length, resting on a new `Huffman.Spec.kraftSumFrom_zero_eq_foldl` factored out of the existing `kraftSumFrom_eq_kraft_foldl` (the Kraft-sum-over-count = Kraft-sum-over-filtered-list identity). `kraftSumFast` is bridged to the spec recurrence `kraftSumFrom` by `kraftSumFast_go_eq` / `kraftSumFast_eq`. `fromLengths_eq_validate` and every downstream lemma (`validateLengths_ok_iff_fromLengths`, `validateLengths_valid`, ...) keep their statements, so the proven accept-set is unchanged. Full build and `lake exe test` pass with no `sorry`s.

Sharing the single `countLengthsFast` result across validate and the table build (the issue's "ideally" follow-up) is left out: the table is built later in a different function (`decodeHuffmanFastBufTreeFree`), and threading the array across that boundary would touch verified signatures for a second cheap `O(n)` array pass. The large per-block `List` allocation — the actual regression — is gone either way.

Closes https://github.com/kim-em/lean-zip/issues/2686

🤖 Prepared with Claude Code

### DIFF
--- a/Zip/Native/Inflate.lean
+++ b/Zip/Native/Inflate.lean
@@ -110,27 +110,6 @@ def fromLengths (lengths : Array UInt8) (maxBits : Nat := 15) :
     else
       .ok (fromLengthsTree lengths maxBits)
 
-/-- The code-length validity check that `fromLengths` performs, factored out so a
-    decoder that builds no tree (the canonical tree-free path) can reject exactly
-    the malformed length sets `fromLengths` rejects, with the same error messages:
-    `"Inflate: code length exceeds maximum"` for any length `> maxBits`, and
-    `"Inflate: oversubscribed Huffman code"` when the Kraft sum overflows. Computed
-    cheaply from the lengths alone — no tree, no `count` array, no insertion pass.
-    `fromLengths = (validateLengths …).map (fun _ => fromLengthsTree …)`
-    (`Zip.Spec.InflateTreeFreeCorrect.fromLengths_eq_validate`). -/
-def validateLengths (lengths : Array UInt8) (maxBits : Nat := 15) :
-    Except String Unit :=
-  if lengths.any (fun l => l.toNat > maxBits) then
-    .error "Inflate: code length exceeds maximum"
-  else
-    let lsList := lengths.toList.map UInt8.toNat
-    let kraft := (lsList.filter (· != 0)).foldl
-      (fun acc l => acc + 2 ^ (maxBits - l)) 0
-    if kraft > 2 ^ maxBits then
-      .error "Inflate: oversubscribed Huffman code"
-    else
-      .ok ()
-
 /-- Decode one symbol from the bit reader using this Huffman tree. -/
 def decode (tree : HuffTree) (br : BitReader) :
     Except String (UInt16 × BitReader) :=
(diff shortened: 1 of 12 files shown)


kim-em:
## Decode throughput: before/after

Metric `decompress_mbps`, machine `Linux chungus x86_64` (matches the dashboard). Output-neutral: `max |Δratio| = 0.000000` — this PR removes a per-dynamic-block `List` allocation in `validateLengths` and changes no decoded bytes. BEFORE is a one-off merge-base build at #2685 (ffd9ea62), since the committed `latest.json` baseline (df20821f) predated #2685; this PR refreshes the dashboard in-place to f2d6c3a8.

**Canterbury** (geomean over 11 files, median-of-5):

|------:|------:|-------:|------:|:-------:|
| 1 | 0.314 | 127.5 | 137.4 | **1.08x** |
| 3 | 0.309 | 142.2 | 153.7 | **1.08x** |
| 6 | 0.304 | 148.0 | 160.0 | **1.08x** |
| 8 | 0.296 | 149.8 | 161.5 | **1.08x** |

**Silesia** (geomean over 12 files, single pass — ±noise, weight accordingly):

|------:|------:|-------:|------:|:-------:|
| 1 | 0.347 | 146.7 | 152.5 | **1.04x** |
| 3 | 0.338 | 166.9 | 172.9 | **1.04x** |
| 6 | 0.332 | 183.6 | 189.5 | **1.03x** |
| 8 | 0.323 | 185.9 | 195.0 | **1.05x** |

Consistent ~7-8% on Canterbury, ~3-5% on Silesia, every level, both corpora agreeing on direction. Small files spend proportionally more time in per-block header setup, so removing the allocation helps them most. Visually, native (red, AFTER) lifts clearly off the BEFORE curve (gray dashed) and on Silesia edges past Go at the low-ratio end; lean-zip stays in the mid cluster (above Go/OCaml, near JS fflate; below Zig/zlib/miniz_oxide/libdeflate), as expected for a pure-overhead removal.

canterbury decode pareto

silesia decode pareto

Issue #2671, opened 2026-06-20
perf(decode): canonical O(n) table build + subtables, retire the HuffTree (Phase 4 of #2668)


kim-em:
**Phase 4 of #2668. Available now (no lean4#14053), but heaviest proof — gate on Phases 2-3 measurements before starting.**

Build the decode table directly from the code lengths, libdeflate-style (`build_decode_table` in `lib/deflate_decompress.c`): count codeword lengths, counting-sort symbols by (length, value), then fill the table by enumerating codewords in canonical order with an incremental table-doubling `memcpy` — O(num_syms + 2^table_bits), no Huffman tree. Codes longer than `table_bits` go to one level of **subtables** (a second array lookup), not a `walkTree`.

This retires `HuffTree`/`fromLengthsTree`/`walkTree` from the fast path: no per-block tree allocation, no per-slot build walk, no pointer-chasing long-code fallback. It is also what makes a wider root table affordable (#2668 Phase 5).

The eager-table prototype on #2653 measured −31% to −53% precisely because it kept a naive per-slot tree-walk build (O(2^bits · depth)) rebuilt per block; this phase is the fix. Proof: a new table-construction correctness story — code lengths ⇒ canonical codewords ⇒ table entries, plus subtable indexing — replacing the current `tableEntry`/`buildTable` ⇔ tree-walk correspondence. This is "the heaviest proof" #2653 flagged, now well-motivated.

depends-on: #2668

kim-em:
Step-0 profile (post-#2670, dickens decode) justifies this: `walkTree` (the >9-bit tree-walk fallback) is **5.0%** self-time and `tableEntry_go` (the per-slot build walk) **2.0%**, plus per-block `HuffTree` allocation in the ~13% RC/alloc bucket — all of which the canonical build + subtables eliminate. Concrete ~7-10% target on dickens, more on smaller inputs. Cleared to proceed (heaviest proof, but now evidence-backed).

kim-em:
**Coordination with #2674 (in progress) — can run in parallel if scoped right.** This and #2674 both touch `buildTable`/the packed entry and the `goFusedP`/`goFusedPU` loop in `Zip/Native/Inflate.lean` (+ `InflateBufCorrect.lean`), and the canonical build here must *emit* the packed entry format that #2674 is changing (adding length base/extra).

[...]

PR #2690, opened 2026-06-23
perf(decode): widen the root decode table to 11 bits (Phase 5 of #2668)

### LINKED ISSUE #2672: perf(decode): widen the root decode table to ~11 bits (Phase 5 of #2668)
**Phase 5 of #2668. Available now in principle, but cheap and sensible only after Phase 4 (#2671).**

Widen the fast root table from `fastBits = 9` to ~11 (libdeflate's `LITLEN_TABLEBITS = 11`), so fewer codewords need the slow/subtable path. This is affordable only once the table build is canonical O(2^bits) (Phase 4, #2671); under today's per-slot tree-walk build it would multiply build cost (the #2653 lesson). Also widens `peekFast` to read 3 bytes (12 bits after a ≤7-bit bit-offset) and reworks the `0x1FF`/`2^9` constants threaded through `Zip/Native/Inflate.lean` + the two proof files (~129 sites).

depends-on: #2671

kim-em:
This PR widens the fast Huffman root decode table from `fastBits = 9` to `11` (libdeflate's `LITLEN_TABLEBITS`), so fewer codewords spill to the slow/subtable path. The canonical O(2^bits) table build from Phase 4 (https://github.com/kim-em/lean-zip/issues/2671) keeps the wider 2048-entry table cheap to construct, so this is now the cheap, sensible width. `peekFast` reads 3 bytes instead of 2 (an 11-bit window after a ≤7-bit bit-offset spans bits 0..17, which three bytes always cover), and the `0x1FF`/`2^9`/`512` constants threaded through `Zip/Native/Inflate.lean` and the proof files become `0x7FF`/`2^11`/`2048`. The `peekFast_testBit` and `peekFast_testBit_eof` bit-correspondence proofs gain a third byte case (`pos` / `pos+1` / `pos+2`), `peek_eq`/`peek_eq_refilled` require the buffer to hold 11 bits rather than 9, and `decodeSym_corr`/`buildTable_codeLen_le` bound table code lengths by `fastBits = 11`. The `termination_by … * 9 + cnt` refill measures keep their literal `9` (one byte refilled adds 8 bits, unrelated to the table width).

It also generalizes `go_corr` over abstract `litTable distTable : DecodeTable` (with per-index `lenAt … ≤ 11` hypotheses, exactly as `goFusedP_eq` and `goFused_eq_go` already are), instantiating with `buildTable` + `buildTable_codeLen_le` at its single call site. At the wider table the old concrete-table statement made the kernel reduce the 2048-entry `Array.ofFn` while checking the loop induction — a ~35 GB proof term that OOMs CI's 16 GB runner; abstracting the table drops the check to ~9s / 1.25 GB with no statement weakening (the concrete result is recovered by instantiation).

There is no new `sorry` and no `@[implemented_by]` trust gap; the full `lake build` and the differential conformance tests (`decodeWithTable` vs tree walk, canonical-fill vs tree build, native vs FFI fuzz) all pass, with the test depth sweeps updated to straddle the new boundary at 11.

Closes #2672.

🤖 Prepared with Claude Code

### DIFF
--- a/Zip/Native/Inflate.lean
+++ b/Zip/Native/Inflate.lean
@@ -135,8 +135,11 @@ up to 15) fall back to the tree walk. `decode` stays the canonical spec;
 `decodeWithTable_eq`) and the fast block loop `decodeHuffmanFast` is proven
 equal to `decodeHuffman` — there is no `@[implemented_by]` trust gap. -/
 
-/-- Number of bits the fast decode table indexes on (the common "fast bits"). -/
-def fastBits : Nat := 9
+/-- Number of bits the fast decode table indexes on (the common "fast bits").
+    Widened to libdeflate's `LITLEN_TABLEBITS = 11`: fewer codewords spill to the
+    slow/subtable path, and the canonical O(2^bits) table build (#2671) keeps the
+    wider table cheap to construct. -/
+def fastBits : Nat := 11
 
 /-- Walk `tree` using the low bits of `idx` (LSB first, matching `readBit`),
     for up to `fastBits` steps. Returns `(symbol, codeLen)` if a leaf is reached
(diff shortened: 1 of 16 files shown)


kim-em:
## Before/after native decode throughput (`decompress_mbps`)

Machine `Linux chungus x86_64`. BEFORE = branch merge-base at measurement time (`ffd9ea62`, #2685's tree-free decoder at `fastBits = 9`); AFTER = this branch (`fastBits = 11`). Both measured back-to-back on a quiet machine, pinned to core 2, both seeing Silesia. `max |Δratio| = 0.000000` — output-neutral (decode is byte-identical, as expected for a decode-only change).

**This is the classic cache-bound split** (cf. #2650): the wider 2048-entry table is a net win on large real-world data but slightly costs the tiny synthetic files that fit in cache regardless. Per the bench policy a **Silesia win outweighs a Canterbury loss**.

### Silesia (large real-world files) — **+7–8%**
silesia decode

| 1 | 0.347 | 148.7 | 160.0 | **1.08x** |
| 2 | 0.342 | 157.7 | 169.2 | **1.07x** |
| 3 | 0.338 | 167.6 | 179.5 | **1.07x** |
| 4 | 0.334 | 169.7 | 182.7 | **1.08x** |
| 5 | 0.333 | 177.0 | 192.0 | **1.08x** |
| 6 | 0.332 | 184.2 | 198.0 | **1.08x** |
| 7 | 0.323 | 187.0 | 200.2 | **1.07x** |
| 8 | 0.323 | 187.8 | 202.4 | **1.08x** |

### Canterbury (small synthetic files) — −1 to −4%
canterbury decode

| lvl | before MB/s | after MB/s | speedup |
|----:|------------:|-----------:|--------:|
| 1 | 129.1 | 125.4 | 0.97x |
| 4 | 146.1 | 141.2 | 0.97x |
| 8 | 150.8 | 150.1 | 1.00x |

**Read:** on Silesia the AFTER curve (red) sits clearly above BEFORE (grey), moving lean-zip native up past Go/flate toward the JS fflate cluster; on Canterbury the two curves overlap (the small loss is within the log-axis noise but real at median-of-5). The +8% is uniform across all 8 Silesia levels — a strong signal it is the fastBits widening, not single-pass noise.

**Caveats:** Silesia is single-pass (±30% per-run noise, but the cross-level uniformity is convincing); BEFORE is `ffd9ea62`, not current master `79f98bb7` — #2689 (validateLengths Kraft from the count array) landed concurrently and is a per-block header-validation change orthogonal to per-byte decode throughput, so the delta is attributable to the table widening. The dashboard (`bench/results/latest.json` + `bench/graphs`) is refreshed in this PR on top of master's current snapshot.

PR #2692, opened 2026-06-23
perf(decode): skip per-block buildSymbols when no code exceeds the fast table, share countLengthsFast

### LINKED ISSUE #2687: perf(decode): skip the per-block buildLongDecode when no code exceeds 9 bits, and share countLengthsFast with the table build
**Two cheap per-block savings on the tree-free decoder landed in #2685. Interim wins on the `walkCanonical` long-code path — partially subsumed by the subtable rewrite #2671, so worth doing only while that remains deferred.**

## Finding
`InflateBuf.decodeHuffmanFastBufTreeFree` (now the production block decoder, `Zip/Native/InflateTreeFree.lean`) builds, per block and per table (lit + dist):
1. `buildTableCanonicalFast lengths 15` — the 9-bit fast table, and
2. `buildLongDecode lengths 15` — the long-code (>9 bit) structures: `countLengthsFast` + `nextCodes` + `buildFirstIndex` + `buildSymbols` (a counting sort that allocates a `total`-element `Array UInt16`).

Two redundancies:
- **`buildLongDecode` is built eagerly even when no codeword exceeds 9 bits.** `walkCanonical` (its only consumer) is reached only for codes longer than the 9-bit table. For any block whose maximum code length ≤ 9 — common, especially the distance table — the entire counting-sort + array allocation is dead work. A guard on "does any `count[len]` with `len > 9` exist?" (one pass over the small `count` array) lets the build return a trivial `LongDecode` and skip `buildSymbols`.
- **`countLengthsFast lengths 15` is computed twice** per table — once inside `buildTableCanonicalFast`, once inside `buildLongDecode` — over the identical length vector. Compute once, thread it into both.

## Magnitude / risk
- Per-block table setup (×2 tables/block), off the inner symbol loop; impact scales with block count and is largest on streams of many small dynamic blocks. Pure overhead removal.
- Low–moderate proof churn: `buildLongDecode_placement` / the `walkCanonical_*` correctness lemmas in `Zip/Spec/InflateTreeFreeCorrect.lean` would need the "skipped ⇒ no long code is ever looked up" branch, and `decodeHuffmanFastBufTreeFree_ok_iff` re-threaded through the shared `countLengthsFast`.

## Relationship to the architecture work
This is a micro-optimization on the **currently-landed** tree-free build. #2671 (Phase 4 of #2668) would replace the `walkCanonical` bit-by-bit fallback with one level of **subtables**, which retires `buildLongDecode` entirely — so the "skip buildLongDecode" half is subsumed there (though "skip subtable build when there are no long codes" still applies), and the `countLengthsFast`-dedup half is good hygiene either way. Do this only if #2671 stays deferred; otherwise fold the count-dedup into #2671. Related: #2668.

## Touch points
`Zip/Native/InflateTreeFree.lean` (`buildLongDecode`, `decodeHuffmanFastBufTreeFree`); `Zip/Native/Inflate.lean` (`countLengthsFast`, `buildTableCanonicalFast`); `Zip/Spec/InflateTreeFreeCorrect.lean` (placement / `walkCanonical` / `decodeHuffmanFastBufTreeFree_ok_iff`).

🤖 Prepared with Claude Code

kim-em:
This PR builds the per-block long-code structures only when a codeword actually exceeds the `fastBits`-wide (11-bit) fast table, and shares one length-histogram pass between the fast-table build and the long-code build. `decodeHuffmanFastBufTreeFree` now computes `countLengthsFast` once per table and threads it into both `buildTableCanonicalFastWithCount` and `buildLongDecodeWithCount`, removing the duplicate histogram pass. When no codeword exceeds the fast table (`hasLongCode = false` — common, especially for the distance table), `buildLongDecodeWithCount` skips the `buildSymbols` counting sort and its `total`-element allocation entirely, leaving `symbols` empty; the `walkCanonical` fallback is then dead, since every codeword resolves in the fast table.

Correctness is preserved against the verified reference (`inflateReference` / `inflateRawReference`), so the two-sided `inflate_ok_iff_reference` and every downstream theorem still transfer. The new `walkCanonical_dead_of_no_long` shows that when no long code exists, the fast table's length at the buffer index equals any spelled codeword's length, contradicting the `walkCanonical` fallback guard (`len = 0` or `len > cnt`), so the empty `symbols` array is never read — including at end-of-stream / insufficient-bits. `goTreeFree_ok_iff_goFusedP` is generalized over the long-code structure via a per-table `decodeSymCanon`/`decodeSym` agreement hypothesis, and `decodeHuffmanFastBufTreeFree_ok_iff` is re-threaded through the shared histogram and the new `decodeSymCanon_withCount_ok_iff_decodeSym`.

This is a micro-optimization on the currently-landed tree-free build (per the issue, done while the subtable rewrite #2671 stays deferred). It removes pure per-block setup overhead off the inner symbol loop; impact scales with block count and is largest on streams of many small dynamic blocks.

Closes #2687.

🤖 Prepared with Claude Code

### DIFF
--- a/Zip/Native/Inflate.lean
+++ b/Zip/Native/Inflate.lean
@@ -410,6 +410,17 @@ def buildTableCanonicalFast (lengths : Array UInt8) (maxBits : Nat := 15) : Deco
     let nextCode := nextCodesFast count maxBits
     buildCanonicalLoop lengths nextCode 0 (Array.replicate (2 ^ fastBits) (packEntry 0 0))
 
+/-- `buildTableCanonicalFast` taking a precomputed length histogram, so a per-block
+    decoder building both the fast table and the long-code structures can share a
+    single `countLengthsFast` pass over the length vector. Definitionally equal to
+    `buildTableCanonicalFast lengths maxBits` when `count = countLengthsFast lengths
+    maxBits` (`buildTableCanonicalFastWithCount_eq`). -/
+def buildTableCanonicalFastWithCount (lengths : Array UInt8) (count : Array Nat)
+    (maxBits : Nat := 15) : DecodeTable where
+  packed :=
+    let nextCode := nextCodesFast count maxBits
+    buildCanonicalLoop lengths nextCode 0 (Array.replicate (2 ^ fastBits) (packEntry 0 0))
+
 /-- Bits remaining in the reader from its current `(pos, bitOff)`. -/
 def bitsAvail (br : BitReader) : Nat :=
   if br.pos ≥ br.data.size then 0 else (br.data.size - br.pos) * 8 - br.bitOff
--- a/Zip/Native/InflateTreeFree.lean
+++ b/Zip/Native/InflateTreeFree.lean
@@ -78,6 +78,38 @@ def buildLongDecode (lengths : Array UInt8) (maxBits : Nat) : LongDecode :=
   let symbols := buildSymbols lengths maxBits total firstIndex
   { count, firstCode, firstIndex, symbols }
 
+/-- Does any codeword exceed the `fastBits`-wide fast table — i.e. can the
+    `walkCanonical` long-code fallback ever fire productively? `count[len] > 0` for
+    some `fastBits < len ≤ maxBits`. When this is `false`, every codeword resolves
+    in the fast table, so the long-code counting sort (`buildSymbols`, allocating a
+    `total`-element array) is dead work. One pass over the small `count` array. -/
+def hasLongCode (count : Array Nat) (maxBits : Nat) : Bool :=
+  go count maxBits (fastBits + 1)
+where
+  go (count : Array Nat) (maxBits len : Nat) : Bool :=
+    if len > maxBits then false
+    else if count[len]! > 0 then true
+    else go count maxBits (len + 1)
+  termination_by maxBits + 1 - len
+
+/-- `buildLongDecode` sharing a precomputed length histogram, and skipping the
+    `buildSymbols` counting sort entirely when no codeword exceeds the fast table
+    (`hasLongCode count maxBits = false`). In that case `walkCanonical` is never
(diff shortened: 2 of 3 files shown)


kim-em:
_(Corrected — the graph first posted here was contaminated: a concurrent benchmark run from another worktree on the shared machine overwrote the shared `/tmp` PNG, so it showed a different PR's decode win. Re-measured cleanly into private paths with an interleaved paired design.)_

**Native decode throughput (`decompress_mbps`), machine `Linux chungus x86_64`.** BEFORE = merge-base `48364a8c` binary, AFTER = this branch. Decode-only levels 1–8.

**This PR is decode-neutral.** Output-neutral (`max |Δratio| = 0.000000` — accepts/produces identically to the verified reference), and the interleaved paired after/before ratio (3 reps, pinned core, common-mode cancelled) is within noise:

| corpus | paired after/before | 95% CI |
|---|---|---|
| all | 1.0021 | [0.9987, 1.0054] |
| canterbury | 1.0041 | [0.9982, 1.0100] |
| silesia | 1.0002 | [0.9966, 1.0037] |

Both CIs span 1.0. This is expected: the change removes per-block decode *setup* work (the `buildSymbols` counting sort + its allocation, and a duplicate histogram pass), which is below measurement noise on Canterbury/Silesia (few large blocks per stream). The win is for streams of many small dynamic blocks, which these corpora don't exercise — so this is a correctness-preserving cleanup with no throughput regression.

silesia decode

canterbury decode

Before/after curves coincide on both corpora. Visual read: lean-zip native sits in the mid cluster — above Go/OCaml, around JS fflate, below Zig std.flate, zlib, miniz_oxide, libdeflate — unchanged before vs after. Dashboard (`bench/results/latest.json` + `bench/graphs/*.svg`) refreshed in-PR.

PR #2693, opened 2026-06-23
perf: single-pass back-reference copy via @[extern] ByteArray.copyWithin

### LINKED ISSUE #2675: perf(decode): single-pass back-reference copy via a temporary @[extern] copyWithin (lean#14158), ≈+35% decode

kim-em:
This PR replaces the non-overlapping LZ77 back-reference copy's `buf ++ buf.extract start (start + length)` — two memcpys plus an allocation — with a single-pass `ByteArray.copyWithin`, taking ~+26-35% off native DEFLATE decode throughput on the standard corpora. `copyWithin` is a project-local stopgap for the upstream Lean-core primitive proposed in [lean#14158 `feat: add ByteArray.copyWithin`](https://github.com/leanprover/lean4/pull/14158), and is a temporary, owner-approved exception to the no-`@[extern]` rule; its C symbol is namespaced `lean_zip_byte_array_copy_within` so it will not clash with core's `lean_byte_array_copy_within` after the toolchain bump, and migration once lean#14158 lands is to delete the shim and use core directly.

The primitive's reference body is exactly the expression `copyLoop`'s non-overlapping branch already computed (`a ++ a.extract srcOff (srcOff + len)`), so it is a pure runtime swap with identical semantics: `copyLoop_eq_ofFn` and the whole decode-correctness chain transfer with a one-line `unfold`, with no `sorry` and no proof-shaped trust gap beyond the `@[extern]` itself. The C does the append in one `memcpy` with no intermediate allocation — `buf` is a single owned argument, so the caller emits no protective `inc`, the buffer stays exclusive, and the grow is in-place/amortized (a self-`copySlice`, by contrast, is safe but quadratic). The overlapping/RLE branch stays on `fillDouble` and is out of scope — `copyWithin` is the non-overlapping primitive only.

Measured native decode (decode-only, `decompress_mbps`, machine matches the dashboard, both sides measured back-to-back this session on a pinned core): +31-35% on Canterbury (median-of-5), +26-33% on Silesia (single-pass), uniform across levels, output-neutral (`max |Δratio| = 0.000000`). FuzzInflate confirms byte-identical native-vs-FFI output. The dashboard is refreshed in-PR. Before/after graphs are posted as a PR comment.

Closes #2675.

🤖 Prepared with Claude Code

### DIFF
--- a/Zip/Native/CopyWithin.lean
+++ b/Zip/Native/CopyWithin.lean
@@ -1,0 +1,41 @@
+/-!
+  Single-pass self-append primitive for the DEFLATE back-reference copy.
+
+  `ByteArray.copyWithin a srcOff len` appends the slice `a[srcOff, srcOff + len)`
+  to `a`'s own tail. The reference body — and the trusted specification of the
+  `@[extern]` — is literally
+
+      a ++ a.extract srcOff (srcOff + len)
+
+  which is exactly the expression `Inflate.copyLoop`'s non-overlapping branch
+  already computes. The C implementation (`c/copy_within_ffi.c`) does it in one
+  `memcpy` with no intermediate allocation, where today the pure-Lean
+  `++ extract` is two `memcpy`s plus an allocation. Measured ~+28–38% native
+  decode throughput on the standard corpora (issue #2675).
+
+  `extract` is total and clamps its upper bound to `a.size`, so this is the
+  **non-overlapping** path only (`srcOff + len ≤ a.size`); the C clamps `len` to
+  match, so source and destination never overlap and the body stays a faithful
+  model. The overlapping / RLE case (`len > distance`) stays in pure Lean
+  (`fillDouble`) and does not use this primitive.
+
+  This is a project-local stopgap for the upstream Lean-core primitive proposed
+  in lean#14158 (`feat: add ByteArray.copyWithin`); the owner has approved it as
+  a temporary exception to the no-`@[extern]` rule. The C symbol is namespaced
+  `lean_zip_*` so it does not clash with core's `lean_byte_array_copy_within`
+  after the toolchain bump. Migration once lean#14158 lands: delete this file and
+  `c/copy_within_ffi.c`, and use core's `ByteArray.copyWithin` (its `@[extern]`
+  and reference body are identical, so callers and proofs are unchanged).
+-/
+
+namespace ByteArray
+
+/-- Append the slice `a[srcOff, srcOff + len)` to `a`'s own tail in one pass.
+    Non-overlapping only: intended for `srcOff + len ≤ a.size`. The reference
+    body `a ++ a.extract srcOff (srcOff + len)` is the trusted specification of
+    the `@[extern]`; `extract` clamps, and the C clamps `len` to match. -/
+@[extern "lean_zip_byte_array_copy_within"]
+def copyWithin (a : ByteArray) (srcOff len : Nat) : ByteArray :=
+  a ++ a.extract srcOff (srcOff + len)
+
+end ByteArray
--- a/Zip/Native/Inflate.lean
+++ b/Zip/Native/Inflate.lean
@@ -1,5 +1,6 @@
 import ZipCommon.Spec.BitReaderInvariant
 import Zip.Spec.Huffman
+import Zip.Native.CopyWithin
 
 /-!
   Pure Lean DEFLATE decompressor (RFC 1951).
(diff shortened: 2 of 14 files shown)


kim-em:
Before/after native decode throughput (`decompress_mbps`; machine `Linux chungus x86_64`, matches the dashboard; both sides measured back-to-back this session on a pinned core; BEFORE = merge-base `48364a8c`, AFTER = `8099528f`; levels 1-8, L9 optimal-parse compress skipped as irrelevant to decode; **output-neutral, `max |Δratio| = 0.000000`**).

Silesia (geomean over 12 files, single-pass):

silesia decode before/after

Canterbury (geomean over 11 files, median-of-5):

canterbury decode before/after

| level | Canterbury before→after | Silesia before→after |
|---|---|---|
| 1 | 136.0 → 182.2 (1.34×) | 168.3 → 222.7 (1.32×) |
| 2 | 142.5 → 192.4 (1.35×) | 178.0 → 236.3 (1.33×) |
| 3 | 150.7 → 202.0 (1.34×) | 189.8 → 249.7 (1.32×) |
| 4 | 153.0 → 203.1 (1.33×) | 194.1 → 254.3 (1.31×) |
| 5 | 157.1 → 205.7 (1.31×) | 203.9 → 259.8 (1.27×) |
| 6 | 160.7 → 210.4 (1.31×) | 210.4 → 268.3 (1.28×) |
| 7 | 161.8 → 213.1 (1.32×) | 214.3 → 270.8 (1.26×) |
| 8 | 163.7 → 215.1 (1.31×) | 215.4 → 271.7 (1.26×) |

Visually the change lifts lean-zip native decode out of the JS-fflate / OCaml / Go tier up into nearly the Zig std.flate tier; it stays below the C/Rust/SIMD cluster (libdeflate, miniz_oxide, zlib), as expected for a verified pure-Lean decoder. The dashboard (`bench/results/latest.json` + `bench/graphs/*.svg`) is refreshed in this PR at commit `abcabf9e`.

PR #2702, opened 2026-06-23
perf(matcher): window-relative circular Array Nat prev buffer

### LINKED ISSUE #2695: perf(explore): window-relative circular UInt32 prev buffer — cut O(filesize) chain memory, fix cache thrash
The LZ77 hash-chain `prev` array is allocated at full input size and boxed: `Array.replicate data.size data.size` (`Array Nat`), e.g. `Zip/Native/Deflate.lean:730` (`lz77Chain`), `:936` (`lz77ChainIter`), and the lazy/packed twins. For a 50 MB Silesia member that is a 50M-entry array (~400 MB aux), accessed at scattered positions during each chain walk, so deep walks miss cache. The L9 DP candidate cache threads the same full-size sentinel chain across regions (`Zip/Native/DeflateParse.lean:176-196`), so it has the same footprint.

Replace `prev` with a fixed 32768-entry circular buffer of `UInt32`, indexed by `pos &&& 0x7FFF`, storing absolute positions (files are < 4 GiB). This is zlib's scheme. Memory drops from O(filesize) to 128 KiB (fits L2), and the element type de-boxes (no tagged-Nat indirection). Expected: a broad compress speedup at every level, largest on Silesia (native L6 33.5 vs Go 48 MB/s) and on any deep walk, plus the removal of a real memory-scaling defect.

## Why it is correct / proof-safe

The matcher's correctness is via validity contracts (`lzMatch_{encodable,empty,resolves}`), not equality to a reference matcher. `Zip/Spec/LZ77ChainCorrect.lean` states it explicitly: "validity is re-established at emission by `countMatch` + the explicit window guards, so the `prev`/`hashTable` contents never enter the proof." Termination of the walk is `termination_by fuel` (`Deflate.lean` chainWalk), not chain monotonicity, so a circular buffer with stale/overwritten entries still terminates and still emits only `countMatch`-verified, window-guarded matches. The `pos - cand ≤ windowSize` guard already discards entries older than the window, which is exactly what makes the circular reuse sound.

## Scope

Each matcher variant has a boxed reference + packed twin + lockstep equality to re-establish:

- `lz77Chain` / `lz77ChainIter` / `lz77ChainLazy` / `lz77ChainLazyIter` and the packed `*P` twins in `Zip/Native/Deflate.lean` — change `prev` allocation, element type, and the read (`prev[cand]`) / write (`guardedSet prev pos head`) to circular `UInt32` indexing.
- The DP candidate cache `buildCache` in `Zip/Native/DeflateParse.lean` (same change; verify cross-region distances still reach 32 KiB back, which they do because the circular size equals the window).
- Re-prove the boxed=packed lockstep equalities (`chainWalkPacked_eq` and friends in `Zip/Spec/LZ77ChainCorrect.lean` / `LZ77PackedCorrect.lean`); the validity contracts should transfer unchanged.

## Suggested staging

So each commit compiles + proves:

1. Element type only: `Array Nat` to `Array UInt32`, still full size. Smaller proof delta (de-box win, ~2x memory). Land first.
2. Circular indexing: full size to 32768, index `pos &&& 0x7FFF`. The bigger win. If step 2's proof churn is too large, step 1 alone is a worthwhile standalone improvement and a safe fallback.

## Acceptance / measurement

- No ratio regression at any level on Canterbury + Silesia (the validity-contract structure guarantees correctness; ratio should be byte-identical since the same in-window candidates are reachable).
- Compress throughput: `bench/run.sh` (or `lake exe bench-report`) geomeans before/after; expect Silesia L4-L6 to move toward Go, and peak-RSS to drop sharply on large members.
- This is a `perf:` change to compress, so the `perf-pr-graphs` skill is REQUIRED before merge (Kim must see before/after speed-vs-ratio graphs).

Context: filed from a fresh-eyes performance reconciliation against master (bench snapshot 2026-06-23). Most of that analysis was already superseded by shipped decode work (#2690/#2685) and the in-flight L9 DP-parser effort (#2638/#2640/#2641); this `prev`-buffer change is the remaining novel, independent item.

kim-em:
This PR replaces the LZ77 hash-chain `prev` link array — allocated at full input size, so O(filesize) auxiliary memory (≈400 MB on a 50 MB Silesia member) accessed at scattered positions during each chain walk — with a circular ring of `min(chainWinSize, data.size)` entries (`chainWinSize` = 32768 = the 32 KiB DEFLATE window) indexed by `pos &&& 0x7FFF` (zlib's scheme). Chain memory drops to O(min(window, filesize)) — never larger than the old buffer, 256 KiB on any input ≥ the window — which fits L2 and removes the cache thrash. It applies to every matcher variant (`lz77Chain`/`lz77ChainIter`/`lz77ChainLazy`/`lz77ChainLazyIter`, the packed `*P` twins) and the L9 DP candidate cache (`chainWalkAll`/`buildCache`).

The ring stays `Array Nat`. An earlier `UInt32` de-box was measured and rejected: small positions are already stored inline-unboxed in `Array Nat`, so a `UInt32` element type only adds `.toNat`/`.toUInt32` conversions to the hot chain walk — a net ~5% tax (paired Silesia 0.95x) with no offsetting cache win. Sizing the ring to `min(window, filesize)` rather than always 32768 keeps sub-window inputs allocation-identical to the old buffer (for them `pos &&& 0x7FFF = pos`, no wrap), so tiny files pay no over-allocation tax.

Output is byte-identical to master at every level on both corpora (`max |Δratio| = 0`); positions within a window of size ≤ 32768 have distinct `&&& 0x7FFF` indices, so every in-window chain link matches the full-size buffer. Measured paired (core-pinned, vs the pre-change matcher): **Silesia +2.1%** `[+1.7%, +2.4%]` (L1 +4.6%, L2 +3.7%, deeper tapering to neutral), **Canterbury +0.8%** `[+0.1%, +1.4%]` (neutral, no size-dependent penalty). Before/after speed-vs-ratio graphs and the full paired tables are in the PR comments.

The `prev` contents are a pure matching heuristic — validity is re-established at emission by `countMatch` plus the window guards (`matchPos < pos`, `pos − matchPos ≤ windowSize`), and chain-walk termination is `termination_by fuel` — so the ring never enters a correctness proof. The proven-bounds reads discharge `cand &&& 0x7FFF < prev.size` from `winMask_lt` (`< chainWinSize`) and `Nat.and_le_left` (`≤ cand < data.size`), i.e. `< min(chainWinSize, data.size) ≤ prev.size`, under the runtime guard `min chainWinSize data.size ≤ prev.size`; the boxed=fast=packed lockstep equalities carry through with only the index change.

Closes #2695

🤖 Prepared with Claude Code

### DIFF
--- a/Zip/Native/Deflate.lean
+++ b/Zip/Native/Deflate.lean
@@ -674,21 +674,22 @@ where
 /-- Per-position chain-head insertion with one runtime bounds check (Wave 3
     Step 0.2, same pattern as `chainWalkGuarded`/`updateHashesGuarded` below):
     read the old head of bucket `h`, point the bucket at `pos`, and link
-    `prev[pos]` to the old head. The single guard discharges all three accesses
+    `prev[pos &&& 0x7FFF]` (the circular slot) to the old head. The single guard
+    discharges all three accesses
     statically; the fallback keeps the original panic-checked operations and is
     dead in practice — `hashTable`/`prev` sizes are fixed at allocation.
     Provably equal to the panic-checked sequence (`headInsertGuarded_eq` in
     `LZ77ChainCorrect`). Superseded in the mainLoops by `headProbeGuarded` +
     two `guardedSet`s (Wave 5 de-boxing — this tuple return allocated a heap
     constructor per position); kept with its `_eq` lemma. -/
-@[inline] def headInsertGuarded (hashTable prev : Array Nat) (h pos : Nat) :
+@[inline] def headInsertGuarded (hashTable : Array Nat) (prev : Array Nat) (h pos : Nat) :
     Nat × Array Nat × Array Nat :=
-  if hg : h < hashTable.size ∧ pos < prev.size then
+  if hg : h < hashTable.size ∧ (pos &&& 0x7FFF) < prev.size then
     let head := hashTable[h]'hg.1
-    (head, hashTable.set h pos hg.1, prev.set pos head hg.2)
+    (head, hashTable.set h pos hg.1, prev.set (pos &&& 0x7FFF) head hg.2)
   else
     let head := hashTable[h]!
-    (head, hashTable.set! h pos, prev.set! pos head)
+    (head, hashTable.set! h pos, prev.set! (pos &&& 0x7FFF) head)
 
 /-- Single guarded chain-head probe (for the lazy matcher's lookahead position,
     which reads a bucket head without inserting): one runtime bounds check,
@@ -703,8 +704,28 @@ where
     `LZ77ChainCorrect`). The mainLoops do their per-position head insertion as
     `headProbeGuarded` + two `guardedSet`s — three single-value steps in place
     of `headInsertGuarded`'s tuple, so nothing is heap-allocated per position. -/
-@[inline] def guardedSet (a : Array Nat) (i v : Nat) : Array Nat :=
+@[inline] def guardedSet {α : Type} (a : Array α) (i : Nat) (v : α) : Array α :=
   if h : i < a.size then a.set i v h else a.set! i v
+
+/-- The circular `prev`-buffer window: a fixed 32768-entry (= 32 KiB DEFLATE
+    window) ring of absolute positions, indexed by the low 15 bits of a
+    position (`pos &&& 0x7FFF`, zlib's scheme). Memory is O(window) not
+    O(filesize). With `windowSize ≤ 32768` the in-window guard
+    `pos - cand ≤ windowSize` discards any candidate a stale slot could surface,
+    and the one aliasing pair inside the window (`pos` and `pos - 32768`) only
+    perturbs the heuristic, never validity — the `prev` contents never enter
+    the proof (see `lz77Chain`). The ring stores absolute positions as `Nat`
+    (small values stay inline-unboxed, so no boxing or conversion in the hot
+    walk); only the shrink from O(filesize) to a window-sized ring matters for
+    speed (it fits L2), which is why this is a `Nat` ring rather than a de-boxed
+    `UInt32` one (the `UInt32` conversions measured as a net tax). -/
+def chainWinSize : Nat := 32768
+
+/-- A circular `prev`-buffer index `n &&& 0x7FFF` is always `< 32768`, so a
+    `chainWinSize`-entry buffer indexes every masked position in bounds. -/
+theorem winMask_lt (n : Nat) : (n &&& 0x7FFF) < chainWinSize := by
+  have := @Nat.and_le_right n 0x7FFF
+  unfold chainWinSize; omega
 
 /-- Greedy LZ77 with bounded-depth hash chains: at each position, walk the
     `prev` chain from the bucket head up to `maxChain` candidates and keep the
@@ -715,10 +736,22 @@ where
     re-established at the moment of emission by `countMatch` (verifies the bytes)
     plus the explicit `matchPos < pos` / `pos - matchPos ≤ windowSize` guards, so
     the `prev`/`hashTable` array contents never enter the correctness proof.
-    Buckets and `prev` are initialised to the sentinel `data.size` (≥ every real
+    `prev` is a fixed `chainWinSize` (32768 = 32 KiB window) circular buffer of
+    de-boxed `UInt32` absolute positions, indexed by `pos &&& 0x7FFF` (zlib's
+    scheme): memory is O(window) not O(filesize) and the buffer fits L2. Buckets
+    and `prev` are initialised to the sentinel `data.size` (≥ every real
     position), so an unset chain link fails the `cand < pos` guard and stops the
-    walk — no separate validity bitmap needed. Reuses `lz77Greedy`'s
-    `hash3`/`countMatch`/`trailing` helpers (and their proofs). -/
+    walk — no separate validity bitmap needed. Circular reuse stays sound: for
+    every candidate strictly inside the window (`pos - cand < 32768`) the slot
+    `cand &&& 0x7FFF` differs from `pos`'s own slot and from every other
+    in-window position's, so the link is exactly the full-size buffer's link;
+    a *stale* slot can only hold a position older than the window, discarded by
+    `pos - cand ≤ windowSize`; and at the exact boundary `cand = pos - 32768`
+    the slot was just overwritten by `pos`'s own insertion, so the read may
+    revisit the current bucket head or form a short cycle — harmless, because
+    the walk is fuel-bounded and `countMatch` + the window guard re-verify every
+    emitted match. Reuses `lz77Greedy`'s `hash3`/`countMatch`/`trailing` helpers
+    (and their proofs). -/
 def lz77Chain (data : ByteArray) (maxChain : Nat) (windowSize : Nat := 32768)
     (insertCap : Nat := 1000000000) :
     Array LZ77Token :=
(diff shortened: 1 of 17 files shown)


kim-em:
## Before/after `compress_mbps` (machine `Linux chungus x86_64`)

Output is **byte-identical** to master: `max |Δratio| = 0.000000` at every level on both corpora, and all 99 Canterbury (corpus × level) compressed sizes match exactly. So this is a pure memory/representation change — the question is only throughput.

**The headline result: this is throughput-neutral, not the speedup the issue predicted.** It is primarily a **memory** change — chain aux drops O(filesize) → 128 KiB.

silesia compress pareto
canterbury compress pareto

The before/after native (red) curves sit on top of each other in both graphs — native stays in the same cluster relative to the other languages.

### Silesia — controlled interleaved paired measurement (6 pairs, core-pinned, L1-6)

The cross-session single-pass Silesia graph is too noisy (±30%) to read a small delta, so I measured both binaries back-to-back 6×; paired geomean of after/before with 95% CI on the log-ratios:

| level | n | geomean | 95% CI | verdict |
|------:|--:|--------:|:-------|:--------|
| 1 | 72 | 1.039x | [0.994, 1.085] | within noise |
| 2 | 72 | 1.031x | [0.996, 1.068] | within noise |
| 3 | 72 | 0.991x | [0.964, 1.019] | within noise |
| 4 | 72 | 1.002x | [0.979, 1.026] | within noise |
| 5 | 72 | 0.999x | [0.977, 1.022] | within noise |
| 6 | 72 | 0.990x | [0.963, 1.018] | within noise |
| **ALL** | 432 | **1.009x** | **[0.996, 1.022]** | **within noise** |

Silesia is statistically indistinguishable from no change (>1.0 = faster).

### Canterbury — median-of-5 geomean by level

|------:|------:|------------:|-----------:|--------:|
| 1 | 0.314 | 33.4 | 32.2 | 0.96x |
| 4 | 0.305 | 26.4 | 25.1 | 0.95x |
| 6 | 0.304 | 24.6 | 23.3 | 0.95x |
| 9 | 0.284 | 1.3 | 1.2 | 0.95x |

Canterbury (small files that fit cache regardless) shows a consistent ~4-5% **regression** — the per-access `&&& 0x7FFF` mask + `UInt32`↔`Nat` conversions are a small constant tax with no offsetting cache benefit on inputs that already fit cache.

### Read

- **Correctness/memory**: byte-identical output, proof-safe, and the O(filesize)→128 KiB chain-aux reduction is real (≈400 MB → 128 KiB on a 50 MB member; removes an allocation/RSS scaling defect).
- **Speed**: the issue's thesis (broad speedup, largest on Silesia) is **not** borne out — Silesia is neutral, small files regress slightly. The prev-array access is amortized against `countMatch`'s byte loop, so shrinking it does not move throughput.

This is a memory-scaling fix that is speed-neutral on large data and ~4-5% slower on small data, not a perf win. Flagging for a merge decision rather than auto-merging.

kim-em:
## Final design: circular `Array Nat` ring (supersedes the UInt32 variant above)

Reworked after measurement showed the `UInt32` de-box was a net tax. Numbers below are vs the pre-change full-size `Array Nat` matcher, machine `Linux chungus x86_64`. Output **byte-identical** (`max |Δratio| = 0` at every level on both corpora).

silesia compress pareto
canterbury compress pareto

### Design comparison (paired, core-pinned, Silesia L1-6 geomean vs master)

| variant | Silesia | memory |
|:--------|:--------|:-------|
| `UInt32` full-size (de-box only) | 0.951x [0.939, 0.963] — **−5% regression** | O(filesize), halved |
| `UInt32` ring | 1.009x [0.996, 1.022] — neutral | O(1) 128 KiB |
| **`Nat` ring (this PR)** | **1.016x [1.008, 1.024] — +1.6% real** | O(1) 256 KiB |

The de-box is a tax (small `Nat`s are already inline-unboxed; the conversions are pure overhead); the ring's cache benefit is the win; keeping `Nat` lets it show as a net speedup.

### Paired verdict (core-pinned, 95% CI on log-ratios; >1.0 = faster)

**Silesia (6 pairs, L1-6)** — the large-file, cache-bound corpus that matters most:

| level | geomean | 95% CI |
|------:|--------:|:-------|
| 1 | 1.047x | [1.026, 1.069] (real) |
| 2 | 1.033x | [1.013, 1.053] (real) |
| 3 | 1.017x | [0.997, 1.038] |
| 4 | 1.010x | [1.000, 1.020] |
| 5 | 0.995x | [0.976, 1.014] |
| 6 | 0.996x | [0.979, 1.013] |
| **ALL** | **1.016x** | **[1.008, 1.024] (real)** |

**Canterbury (5 pairs, L1-6)** — small synthetic files that fit cache regardless:

| level | geomean | 95% CI |
|------:|--------:|:-------|
| **ALL** | **0.985x** | **[0.982, 0.988] (real −1.5%)** |

The shallow Silesia levels win most (+3-5%): they do less `countMatch` byte work per chain step, so the prev-access cache behaviour is a larger fraction. Deeper levels amortize it. The Canterbury regression is the per-access `&&& 0x7FFF` mask with no offsetting cache benefit on files that already fit cache.

Net: a real speedup on the corpus that matters + an O(filesize)→256 KiB memory-scaling fix, byte-identical and proof-safe.

kim-em:
## Update: ring right-sized to `min(window, filesize)` — now a win on **both** corpora

The earlier ring regressed sub-32 KB files (over-allocation: a tiny file allocated a full 32768 ring instead of `data.size`). A per-file breakdown showed a clean size crossover at 32768. Fix folded in: allocate `min(chainWinSize, data.size)` — sub-window inputs are now allocation-identical to the old buffer and `pos &&& 0x7FFF = pos` (no wrap), so they pay no tax; inputs ≥ window keep the full ring and the cache win. Output stays byte-identical (`max |Δratio| = 0`).

### Paired verdict (core-pinned, 95% CI on log-ratios; >1.0 = faster, vs master)

**Silesia (6 pairs, L1-6)** — large cache-bound data, the corpus that matters most:

| level | geomean | 95% CI |
|------:|--------:|:-------|
| 1 | 1.046x | [1.037, 1.055] (real) |
| 2 | 1.037x | [1.031, 1.042] (real) |
| 3 | 1.017x | [1.010, 1.024] (real) |
| 4 | 1.014x | [1.006, 1.021] (real) |
| 5 | 1.006x | [0.998, 1.014] |
| 6 | 1.005x | [1.000, 1.011] |
| **ALL** | **1.021x** | **[1.017, 1.024] (real)** |

**Canterbury (5 pairs, L1-6)** — small files; the −1.5% regression is gone:

| | geomean | 95% CI |
|---|--------:|:-------|
| **ALL** | **1.008x** | **[1.001, 1.014]** — neutral, no size penalty |

Per-file Canterbury crossover, before → after the right-size fix: grammar.lsp (3.7 KB) 0.948x → **1.047x**, fields.c (11 KB) 0.962x → 1.006x, cp.html (24 KB) 0.974x → 1.004x; large files unchanged. The size-dependent penalty is eliminated.

### Graphs (full-run, cross-session — understates the delta vs the paired numbers above; shown for curve positioning + output-neutrality)

silesia compress pareto
canterbury compress pareto

Net: **Silesia +2.1%, Canterbury neutral, byte-identical output, O(min(window, filesize)) memory** (never larger than before, removes the O(filesize) defect). The history is now two commits — the ring (with right-sizing) and the in-PR dashboard refresh.

Issue #2631, opened 2026-06-16
perf(explore): BitWriter wide-store — local usetUInt64LE + pre-sized buffer (unblocked by #2707)


kim-em:
**Status: exploration — unblocked. The lean4#14053 dependency is superseded: PR #2707 landed local `@[extern]` word-sized ByteArray *readers* (`ugetUInt32LE`/`ugetUInt64LE`, `Zip/Native/Wide.lean` + `c/bytearray_wide_ffi.c`, always compiled), so the *writer* (`usetUInt64LE`) can be added the same way in-repo now, no upstream wait. This is the writer follow-up the readers PR explicitly deferred. Still gated on a measured win (see step 0).**

## Idea
`BitWriter.flushAcc` pushes output one byte at a time (`data.push acc.toUInt8`) — ~1–2 byte pushes per token, ~630k pushes for a 2 MB L6 stream. Replace the growing-push model with a pre-sized output buffer + a byte position, storing 8 bytes per `usetUInt64LE` (no realloc, fewer ops). Emit is ~4–9% of compress, so the realistic ceiling is part of that slice.

## Staged plan

**Step 0 — kernel microbenchmark first (the gate).** Unlike the hash-load (#2629, 1.26×) and refill (#2630, 1.63×), the writer wide-store is *untested even at kernel level*. Before any restructure, microbench an 8-byte `usetUInt64LE` store loop vs the current per-byte `data.push` over a representative emit trace. If it does not clear a clear margin at the kernel, stop here — the buffer restructure (step 2) is not worth it for a fraction of a 4–9% slice.

**Step 1 — `usetUInt64LE` extern + model + conformance (easy, mirrors #2707).** Add the store form to `Zip/Native/Wide.lean` + `c/bytearray_wide_ffi.c`: `usetUInt64LE (a : ByteArray) (off : USize) (v : UInt64) (h : off.toNat + 8 ≤ a.size) : ByteArray`, owned arg, in-place byte writes; Lean model is the 8 little-endian `a.set` writes the store assembles, `@[extern]` swaps in one wide store. Conformance test sweeps offsets/values against the model, exactly like the readers.

[...]

Issue #2630, opened 2026-06-16
perf(explore, needs lean4#14053): decode wide bit-refill — ugetUInt64LE in refill


kim-em:
**Status: exploration — BLOCKED on lean4#14053 (open). Do not claim until #14053 merges.**

## Idea
`InflateBuf.refill` fills the bit buffer with a 7-iteration byte loop. With #14053 it becomes one bulk `ugetUInt64LE` + shift + advance (byte-assembly fallback near the buffer end). `refill` is 13.9% of decode (clean decode-ab profile).

## Measured (pr-release-14053 kernel microbench)
byte-refill vs wide-refill = **1.63×** (earlier campaign run measured 2.08×; ≥1.5× robustly). refill at 13.9% → estimate **~5% decode**.

## Why we might reject
- Earlier the *matcher* USize-refill in-situ regressed (−19%) due to per-call tuple allocation on a short loop — the decode refill must be wired so it does not reintroduce that tuple-alloc boundary. Confirm in situ.

## Status
Kernel numbers staged (Refill.lean on pr-release-14053); lands when #14053 merges.

kim-em:
Tracked as Phase 1 of the decoder-architecture umbrella #2668 (the only phase blocked on lean4#14053; Phases 2-5 in #2669/#2670/#2671/#2672 are doable now).

kim-em:
## Unblocked + "prototype both, measure" results

The reader half is done: **`ByteArray.ugetUInt64LE` landed in #2707** (local `@[extern]`, USize-indexed, model = the byte recombination, conformance-tested), so this is no longer blocked on lean4#14053 — only the refill *integration* remains.

### Microbench: byte-loop vs two wide variants

Compiled, core-pinned, refill→consume cycle over a 2 MB buffer (40 iters, repeated):

| variant | speed | vs byte-loop |
|---|---|---|
| byte-loop (current) | ~13–14 MB/s | 1.0× |
| **masked variable-width** | ~18–19 MB/s | **~1.40×** |
| **giesen trick** | ~19–20 MB/s | **~1.49×** |

- Both wide variants beat the per-byte loop by ~1.4× (consistent with the kernel 1.63× noted here; the synthetic consume is a little heavier than real Huffman). In-situ this will dilute like the hash3 reader did (kernel 1.26× → +2.2% Silesia compress in #2707), so expect a single-digit **decode** gain.
[...]

Issue #2629, opened 2026-06-16
perf(explore, needs lean4#14053): compress hash-load — ugetUInt32LE in hash3


kim-em:
**Status: exploration — BLOCKED on lean4#14053 (open). Do not claim until #14053 merges.**

## Idea
The 4-byte hash (PR #2620) assembles its word as `a ||| (b<<<8) ||| (c<<<16) ||| (d<<<24)` from 4 `uget`s — that is exactly little-endian `data.ugetUInt32LE pos`. With #14053 the whole assembly collapses to one wide load (keep the byte-assembly as the near-EOF fallback). `hash3` runs per position (mainLoop 8.3% of L6) and per interior insertion (`updateHashesFast` 10.8%), so it is one of the most frequent ops in the compressor.

## Measured (pr-release-14053 kernel microbench, dickens)
byte-assembly 2235 vs wide-load **2810 Mpos/s = 1.26×** on the hash kernel. The assembly is part of each `hash3` call → estimate **~2–3% overall compress**, free (no algorithm change), all levels. Distinct from the earlier matcher word-*compare* null (strided short-match); this is a fixed single full-width load — the good wide-load case.

## Why we might reject
- In-situ dilution: 1.26× on the kernel is only part of each call site, so the realized gain is single-digit %. Confirm in situ once #14053 lands.

## Status
Patch + kernel numbers staged; lands when #14053 merges. We are on v4.30.0-rc2; #14053 is open against master.

kim-em:
Note for when #14053 lands: the sibling **matcher word-*compare*** (word-at-a-time `countMatch`) — dismissed above as a "null (strided short-match)" — is worth re-measuring rather than inheriting that verdict. The original null was on a much earlier codebase, and a lot has changed since:

- the hash-chain matcher landed (#2485) and is now the default for all levels, so `countMatch` runs many times per position (once per chain candidate), not once;
- the `scan_end`-style prefilter (#2622) already skips the full compare when the byte at `bestLen` mismatches, so the compares that *do* run are the longer ones — exactly the case where a wide compare amortizes, shifting the length distribution away from the "strided short-match" regime the null was measured in;
- the circular `prev` ring (#2695 / #2702) just confirmed empirically that `countMatch`'s per-byte `goU` loop is the dominant cost of each chain step: shrinking the chain array to a cache-resident ring moved compress only ~+2% on Silesia precisely because the chain access is amortized against that byte loop. The byte loop is where the remaining time is.

[...]

PR #2707, opened 2026-06-24
perf(deflate): local @[extern] word-sized ByteArray readers, wired into hash3

kim-em:
This PR adds project-local `@[extern]` word-sized little-endian ByteArray readers — `ByteArray.ugetUInt32LE` / `ugetUInt64LE` (`Zip/Native/Wide.lean` + `c/bytearray_wide_ffi.c`, always compiled like `copyWithin`) — mirroring the readers of [lean4#14053](https://github.com/leanprover/lean4/pull/14053) (`wide fixed-width load/store`): a USize-indexed byte-recombination whose Lean model is the inline byte assembly the hot loops use today, so swapping it in is definitional modulo the offset's Nat/USize round-trip and `@[extern]` only changes the compiled implementation to one wide load. It wires `ugetUInt32LE` into the three `lz77*.hash3` matcher variants (the four-byte hash word), keeping the byte-recombination as the near-EOF / non-USize-addressable fallback, and adds a conformance test that sweeps every in-bounds offset over several byte patterns, comparing the `@[extern]` result against a pure-Lean recombination (validating the C against the model).

Because the three `hash3` bodies stay textually identical the `hash3_eq` bridge is still `rfl`, and the body still ends in `_ % hashSize` so the `< hashSize` bound is unchanged; the whole Spec suite holds and matcher output stays byte-identical. This supersedes the "blocked on lean4#14053" framing of [#2629](https://github.com/kim-em/lean-zip/issues/2629) by proving the reader value in-repo now.

In-situ controlled interleaved measurement (paired AFTER/BEFORE, one core, Canterbury, median-of-5 per run) shows the hash3 wide load is a small but real compress speedup: about +1% geomean overall, +1.3–1.6% on L1–L6 where hashing is hottest, flat at L7–L9 (match-finding dominated). The 95% CI excludes 1.0, so it clears machine noise. Full before/after overlay graphs to follow as a comment.

Scope is the readers plus the hash3 site only. The `refill` wide load ([#2630](https://github.com/kim-em/lean-zip/issues/2630)) is deliberately deferred: the production decoder's per-byte refill is woven into the `goFusedPU`/`goTreeFree` decode-equivalence chain under the `BufCorr` invariant, and a general-`cnt` wide load loads a variable number of bytes `k` (a literal 8-byte load gives `cnt+64 > 64`, breaking `BufCorr`), so landing it with proofs intact needs a masked variable-width equivalence lemma threaded through that chain — a real proof refactor, exactly the dilution pitfall #2630 flags. The BitWriter wide-store ([#2631](https://github.com/kim-em/lean-zip/issues/2631)) stays deferred. If the readers prove their worth this is direct evidence for upstreaming lean4#14053, or for moving the accessors into lean-zip-common.

Closes the reader half of #2629.

🤖 Prepared with Claude Code

### DIFF
--- a/Zip.lean
+++ b/Zip.lean
@@ -36,6 +36,7 @@ import Zip.Spec.DynamicTreesCorrect
 import Zip.Spec.DynamicTreesComplete
 import Zip.Spec.InflateCorrect
 import Zip.Spec.InflateComplete
+import Zip.Native.Wide
 import Zip.Native.Adler32
 import Zip.Native.Crc32
 import Zip.Native.Inflate
--- a/Zip/Native/Deflate.lean
+++ b/Zip/Native/Deflate.lean
@@ -1,5 +1,6 @@
 import Zip.Native.BitWriter
 import Zip.Native.Inflate
+import Zip.Native.Wide
 import Std.Tactic.BVDecide
 
 /-!
@@ -458,11 +459,28 @@ where
     -- Hash arithmetic in `UInt32` (single machine ops) rather than `Nat`
     -- (whose bitwise XOR/shift are slow); `.toNat % hashSize` keeps the exact
     -- same index, so `hash3_eq` stays `rfl` and the `< hashSize` bound holds.
-    let a := (data[pos]'(by omega)).toUInt32
-    let b := (data[pos + 1]'(by omega)).toUInt32
-    let c := (data[pos + 2]'(by omega)).toUInt32
-    let d := if h3 : pos + 3 < data.size then (data[pos + 3]'h3).toUInt32 else 0
-    let word := a ||| (b <<< 8) ||| (c <<< 16) ||| (d <<< 24)
+    -- When the four hash bytes are in bounds and the buffer is `USize`-
+    -- addressable, read them as one wide load (`ugetUInt32LE`, #2706); its model
+    -- is exactly the `a ||| b<<<8 ||| c<<<16 ||| d<<<24` recombination, so the
+    -- hash value is identical and all three `hash3` variants stay defeq.
+    let word :=
+      if h4 : pos + 4 ≤ data.size then
+        if hsz : data.size.toUSize.toNat = data.size then
+          ByteArray.ugetUInt32LE data pos.toUSize (by
+            have hds : data.size < USize.size := by
+              rw [← hsz]; exact USize.toNat_lt_two_pow_numBits _
+            rw [toUSize_toNat_of_lt (show pos < USize.size by omega)]; omega)
+        else
+          let a := (data[pos]'(by omega)).toUInt32
+          let b := (data[pos + 1]'(by omega)).toUInt32
+          let c := (data[pos + 2]'(by omega)).toUInt32
+          let d := (data[pos + 3]'(by omega)).toUInt32
+          a ||| (b <<< 8) ||| (c <<< 16) ||| (d <<< 24)
+      else
+        let a := (data[pos]'(by omega)).toUInt32
+        let b := (data[pos + 1]'(by omega)).toUInt32
+        let c := (data[pos + 2]'(by omega)).toUInt32
+        a ||| (b <<< 8) ||| (c <<< 16)
     (((word * 2654435761) >>> 16).toNat % hashSize)
   countMatch (data : ByteArray) (p1 p2 maxLen : Nat)
       (h1 : p1 + maxLen ≤ data.size) (h2 : p2 + maxLen ≤ data.size) : Nat :=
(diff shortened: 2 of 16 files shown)


kim-em:
## In-situ perf: hash3 wide load (controlled paired measurement)

A +1–2%-class compress delta is below the routine cross-session overlay's noise
floor and invisible on a log-axis speed/ratio graph, so the verdict here is the
**controlled interleaved paired measurement** the `perf-pr-graphs` skill
prescribes for single-digit deltas: AFTER (this branch) and BEFORE (merge-base
`bd0b9188`) binaries run **alternately, pinned to one core**, median-of-5 per
run, so common-mode machine load cancels in the per-run ratio. Machine `chungus`
(shared with concurrent perf agents during this run — pairing is exactly what
makes the ratio robust to that).

**Silesia — 4 pairs (n=276 file·level·pair obs), the representative corpus:**

| level | compress Δ (geomean after/before) |
|-------|-----------------------------------|
| L1 | **+2.57%** |
| L4 | +2.07% |
| L6 | +1.97% |
| **overall** | **+2.20%  (95% CI [+1.98%, +2.42%])** |

**Canterbury — 3 pairs (n=198):**

| level | compress Δ |
|-------|-----------|
| L1 | +1.04% |
| L2 | +1.16% |
| L4 | +1.27% |
| L6 | +1.35% |
| L8 | −0.40% |
| L9 | +0.14% |
| **overall** | **+0.76%  (95% CI [+0.50%, +1.01%])** |

Both CIs exclude 1.0 → the win is real, not noise. It is strongest where the
hash3 probe is the hot inner work (Silesia L1 +2.6%, the lower Canterbury levels
+1–1.4%); on Canterbury L7–L9 the cost is chain-walk match finding, so the hash
speedup dilutes into noise. This matches #2629's in-situ-dilution concern: the
kernel 1.26× on the hash itself lands as ~+2% on Silesia / ~+1% on small-file
Canterbury once amortized over the rest of the matcher.

**Ratio is byte-identical** at every level on both corpora (the matcher token
stream is unchanged by construction), so there is no speed/ratio tradeoff —
purely a speed shift up the frontier.

Decode is unchanged: the `refill` wide load (#2630) is deferred, since landing it
with the `BufCorr` / `goFusedPU` decode-equivalence proofs intact is a separate
masked-variable-width-equivalence refactor (see the PR description).

_Committed dashboard (`bench/results/latest.json` + overlay SVGs) refresh to
follow on a quiet machine — the box is currently saturated by concurrent perf
agents, so absolute MB/s snapshots would be depressed; the paired ratios above
are the contention-robust evidence._

kim-em:
## Before/after overlay — `compress_mbps` (machine `chungus`, native before vs after)

Comprehensive cross-session overlay (BEFORE = merge-base `bd0b9188`, AFTER = this
branch; both all-9-levels, pinned to free cores to avoid the concurrent agents on
the box). It **agrees with the controlled paired measurement above**: a uniform
small upward speed shift at **byte-identical ratio**.

**`max |Δratio| = 0.000000` → output-neutral** (the matcher token stream is
unchanged by construction; the wide load only changes *how* the hash word is read).

### Silesia (geomean over 12 files)
silesia

|----:|------:|-------:|------:|:-------:|
| 1 | 0.347 | 54.3 | 55.7 | 1.02x |
| 2 | 0.342 | 49.7 | 51.3 | 1.03x |
| 4 | 0.334 | 37.7 | 38.9 | 1.03x |
| 6 | 0.332 | 33.8 | 34.9 | 1.03x |
| 7 | 0.330 | 26.4 | 27.1 | 1.03x |

### Canterbury (geomean over 11 files)
canterbury

|----:|------:|-------:|------:|:-------:|
| 1 | 0.314 | 33.7 | 34.2 | 1.02x |
| 4 | 0.305 | 26.2 | 26.7 | 1.02x |
| 6 | 0.304 | 24.5 | 25.0 | 1.02x |
| 7 | 0.300 | 19.3 | 19.7 | 1.02x |

The gap is small on a log axis (a +2–3% shift is a hair's width), but the AFTER
curve sits at-or-above BEFORE at every point and never moves right — pure speed,
no ratio cost. Silesia (the representative corpus) carries the win at +2–3% L1–L7;
it dilutes at L8–L9 where the optimal/lazy parse dominates. Dashboard refreshed
in-PR (commit `e2d3534f`, compress rows only — decode is untouched by this change).

kim-em:
## Post-merge verification: the +3.4% is real (and the cross-worktree figure understated it)

Follow-up note for the record. The headline +2.2% here was measured **cross-worktree** (AFTER in the main worktree vs a BEFORE binary built in a separate `/tmp` merge-base worktree). A later investigation found that methodology can silently compare two *different C toolchains* — it produced a −37% decode phantom elsewhere that vanished to +0.02% when rebuilt in one worktree.

So I re-verified this PR the right way: **both binaries built in the same worktree, toggling only the hash3 body** (`ugetUInt32LE` wide load vs the pre-#2707 inline byte reads), same compiler/flags/`.lake`:

```
hash3 WIDE vs ORIGINAL, same-worktree, 4 rounds, Silesia compress:
  L1 +3.30%   L4 +3.53%   L6 +3.30%   OVERALL +3.38%
  every per-file ratio positive (+1.2% .. +6.6%)   max|Δratio| = 0.00000
```

**The win is genuine and ~50% larger than the merged figure** (+3.4% vs +2.2%). Mechanism: hash3 reads a *fixed* 4 bytes per call, so the wide load is a guaranteed 4-`uget`-calls → 1-call reduction every position. (The win is the call-count, not single-instruction folding — folding the C to one `mov` measured neutral.) Nothing to change; #2707 stands, validated.

Issue #2622, opened 2026-06-16
perf(explore): L9 rescope — usable-speed optimal-parse tier (prefilter the DP walk + split L9-fast / L10-best)


kim-em:
**Status: exploration — may be rejected after measurement.** Not committed work; do not claim until triaged.

## Context
Native L9 is the optimal-parse cost-model DP. It has the best ratio in the field — dickens 0.364 (libdeflate 0.374, zig 0.376), nci 0.0855 (beats everyone) — but runs at ~1 MB/s, 30–70× slower than the fast pack's L9. On the speed×ratio Pareto it is a stranded bottom-left point. No competitor reaches that ratio at *any* speed (they bottom out near ratio 0.315), so making it usable would be a unique, defensible position rather than a nudge to the Pareto-dominated mid-range.

The DP's match collector `chainWalkAll` (`Zip/Native/DeflateParse.lean:73`) walks up to `optChainDepth = 256` candidates at every position, calling `lz77Greedy.countMatch` on every one (no prefilter), collecting up to `optCacheSlots = 8` increasing-length matches per position. It never received the discriminating-byte prefilter that landed on the L1–L8 walk (`chainWalkPacked`, PR #2618); the post-4-byte-hash dashboard shows L9 essentially unchanged (×1.03).

## Part A — speed the optimal parse at the same ratio (frontier push, byte-identical)
Port the prefilter to `chainWalkAll`: before `countMatch` on a candidate, check the byte at the current best length; if it differs, the candidate cannot extend the best match, so skip it. The collected match set is unchanged (byte-identical output), proven by the same equality-transfer pattern as `chainWalkPacked_eq`/`countMatch_le_of_byte_ne`. L9 walks 2× deeper than L6 (256 vs 128) with no prefilter today, so the doomed-candidate fraction — and the win — should exceed the 1.2–1.4× seen at L1–L8.

## Part B — split the tier (Pareto retune)
Sweep `optChainDepth` / `optCacheSlots` / rounds / `optNiceLen` on a speed×ratio Pareto. Hypothesis: most of the optimal-parse ratio gain comes from the first slice of depth/rounds with diminishing returns, so a retuned config keeps ~95% of the ratio at several× the speed. Promote that config to the default **L9-fast** (~8–15 MB/s target); preserve the current full-DP config as **L10 / `--best`** — the documented max-ratio ceiling, with the L9 ratio canary frozen onto it. Each level keeps a monotonic speed/ratio meaning.

## Payoff
[...]

Issue #2580, opened 2026-06-13
feat: proven-bounds in Zip/Native/DeflateParse.lean (16 ]! → ]'h; self-contained, no spec churn)


kim-em:
## Current state

`Zip/Native/DeflateParse.lean` (the near-optimal LZ-parse support: the
per-position match-candidate cache and bit-cost DP consumed by
`lz77Optimal`) still contains **16 `]!` runtime-bounds-checked array
accesses**. Each `]!` compiles to a runtime length check with a `panic`
branch on out-of-bounds; the proven-bounds campaign (recent landings:
PR #2548, #2551, #2553) is converting these hot-loop accesses to
proven-bounds `]'h` form across the native implementation.

The 16 accesses are at (as of this writing) lines 85, 88, 201, 220, 273,
291, 294, 295, 318, 334, 337, 340, 408, 409, 429, 430 — `prev[cand]!`,
`codeLens[...]!`, `extra[i]!`, `lens[b]!`, `lenCost[b]!`,
`distCost[dist]!`, `cost[i+b]!`, `cacheLens[...]!`, `cacheDists[...]!`,
`litCost[...]!`, `data[pos]!`, `chLen[pos]!`, `chDist[pos]!`.

**This file is self-contained for proof purposes.** Its module docstring
states explicitly: *"Everything in this file is heuristic: the candidate
cache and the cost tables never enter any correctness proof … the proofs
are completely independent of what this file computes, so it can be tuned
freely."* Validity of the emitted token stream is re-established
downstream by `lz77Greedy.countMatch` + explicit guards. Therefore this
conversion touches **only `DeflateParse.lean`** — no `Zip/Spec/` file
changes, no proof churn.

## Deliverables

1. Convert all 16 `]!` accesses in `Zip/Native/DeflateParse.lean` to
   proven-bounds `]'h` (or `]` where the index is structurally in range),
   introducing `dite` guards / loop-bound captures as needed. Follow the
   `proven-bounds` skill: guard capture, omega proofs, loop bound capture.
   The idiom in the codebase is e.g.
   `arr[idx]'(by simp [size_lemma] at h ⊢; omega)`.
2. Preserve behaviour exactly: the function results must be unchanged
   (this is heuristic code with no spec, so equivalence is by
   construction — do not alter the parse/cost logic, only the access
   form). Where a guard cannot be discharged, leave that access as `]!`
   and note why in a comment rather than weakening logic.
[...]

PR #2709, opened 2026-06-24
perf(explore): direct-index scanBounds — skip length codes that cannot qualify

### LINKED ISSUE #2641: perf(explore): direct-index scanBounds — the L9 DP's biggest symbol scans all 29 length codes per candidate
**Status: exploration — byte-identical, modest but clean; do not claim until triaged.**

## Context
After the #2636 prefilter landed, the L9 optimal parser splits roughly (this session, HEAD `8765a9f5`, `chungus`, median-of-5):

| phase | alice29 | share of L9 |
|-------|---------|-------------|
| candidate-cache chain walk (`buildCache`) | 38.9 ms | ~33% |
| two-round DP fill (`fillRegion`→`scanCands`→`scanBounds`) | ~43 ms | ~37% |
| (`computeChoices` total) | 81.6 ms | |
| full L9 (4 candidates) | 116.5 ms | |

The #2622 sampling profile flagged **`scanBounds` as the single biggest DP symbol (~17.8%)**.

## The waste
`scanBounds` (`Zip/Native/DeflateParse.lean:268`) evaluates a candidate at the length-code lower boundaries that fall inside its covered interval `(prevLen, len)`. It finds them by iterating **all 29 entries** of `Inflate.lengthBase` (`s = 0..28`) and testing `prevLen < b ∧ b < len` on each. `Inflate.lengthBase` is sorted ascending, so the qualifying entries are a contiguous sub-range — but the loop always runs the full 29 iterations, and it is called once per cache slot (≤ `optCacheSlots` = 8) per position per round (×2): on the order of `2·n·slots·29` boundary tests, most of which fail the interval predicate.

## Idea
Index directly into the covered range instead of scanning all 29:
- binary-search (or a precomputed `length → first-boundary-slot` table) for the first `s` with `lengthBase[s] > prevLen`,
- iterate while `lengthBase[s] < len`.

Same boundaries evaluated, same order, same running arg-min ⇒ **byte-identical** `chLen`/`chDist`, hence byte-identical output. Zero ratio risk — fits the L9 ratio-priority constraint exactly.

## Relationship to existing work (not a duplicate)
- **#2580** (open) converts the `]!` panic-checked reads in `DeflateParse.lean` to proven-bounds `]'h`. Orthogonal to this — that removes per-access bounds checks; this removes whole loop iterations. They compose (do both).
- **#2638** is a cheaper *approximate* parser that trades ratio. This is the opposite: it keeps the exact DP output and just stops scanning length codes that cannot qualify.

## Magnitude / risk
- Targets the profile's #1 DP symbol; expect a few % of L9, byte-identical. Modest but free.
- Small and self-contained. The candidate cache and emission proofs are untouched (the fill is heuristic, re-verified at emission), so an identity test on the choice arrays is a sufficient gate — no spec churn.

## Evaluation order
1. Spike the direct-index `scanBounds`; assert identical `chLen`/`chDist` vs current across Canterbury (identity gate).
2. A/B the fill time on alice29 / lcet10 / kennedy.xls.
3. Land if it clears machine noise; stack with #2580.

## Provability
Heuristic code (opaque to correctness), so the gate is an identity test (choices arrays unchanged); no theorem statements move.

_Data from an interactive profiling session (throwaway driver, not committed)._

kim-em:
This PR indexes directly into the covered length-code range of the L9 optimal parser's `scanBounds` instead of scanning all 29 entries of `Inflate.lengthBase` for every candidate. `lengthBase` is sorted ascending, so the boundaries that fall inside a candidate's interval `(prevLen, len)` form a contiguous slot range: a precomputed `lengthBoundaryStart[v]` gives the first slot whose base exceeds `v`, the scan starts there (dropping the now-redundant `prevLen < b` test, true for every slot by monotonicity), and stops as soon as `b ≥ len` (all later bases are larger). The DP's #1 sampled symbol from the [#2622](https://github.com/kim-em/lean-zip/issues/2622) profile now runs a few iterations per slot instead of 29.

Same boundaries, same order, same running arg-min as the full `0..28` scan, so `chLen`/`chDist` and the emitted bytes are unchanged. Verified byte-identical (size + CRC) across all 11 Canterbury files, and about 9% off the full `deflateRaw` level-9 path on the corpus (266 to 242 ms per round, stable across trials). This keeps the exact DP output and only stops scanning length codes that cannot qualify, so it carries zero ratio risk and fits the L9 ratio-priority constraint. It is orthogonal to and composes with [#2580](https://github.com/kim-em/lean-zip/issues/2580) (proven-bounds reads).

The code is heuristic (it never enters a correctness proof), so the gate is an identity test on the choice arrays rather than a theorem. A regression test pins `lengthBoundaryStart` against an independent first-exceeding-slot scan and asserts `lengthBase` is strictly ascending, so a future reorder cannot silently invalidate the monotonicity the optimization relies on.

Closes [#2641](https://github.com/kim-em/lean-zip/issues/2641).

🤖 Prepared with Claude Code

### DIFF
--- a/Zip/Native/DeflateParse.lean
+++ b/Zip/Native/DeflateParse.lean
@@ -356,27 +356,44 @@ private theorem seedTailCosts_size (cost : Array Nat) (r perByte j : Nat) :
 termination_by 259 - j
 decreasing_by omega
 
+/-- `lengthBoundaryStart[v]` is the smallest length-code slot `s` with
+    `lengthBase[s] > v` (equivalently, the count of length bases `≤ v`),
+    `29` if none. `Inflate.lengthBase` is sorted ascending, so this is the
+    first slot a covered interval `(v, len)` can include — `scanBounds`
+    starts its scan here instead of from slot 0, skipping the boundaries that
+    cannot qualify. Indexed by the previous candidate length, `0..258`. -/
+def lengthBoundaryStart : Array Nat :=
+  Array.ofFn (n := 259) fun v =>
+    (Inflate.lengthBase.filter fun b => decide (b.toNat ≤ v.val)).size
+
 /-- Evaluate one candidate at the length-code lower boundaries inside
     `(prevLen, len)` (its covered interval, exclusive of `len` which the
     caller already evaluated): truncating a match to a boundary can buy a
     cheaper length code or a better continuation. `(bestC, bestL, bestD)`
-    is the running arg-min. -/
-def scanBounds (lenCost distCost cost : Array Nat) (i dist prevLen len s : Nat)
+    is the running arg-min.
+
+    The caller starts `s` at `lengthBoundaryStart[prevLen]`, the first slot
+    with `lengthBase[s] > prevLen`, so the lower bound `prevLen < b` already
+    holds for every slot scanned (`lengthBase` ascending). Only the upper
+    bound `b < len` is tested, and once it fails the scan stops — all later
+    slots have an even larger base. Same boundaries, same order, same
+    arg-min as a full `0..28` scan with the `prevLen < b ∧ b < len`
+    predicate, just without the iterations that cannot qualify. -/
+def scanBounds (lenCost distCost cost : Array Nat) (i dist len s : Nat)
     (bestC bestL bestD : Nat) : Nat × Nat × Nat :=
   if h : s < Inflate.lengthBase.size then
     let b := Inflate.lengthBase[s].toNat
-    if prevLen < b ∧ b < len then
+    if b < len then
       -- These three reads stay panic-checked: the indices (`b` a length-base
       -- value, `dist` a cached distance, `i + b` a cost-array offset) are
       -- heuristic cache/table values whose bounds would require cache-content
       -- invariants the design deliberately keeps out of the proof boundary.
       let c := lenCost[b]! + distCost[dist]! + cost[i + b]!
       if c < bestC then
-        scanBounds lenCost distCost cost i dist prevLen len (s + 1) c b dist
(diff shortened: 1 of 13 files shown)


kim-em:
Before/after `compress_mbps` — native speed vs ratio on Canterbury + Silesia (machine `chungus`, Linux x86_64; geomean over files; other-language curves reused from the committed dashboard).

**Silesia**
silesia pareto

**Canterbury**
canterbury pareto

Output is byte-identical (`max |Δratio| = 0.000000` across all 23 files), so the curve does not move horizontally — only the L9 tail point shifts up (faster at the same ratio). The change touches only the level-9 optimal parser; levels 1-8 use the greedy matcher and are unaffected.

| level | Canterbury | Silesia |
|------:|-----------:|--------:|
| 1-8   | 0.99-1.00x (unchanged) | 0.99-1.00x (unchanged) |
| **9** | **+14.7%** | **+14.0%** (every file +11%..+18%) |

L9 geomeans: Canterbury +14.7% (n=11), Silesia +14.0% (n=12). An interleaved-paired measurement on a single pinned core (8 pairs) puts Canterbury L9 at 1.147x, 95% CI [1.137, 1.157], with levels 1-8 flat at 0.996-1.001x.

Methodology note: an initial cross-worktree comparison showed a spurious ~8% uniform slowdown on *all* levels — traced to a contaminated `.lake` that built the two binaries with different C compilers (clang vs gcc, 1 vs 188 LLVM symbols, 454 KB size gap). After `rm -rf .lake` and rebuilding both sides in one environment (binaries then differ by exactly 3,336 bytes = this change), the artifact is gone and levels 1-8 sit on the noise floor as expected.

Issue #2710, opened 2026-06-24
perf: ladder redesign — fast L1 (encode-bound), even Pareto spread, libdeflate 1-12 (L9 unchanged)


kim-em:
**Consolidated ladder redesign.** Folds in #2698 (L4-6/L7-8 collapse) and is informed by two measurement passes run this session on `chonk` (Silesia geomean). Goal: a compress Pareto where the *fast end* and the *bunched middle* are fixed — a genuine fast L1 corner and an evenly-spread L2-L8. **L9 is left unchanged** (see non-goal). No compatibility worry (nothing in production yet), so relabelling the lower levels is free.

## Problem

**No fast corner, and a bunched middle.** Native L1 over-compresses: Silesia ratio 0.347 (better than zlib 0.375, Go 0.388, miniz 0.431) but only ~51 MB/s. Everyone else's L1 trades ratio for speed (miniz 0.431 @ 238 MB/s); we have no fast-but-looser point. And L4/5/6 are bunched (L7 was just promoted by #2703).

## Non-goal: L9 stays as-is

L9 is the optimal-parse cost-model DP, ratio ~0.285 (Canterbury) / ~0.308 (Silesia) — within ~2% of zopfli's floor, remarkable for a verified pure-Lean codec — at ~1.2 MB/s, off in the zopfli corner. **We keep it as the numbered max level.** (An earlier draft proposed moving it behind a `--best` flag on the zstd `--ultra` / xz `-e` precedent; dropped, because those are non-DEFLATE programs — the DEFLATE norm, gzip/zlib/libdeflate, is numbered-levels-only with the top simply being the slowest.) So L9's slow-but-best point is expected and fine; this redesign is about the fast end and the middle. (#2638's "usable L9 tier" remains a separate, optional effort to make L9 *itself* faster.)

### (a) Level-0 / memcpy ceiling
`deflateStored` (no match, no Huffman) over Silesia: **ratio 1.0, ~8.4 GB/s** — the compress fast-end anchor. (Decode-side memcpy ceiling was ~40 GB/s.) The gap from L0 (~8400 MB/s) to L1 (~51 MB/s) is **~160×**, and it lives almost entirely in the encode path, not the matcher (see (b)). **Record this number in the README; do NOT plot it** — it blows out both axes.

### (b) Fast-L1 knob sweep — **L1 is emit-bound, not match-bound**
[...]

PR #2711, opened 2026-06-24
perf: respread the L4-L6 ladder, sweep libdeflate 1-12, record memcpy ceilings

kim-em:
This PR respreads the lower-middle of the compression ladder, extends the libdeflate comparison to its full level range, and records the memory-bandwidth ceilings — the parts of the #2710 ladder redesign that survive measurement. The issue's headline (a fixed-Huffman-only fast L1 corner) was implemented, measured, and rejected; that negative finding is recorded rather than shipped.

De-bunch the level 4/5/6 cluster. At the prior uniform lazy `goodMatch = 8` the gate fired so eagerly that the chain-depth ladder (48/64/128) barely moved the ratio, so L4/5/6 sat almost on top of each other on the Silesia Pareto (0.3336/0.3330/0.3320 at ~39/38/35 MB/s). Graduating the lazy-gate threshold up the tier (L4=8, L5=16, L6=64; L7+ stays ungated) and dropping L4's chain depth to 32 makes each level trade speed for ratio in an even step. Verified by controlled paired measurement on one machine (both binaries built in the same clean environment, run same-session, core-pinned): the byte-identical levels (L1-L3, L7-L9) stay at ~1.00x, and only L4/5/6 move — L4 neutral, L5 ~7% slower for a tighter ratio (0.3330 -> 0.3322), L6 ~14% slower for a tighter ratio (0.3320 -> 0.3304). Each level stays Pareto-efficient; the cluster becomes a smooth frontier (Canterbury de-bunches identically). `goodMatch`/`chainDepth` are pure heuristics — the matcher contracts hold for any value — so there is no proof change.

Sweep `libdeflate` over its full levels 1-12 in the bench report (every other codec caps at 9), so its densest points 10-12 appear on the comparison Pareto; `plot.py` already discovers each codec's levels, so no plotter change was needed. This also fixes a latent crash the sweep exposed: the decompress-timing path builds its canonical reference stream with the zlib FFI, whose `deflateInit2` rejects levels above 9, so libdeflate L10-12 raised a `deflateInit2: stream error` — the reference level is now clamped to `min(level, 9)` (the reference is a fixed decode workload whose density barely shifts past L9, and it was never the codec's own output). Record the memcpy-class ceilings in `bench/README.md` prose rather than plotting them (they blow out a log axis): the compress ceiling (`deflateStored`, ~8.4 GB/s) and the decode ceiling (~40 GB/s). The dashboard is refreshed in-PR (native de-bunched, libdeflate 1-12, zlib/miniz re-measured, external comparators reused).

On the fast-L1 corner: a fixed-Huffman-only level-1 fast emit (skip the dynamic-tree construction, pair with a chain-depth-1 matcher) was implemented end-to-end with full roundtrip proofs, then measured in the packed production path with a controlled paired benchmark. It is a Pareto **regression on Silesia** — ~0.86-0.92x the speed at a much looser ratio (0.347 -> 0.40-0.44) — because the fixed block is looser, so the extra output bytes it writes cost more emit time than the dynamic Huffman tree it skips: large files are emit-bound on *bytes written*, and the dynamic block is both tighter and faster there. It only wins on small files (Canterbury +30-40%). This is exactly the caveat the issue flagged ("confirm the emit-bound result in the packed path before sizing the encode work"). So the fast-L1 machinery was reverted; a genuine fast-L1 corner needs a faster per-byte emit (a faster `BitWriter`), not a different block type. The finding is recorded in `bench/README.md`.

Addresses #2710 (D1, D2, D3); the fast-L1 encode-path work is deferred per the measured finding above.

🤖 Prepared with Claude Code

### DIFF
--- a/.claude/skills/perf-pr-graphs/SKILL.md
+++ b/.claude/skills/perf-pr-graphs/SKILL.md
@@ -50,7 +50,9 @@ optional nicety.
 
 The graphs show **native before vs after** on the real corpora (Canterbury +
 Silesia), overlaid on the existing other-language curves (zlib, miniz_oxide,
-libdeflate, Go, JS, Zig, OCaml) for context. The before/after **overlay** PNGs
+libdeflate, Go, JS, Zig, OCaml) for context. Native sweeps levels 1–9;
+`libdeflate` is swept 1–12 (its full range, the others cap at 9), so its
+curve carries the extra dense points 10–12. The before/after **overlay** PNGs
 (`$W/perf_before_after_*`) are a report artifact — NOT committed; they live
 only in the PR comment. The **dashboard** (`bench/results/latest.json` and the
 routine `bench/graphs/*.svg`), by contrast, **is** refreshed and committed inside
--- a/Zip/Native/DeflateDynamic.lean
+++ b/Zip/Native/DeflateDynamic.lean
@@ -548,12 +548,16 @@ attribute [irreducible] symbolBitCount fixedBlockBytes dynBlockBytes dynBlockByt
 /-- Hash-chain search depth per compression level (levels ≥ 5). Higher levels
     search deeper for longer matches (better ratio on diverse input) at higher
     cost; the `chainWalk` early-stop keeps repetitive input fast at any depth.
-    The ratio gain saturates around 256–512 (measured), so level 9 caps there. -/
+    The ratio gain saturates around 256–512 (measured), so level 9 caps there.
+
+    Level 4 uses depth 32 (was 48): with the graduated lazy-gate (see `goodMatch`)
+    it is the fast/loose anchor of the lazy tier, so a shallower chain there evens
+    the L3→L4→L5 ratio spread (#2710 D3). -/
 def chainDepth (level : UInt8) : Nat :=
   if level ≤ 1 then 8
   else if level ≤ 2 then 16
   else if level ≤ 3 then 32
-  else if level ≤ 4 then 48
+  else if level ≤ 4 then 32
   else if level ≤ 5 then 64
   else if level ≤ 6 then 128
   else if level ≤ 7 then 256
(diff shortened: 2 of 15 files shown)


kim-em:
Before/after native `compress_mbps` (machine `chungus`, geomean over real corpora; both binaries built in the same clean nix-shell `.lake` and run same-session, core-pinned — so unchanged levels read ~1.00x and the only movement is the intended L4/5/6 de-bunch). The libdeflate curve now carries its dense 1-12 points.

**Silesia**

silesia compress pareto

|----|-------|------|------|------|
| 1 | 0.347 | 56.6 | 56.8 | 1.00x |
| 2 | 0.342 | 51.9 | 52.0 | 1.00x |
| 3 | 0.338 | 47.6 | 48.0 | 1.01x |
| 4 | 0.335 -> 0.335 | 39.4 | 39.2 | 1.00x |
| 5 | 0.333 -> **0.332** | 38.1 | 35.6 | 0.93x |
| 6 | 0.332 -> **0.330** | 35.5 | 30.4 | 0.86x |
| 7 | 0.330 | 27.4 | 27.4 | 1.00x |
| 8 | 0.323 | 9.7 | 9.7 | 1.01x |
| 9 | 0.308 | 1.7 | 1.7 | 1.00x |

**Canterbury**

canterbury compress pareto

L5/L6 are the de-bunch trade: ~7%/14% slower for a tighter ratio, so the L4/5/6 cluster (which used to sit on top of each other, 0.3336/0.3330/0.3320 within 0.5% ratio) spreads into a smooth frontier. Every other level is byte-identical (1.00x) — D3 only touches the lazy-tier heuristics. Ratio is deterministic; `compress_mbps` is a single-machine snapshot.

Note: the issue's headline fast-L1 corner is **not** in this PR — a fixed-Huffman-only L1 emit was measured as a Pareto regression on Silesia (paired: 0.86-0.92x speed at a much looser ratio) and reverted; the negative finding is documented in `bench/README.md`. See the PR description.

kim-em:
Superseding this. Per discussion on the graphs: the D3 L4/5/6 respread only relabels points along the existing frontier (it doesn't expand the hull) and costs L5/L6 7-14% speed, and the fast-L1 corner needs real encode-path work (a faster per-byte emit), not the fixed-only trick that regressed Silesia. Splitting out the one cleanly-shippable piece — D1 (libdeflate 1-12) — as #2713, and taking on the emit-path work separately. Closing this.

PR #2716, opened 2026-06-26
perf(matcher): uset the updateHashesFast hot writes (~2% L1 compress)

kim-em:
This PR replaces the two panic-checked `Array.set!` writes in the matcher's interior hash-chain insertion loop (`updateHashesFast`, ~18-24% of level-1 compress per `perf`) with proven-bounds `Array.set`, dropping the redundant bounds-check branches. The bucket index `hsh < hashTable.size` and the chain mask `(pos+j) &&& 0x7FFF < chainWinSize ≤ prev.size` are already proven, so the checks `set!` performs are wasted work. Output is byte-identical; the matcher is shared, so every level benefits.

`updateHashesGuarded` gains one runtime `min chainWinSize data.size ≤ prev.size` check (the same guard `chainWalkGuardedPacked` already carries — always true for the production chain ring, so it is defensive, not hot-path branching) to establish the `prev`-mask bound, dispatching to the new `updateHashesFastU`; it falls back to the proven-bounds `set!` walk otherwise. `updateHashesFastU_eq` proves the new walk equal to `updateHashesFast` (the in-bounds `set` collapses to `set!` at each step via `Array.set!_eq_setIfInBounds`), and `updateHashesGuarded_eq` routes through it, so the unified matcher correctness chain is unchanged. Zero `sorry`s; all conformance and roundtrip tests pass.

Measured on `chungus` (same-worktree before/after, core-pinned). Full-corpus geomean compress, output-neutral (`max |Δratio| = 0`):

| level | Silesia | Canterbury |
|---|---|---|
| 1 | +1.9% | +1.3% |
| 2 | +1.3% | +0.8% |
| 3 | +1.8% | +0.7% |
| 4-6 | ~neutral | ~neutral |
| 7-9 (barely touch the loop) | ~1.00x | ~1.00x |

The win concentrates at the matcher-heavy low levels, as expected; a controlled paired `perf stat -r 5` on individual files corroborates (alice29 -2.3%, kennedy.xls -3.9%, dickens -3.1% wall-clock on a level-1 compress loop). The dashboard is refreshed in-PR (native speed rows + the six speed-derived SVGs; ratios and ratio heatmaps are unchanged since the change is output-neutral).

This is the first of the matcher de-boxing levers found by profiling level-1 compress (which is matcher-bound, ~52%, not emit-bound). The sibling lever — USize-ifying `chainWalkPacked` — is filed separately as #2714 (it needs a `prev`/`hashTable` content-bound invariant, a heavier proof).

[...]

PR #2717, opened 2026-06-26
perf(matcher): USize-ify chainWalkPacked — measured ~10% REGRESSION, do not merge

### LINKED ISSUE #2714: perf(matcher): USize-ify chainWalkPacked (+3% L1 compress, all levels) — needs a prev/hashTable bound invariant
## Summary

Carry `chainWalkPacked`'s arithmetic-heavy accumulators (`cand`/`bestLen`/`bestPos`/`pos`/`maxLen`/`windowSize`) in `USize` so the per-chain-step window check, byte prefilter, and chain mask compile to raw `size_t` ops + `uget` instead of boxed-`Nat` arithmetic. Spiked and **measured +2.5-3.1% end-to-end on level-1 compress**, byte-identical output; the win should be at least as large at higher levels (deeper chains → larger matcher share).

## Why

Profiling level-1 compress on the packed production path (Silesia, `perf record`, three contrasting files) shows L1 is **matcher-bound (~52%), not emit-bound** — correcting the #2710 premise. The two matcher hotspots are `chainWalkPacked` (~23%) and `updateHashesFast` (~18-24%). `chainWalkPacked`'s body is wall-to-wall boxed-`Nat`: per chain step it does `lean_nat_dec_lt`/`lean_nat_sub`/`lean_nat_add`/`lean_nat_land` plus boxed-index `lean_byte_array_fget` and a `lean_dec` on every temporary (read the generated C of `lp_..._chainWalkPacked___redArg`). `countMatch`'s inner byte loop is already `USize` (`goU`/`goU_eq`); the chain-walk arithmetic around it is not.

## Measured spike (this is the proven-to-work part)

Back-to-back, same worktree (same compiler, GCC 5.3.0), `taskset`-pinned, `perf stat -r 4`:

| file | base | USize walk | speedup |
|---|---|---|---|
| dickens (text) | 3.605s | 3.504s | -2.8% |
| xml | 0.726s | 0.708s | -2.5% |
| sao (astro) | 3.07s | 2.98s | -3.1% |

All conformance/roundtrip tests pass with the spike wired in (output unchanged). The spike adds `chainWalkPackedU` (USize loop, `fuel` kept `Nat` as the termination measure) and dispatches to it from `chainWalkGuardedPacked` behind the existing `data.size.toUSize.toNat = data.size` addressability guard, falling back to `chainWalkPacked` otherwise.

## The proof work (the reason this is its own issue)

`chainWalkPacked` feeds the chain array's **contents** back as the next loop index (`cand := prev[cand &&& 0x7FFF]`). For the USize loop to provably equal the `Nat` loop, every `prev` entry read must fit in 64 bits — i.e. `prev[idx] < USize.size`. But the matcher's whole "proven-bounds" design treats `prev`/`hashTable` as **unverified heuristic arrays** (their contents never enter any correctness proof), so no such bound currently exists.

So `chainWalkPackedU_eq : chainWalkPackedU … = chainWalkPacked …` needs a new invariant:

> every entry of `prev` (and `hashTable`) is `≤ data.size` (hence `< USize.size` under the addressability guard).

This holds in fact — `prev`/`hashTable` start as `.replicate _ data.size` and are only ever `set!` to positions `pos+j ≤ data.size` or to a previously-stored entry — so it's a straightforward "all entries `≤ data.size`" induction maintained through the matcher's `mainLoop` (`lz77ChainIter`/`lz77ChainLazyIter`), threaded down to the `chainWalkGuardedPacked` call site to discharge the hypothesis. The walk-level proof itself mirrors the existing `chainWalkPacked_eq` fuel-induction (`Zip/Spec/LZ77ChainCorrect.lean`), adding `USize.toNat_add`/no-overflow steps like `countMatch`'s `goU_eq`.

## Plan

1. Add `chainWalkPackedU` (USize walk) + dispatch in `chainWalkGuardedPacked` (the spike — see the saved patch / git history of the #2710 session).
2. Prove + thread the `prev`/`hashTable` entries-`≤ data.size` invariant through the matcher mainLoops.
3. Prove `chainWalkPackedU_eq`, route `chainWalkGuardedPacked_eq`/`_mod`/`_div` through it (replace the spike's `sorry`).
4. Re-measure in situ (back-to-back, same worktree) and refresh the dashboard if landing as a perf PR.

Sibling lever: `updateHashesFast` USize-ification (~18-24% share) — indices come from the position counter / hash, not array contents, so it has **no** content-bound obligation and a much cheaper proof. Being spiked separately.

🤖 Prepared with Claude Code

kim-em:
This PR carries `chainWalkPacked`'s arithmetic-heavy accumulators (`cand`/`bestLen`/`bestPos`/`pos`/`maxLen`) in `USize` so the per-chain-step window check, byte prefilter, and chain mask compile to raw `size_t` ops plus `uget` instead of boxed-`Nat` arithmetic with a `lean_dec` on every temporary. It adds `chainWalkPackedU` (the `USize` loop, with `fuel` kept `Nat` as the termination measure) and dispatches to it from `chainWalkGuardedPacked` behind the existing `data.size.toUSize.toNat = data.size` addressability guard, falling back to `chainWalkPacked` otherwise. The `countMatch` inner byte loop was already `USize` (`goU`); this brings the chain-walk arithmetic around it along. The spike measured +2.5-3.1% end-to-end on level-1 compress (dickens/xml/sao, `perf stat -r 4`, taskset-pinned), byte-identical output, with the win at least as large at higher levels (deeper chains).

The `USize` walk feeds `prev`'s contents back as the next loop index (`cand := prev[cand &&& 0x7FFF]`), so its provable equality with the `Nat` walk needs every chain link to fit in 64 bits. This PR adds the `prev`/`hashTable` entries-`≤ data.size` invariant — the arrays start as `.replicate _ data.size` and are only ever `set!` to positions `≤ data.size` or to a previously-stored entry — proves its preservation through `guardedSet`/`headProbeGuarded`/`updateHashes`, and threads it through `lz77ChainIter`/`lz77ChainLazyIter`'s mainLoops to discharge the new bound side conditions on `chainWalkGuardedPacked_eq`/`_mod`/`_div`, which now route through `chainWalkPackedU_eq`. The window comparison is clamped to `min windowSize data.size` inside the `USize` loop, which is output-preserving for any in-window candidate (`cand < pos ≤ data.size`) and so needs no separate `windowSize < USize.size` hypothesis.

Output is byte-identical (proven by `chainWalkPackedU_eq` plus the existing packed-walk decode lemmas); all conformance/roundtrip tests pass. No new `sorry`.

Closes #2714

🤖 Prepared with Claude Code

### DIFF
--- a/Zip/Native/Deflate.lean
+++ b/Zip/Native/Deflate.lean
@@ -762,6 +762,16 @@ theorem winMask_lt (n : Nat) : (n &&& 0x7FFF) < chainWinSize := by
   have := @Nat.and_le_right n 0x7FFF
   unfold chainWinSize; omega
 
+/-- `USize` form of `winMask_lt`: the masked `USize` index, viewed as a `Nat`,
+    is `< chainWinSize`. Bridges `USize.toNat_and` to the `Nat` mask bound so the
+    `USize` chain walk (`chainWalkPackedU`) indexes the ring in range. -/
+theorem winMaskU_lt (n : USize) : (n &&& 0x7FFF).toNat < chainWinSize := by
+  rw [USize.toNat_and]
+  have e7 : (0x7FFF : USize).toNat = 0x7FFF := USize.toNat_ofNat_of_lt_32 (by decide)
+  rw [e7]
+  have := Nat.and_le_right (n := n.toNat) (m := 0x7FFF)
+  unfold chainWinSize; omega
+
 /-- Greedy LZ77 with bounded-depth hash chains: at each position, walk the
     `prev` chain from the bucket head up to `maxChain` candidates and keep the
     longest in-window match. This finds far longer matches than the single-probe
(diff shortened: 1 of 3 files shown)


kim-em:
## ⚠️ Measured a regression, not a speedup — do not merge as-is

Before/after `compress_mbps`, machine `Linux chungus x86_64`, **output-neutral** (`max |Δratio| = 0.000000`, as the proofs guarantee). BEFORE is the branch merge-base `831a8bb2` (which already includes #2716's `updateHashesFast` USize change); AFTER is this branch `388f8548`.

The in-situ measurement **contradicts the issue's spike premise**: this change is a **~10% regression** on the chain-matcher levels (1-7), not a +3% win. Canterbury is median-of-5 (low noise) and the regression reproduced across two independent AFTER runs (31.4 and 30.9 MB/s at L1, vs 34.9 before) on a quiet machine (load 0.77, no competing jobs). Silesia (single-pass, noisier) agrees in direction.

canterbury pareto

silesia pareto

**Canterbury (median-of-5)** — native compress_mbps geomean:

|--:|--:|--:|--:|--:|
| 1 | 0.330 | 34.9 | 31.4 | 0.90x |
| 2 | 0.327 | 33.0 | 29.8 | 0.90x |
| 3 | 0.325 | 31.2 | 28.3 | 0.91x |
| 4 | 0.321 | 26.9 | 24.8 | 0.92x |
| 5 | 0.321 | 26.4 | 24.3 | 0.92x |
| 6 | 0.320 | 25.1 | 23.3 | 0.93x |
| 7 | 0.317 | 19.7 | 19.3 | 0.98x |
| 8 | 0.315 | 7.6 | 7.7 | 1.00x |
| 9 | 0.303 | 1.8 | 1.8 | 1.00x |

**Silesia (single-pass, ±30% noise)** — native compress_mbps geomean:

|--:|--:|--:|--:|--:|
| 1 | 0.397 | 57.1 | 49.5 | 0.87x |
| 2 | 0.393 | 52.6 | 45.7 | 0.87x |
| 3 | 0.391 | 48.4 | 42.0 | 0.87x |
| 4 | 0.387 | 39.4 | 35.0 | 0.89x |
| 5 | 0.387 | 38.2 | 34.0 | 0.89x |
| 6 | 0.386 | 35.4 | 31.9 | 0.90x |
| 7 | 0.385 | 27.4 | 25.6 | 0.93x |
| 8 | 0.377 | 9.7 | 9.5 | 0.98x |
| 9 | 0.361 | 1.7 | 1.7 | 0.99x |

### Why

The regression correlates exactly with chain-matcher usage and vanishes on L8/L9 (where per-position cost is dominated by deep chains / the optimal parse, not the walk arithmetic). The likely cause: carrying the walk state in `USize` forces `USize`↔`Nat` conversions at the `Array Nat` chain-ring boundary (each `prev` read round-trips `Nat`→`USize`) and at the `countMatch` call (`candU.toNat`/`posU.toNat`/`maxLenU.toNat`, which `countMatch` then re-converts back to `USize` for `goU`). On small positions, Lean's boxed-`Nat` ops are already pointer-tagged scalars (no heap alloc), so the window-check/prefilter savings are small while the boundary conversions are a **net tax** — exactly the effect the `chainWinSize` docstring already records ("the UInt32 conversions measured as a net tax", which is why the ring is `Nat` not de-boxed). The issue's +3% was likely measured against a pre-#2716 baseline or was noise.

### Status

The proofs are sound and the output is provably byte-identical (`chainWalkPackedU_eq` + the `prev`/`hashTable` entries-`≤ data.size` invariant), with no `sorry` and CI green — but a correct change that regresses speed should not land as a perf PR. **Converting to draft and recommending against merge.** The `BoundedEntries` invariant infrastructure may be worth keeping for a future chain-state optimization, but the `USize` walk itself does not pay off. Dashboard intentionally **not** refreshed (would record a regression as the baseline).

🤖 Prepared with Claude Code

kim-em:
### Salvage investigation: not recoverable as a perf win

I tested the obvious fix for the dominant suspected tax — the `countMatch` round-trip (`chainWalkPackedU` passes `candU.toNat`/`posU.toNat`/`maxLenU.toNat`, which `countMatch` then re-converts back to `USize` for `goU`). I replaced that call with a **direct `goU data candU posU 0 maxLenU` call** (all `USize`, no round-trip, also skipping `countMatch`'s internal addressability re-check) and re-measured (Canterbury median-of-5):

| level | before (Nat) | after (countMatch) | goU-direct | goU-direct / before |
|--:|--:|--:|--:|--:|
| 1 | 34.9 | 31.4 | 32.1 | 0.92x |
| 3 | 31.2 | 28.3 | 29.1 | 0.93x |
| 6 | 25.1 | 23.3 | 23.9 | 0.95x |

`goU`-direct recovers ~2-3 points but is **still ~8% slower** than the `Nat` baseline. With that change the match path is strictly cheaper (USize all the way, no `countMatch` dispatch) and the window check / prefilter are machine ops, so the only remaining per-step difference is the chain-link read itself: `prev` is an `Array Nat`, so every `prev[cand &&& 0x7FFF]` link must convert `Nat`→`USize` to feed the next `USize` `cand`. That single per-step boundary conversion — over the hottest inner loop in the matcher — outweighs all the `size_t` savings. De-boxing `prev` to avoid it is exactly the "net tax" the `chainWinSize` docstring already records.

**Conclusion: the `USize` chain walk is a net tax regardless of how cleanly it's done; there is no perf win to salvage.** The correct, reusable piece is the `BoundedEntries` invariant (`prev`/`hashTable` entries `≤ data.size`, threaded through both mainLoops) — but it has no consumer without the `USize` walk, so it isn't worth landing on its own. Recommend **closing** this PR and the issue as a measured-and-rejected negative result.

🤖 Prepared with Claude Code

kim-em:
Closing: measured-and-rejected. The `USize` chain walk is a ~8-10% regression even in its cleanest form (`goU`-direct), because the per-step `Nat`→`USize` conversion of each `prev` chain link outweighs the `size_t` savings on the window check / prefilter — the net tax the `chainWinSize` docstring already documents. No perf win to salvage; the `BoundedEntries` invariant is correct but has no consumer without this walk. Branch `issue-2714` is kept in history if the invariant is ever wanted.

Issue #2638, opened 2026-06-17
perf(explore): a cheaper approximate-optimal parser for a usable L9 tier (algorithm, not retune)


kim-em:
**Status: exploration — large; the only path to a *usable* L9 tier.** Split out from #2622 (whose retune Part B was refuted by measurement).

## Why
#2622 measured that L9 is **DP-cost-model-bound**, not chain-walk-bound: the per-position DP transitions (`scanBounds` 17.8% + `scanCands` 6% + `fillRegion` + `buildCache` ≈ 28%) dominate, and don't scale with the knobs (depth 256→32 = +14%, rounds 2→1 = +20%, ~+35% combined). The prefilter + dense-freq wins (banked, #2636) give ~+10% byte-identical but keep L9 at ~1.6 MB/s. A usable "L9-fast" tier (~10 MB/s near-L9 ratio, a top-left point no competitor reaches) therefore needs a **fundamentally cheaper algorithm**, not a retune of the current backward-DP.

## Candidate approaches
- **Lazy² / deeper bounded lookahead** (2-3 step) instead of a full per-position cost-model DP — much cheaper per position, captures part of the optimal-parse ratio gain.
- **Cheaper cost model** that avoids the per-position `scanBounds`/`scanCands` transition scan (e.g. a coarser bit-cost estimate, or pricing only the cache frontier without the inner bounds scan).
- **Region/block-local optimal** over smaller windows with a lighter DP.

## Magnitude / risk
- Potentially the only lever that makes L9 a competitive *point* (vs Rust L9 at 14-42 MB/s) rather than a stranded slow tier — high payoff if it lands near-L9 ratio at ~10 MB/s.
- **Large**: a new parser + its emission-validity proof (the candidate cache stays heuristic, so validity is re-established at emission as today, which helps). Ratio is the trap — a cheaper parser that loses much ratio collapses the unique position (the whole point is best-ratio-at-usable-speed).

## Why we might reject
If no cheaper algorithm gets within ~1-2% of the DP's ratio, there's no usable tier worth the build, and L9 stays the documented slow max-ratio ceiling.

---

## Ratio constraint (maintainer directive)

[...]

PR #2719, opened 2026-06-26
perf(bench): add enwik8 large-text corpus + zopfli ceiling for L9 ratio evaluation

### LINKED ISSUE #2700: perf(bench): add enwik8 large-text corpus + zopfli ceiling for L9 ratio evaluation
Add a large pure-text corpus (enwik8, the first 100 MB of English Wikipedia, from the Large Text Compression Benchmark) so we can evaluate L9 against the zopfli ceiling on zopfli's home turf.

## Why

Our L9 now sits right at zopfli's operating point (slightly faster, slightly worse ratio). Zopfli ships no corpus of its own; its published evaluation used standard corpora including **enwik8**, and its design target is large text / web content. Big pure-text is exactly the regime where near-optimal parsing pays off most, and where our current ratio gap to zopfli is widest. Canterbury is tiny and Silesia, while large and modern, is heterogeneous; a single large prose file is the missing data point for judging L9-vs-zopfli.

## What to do

- Add `bench/corpora/enwik8/` wired through `bench/fetch_corpora.sh` (SHA-256-pinned, gitignored cache, fetched on demand like Silesia). The harness discovers corpora by directory, so it slots into the per-level Pareto/summary/heatmap charts with no `plot.py` change.
- Generate the **zopfli ceiling** for it (`bench-report --zopfli-ceiling`) so the L9-vs-zopfli overlay appears on the enwik8 ratio graph. Note: zopfli on 100 MB at default iterations is very slow (~100x zlib); if the one-time run is impractical, use a fixed slice (e.g. enwik8 first 10-20 MB, recorded in the corpus name) — deterministic and SHA-pinned either way.
- Run the reduced (single-pass timing) matrix like Silesia, since it is large.

## Acceptance

enwik8 (or the chosen slice) appears in `bench/graphs/enwik8_*` with the zopfli ceiling overlaid on the ratio graph, and `bench/results/latest.json` carries its rows. This becomes the reference for the "is L9 strictly better ratio than libdeflate, and how close to zopfli" question (ties to #2638's ratio floor).

Context: requested to evaluate L9 against zopfli on a large-text corpus; sibling of the bench-regeneration chore #2699.

kim-em:
This PR adds a large pure-prose corpus (`enwik8_20m`, the first 20 MB of enwik8 from the Large Text Compression Benchmark) so L9 can be judged against the zopfli ceiling on zopfli's home turf — the large-text regime where near-optimal parsing pays off most and where our ratio gap to zopfli is widest. enwik8 is fetched on demand into the gitignored corpus cache (`fetch_corpora.sh enwik8`, official zip, SHA-256-verified on both the 100 MB original and the pinned 20 MB slice); the harness discovers it by directory, so it slots into the per-level Pareto/summary/heatmap charts with no `plot.py` change, and the zopfli ceiling overlays its ratio graph.

The full 100 MB at default zopfli iterations is ~100x slower than zlib and impractical, so the deterministic first-20 MB slice is the pinned, SHA-pinned input (recorded in the corpus name). The matrix `big` heuristic is now size-based (>16 MiB) rather than a `silesia` name check, so any large corpus runs the reduced single-pass timing automatically. `bench-report --zopfli-ceiling <out> <corpus>` regenerates one corpus and merges it into the frozen ceiling, keeping every other corpus's deterministic rows byte-identical — so the new corpus joins without re-running the ~1 h Silesia zopfli pass — and fails closed if the requested corpus is not materialized.

On enwik8 native L9 reaches ratio 0.356, beating libdeflate L9 (0.366) and landing just above zopfli's 0.353 ceiling, with libdeflate's densest L10–12 (0.355) in between. This is the reference data point for the "is L9 strictly better ratio than libdeflate, and how close to zopfli" question (ties to https://github.com/kim-em/lean-zip/issues/2638's ratio floor).

Closes https://github.com/kim-em/lean-zip/issues/2700.

🤖 Prepared with Claude Code

### DIFF
--- a/ZipBenchReport.lean
+++ b/ZipBenchReport.lean
@@ -143,6 +143,13 @@ def levels : List Nat := [1, 2, 3, 4, 5, 6, 7, 8, 9]
 def libdeflateLevels : List Nat := [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]
 def reps : Nat := 5
 
+/-- A corpus whose total bytes exceed this runs a single timing pass instead of
+    median-of-`reps` (variance is low on big files, and `reps` passes over a
+    large corpus dominate the matrix wall-clock). Size-based rather than a name
+    list so any large corpus (Silesia ~200 MB, the enwik8 ~20 MB slice, …) is
+    detected automatically. -/
+def bigCorpusThreshold : Nat := 16 * 1024 * 1024
+
 /-- Run one compressor over a list of `(pattern, bytes)` workloads × levels,
     timing compress (and optionally decompress on a canonical raw-deflate
     stream). The size of each row is the workload's byte length. -/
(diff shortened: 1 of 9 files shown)


kim-em:

PR #2728, opened 2026-07-01
feat: ship the L9-fast approximate-optimal parser at level 9; exact crown moves to level 10 (#2638)

kim-em:
This PR adds a cheaper approximate-optimal LZ parser and deploys it as a new level-9 tier, moving the exact backward-DP crown to level 10. **Level-9 semantics change:** native level 9 was the exact max-ratio crown and is now the faster L9-fast tier — callers pinning `level = 9` for absolute best ratio should now pass `level = 10` (the zlib/FFI bindings are a separate 0–9 path and are unchanged).

L9-fast makes three cost cuts versus the exact DP — a single DP round (no round-2 refit), no length-code boundary scan, and a shallower candidate cache (chain depth 64) — reaching near the crown's ratio at ~2× its speed. A knob sweep (measured on Silesia, chungus) put the operating point at chain depth 64 with the boundary scan off, landing ~20% *outside* the L8→L9 mixing frontier: a genuine new Pareto point, not something reachable by blending the existing tiers. On alice29.txt the tier ordering is L10 crown 52111 ≤ L9-fast 53059 < L8 54127.

Correctness is nearly free because the fast parser reuses the exact DP's choice-array format and its re-verifying emitter `optimalEmit`: the validity and encodability contracts in `LZ77OptimalCorrect` are stated for arbitrary choice arrays, so the L9-fast lemmas are the exact ones with `computeChoicesFast` substituted, and `deflateDynamicBlocksOptimalFast`'s roundtrip quadruple in `DeflateBlockSplit` is a byte-for-byte twin of the exact one. The `deflateRaw` dispatch gains a level-9 (fast) / level-10 (exact) split, threaded through the three roundtrip theorems in `DeflateRoundtrip`. Every branch stays proven, with no new axioms and 0 sorries; the tests exercise both parsers through the native and zlib-FFI decoders and pin the tier ordering. Levels ≥ 10 alias the crown; above the `optimalMaxSize` memory gate both tiers fall through to the split path.

🤖 Prepared with Claude Code

### DIFF
--- a/.claude/skills/perf-pr-graphs/SKILL.md
+++ b/.claude/skills/perf-pr-graphs/SKILL.md
@@ -50,9 +50,12 @@ optional nicety.
 
 The graphs show **native before vs after** on the real corpora (Canterbury +
 Silesia), overlaid on the existing other-language curves (zlib, miniz_oxide,
-libdeflate, Go, JS, Zig, OCaml) for context. Native sweeps levels 1–9;
-`libdeflate` is swept 1–12 (its full range, the others cap at 9), so its
-curve carries the extra dense points 10–12. The before/after **overlay** PNGs
+libdeflate, Go, JS, Zig, OCaml) for context. Native sweeps levels **1–10** —
+since #2638 level 9 is the L9-fast tier and **level 10 is the exact-DP crown**;
+always include level 10 so the crown stays on the Pareto (drop it and the graph
+looks like an L9 ratio regression with no crown shown). `libdeflate` is swept
+1–12 (its full range; the zlib/miniz FFI cap at 9), so its curve carries the
+extra dense points 10–12. The before/after **overlay** PNGs
 (`$W/perf_before_after_*`) are a report artifact — NOT committed; they live
 only in the PR comment. The **dashboard** (`bench/results/latest.json` and the
 routine `bench/graphs/*.svg`), by contrast, **is** refreshed and committed inside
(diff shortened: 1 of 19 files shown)


kim-em:
## Before/after native `compress_mbps` (machine `Linux chungus x86_64`, geomean over corpus files)

Native swept **1–10** after this PR (level 9 = L9-fast, level 10 = exact crown). Levels 1–8 are byte-identical to before (Δratio = 0 there); the only intended output change is level 9's ratio (max |Δratio| = 0.0145 at silesia/x-ray L9). BEFORE is a clean merge-base build; AFTER is a clean fresh build — an earlier AFTER run was CPU-contended by a concurrent build and discarded.

silesia before/after
canterbury before/after

**Silesia** (12 files, single-pass):

|------:|------:|------------:|-----------:|--------:|
| 1–7   | —     | —           | —          | **1.00× (flat, byte-identical)** |
| 8     | 0.323 | 9.7         | 9.7        | 1.00× |
| **9 (L9-fast)** | 0.315 | 1.7 (was crown) | **3.4** | **1.96×** |
| **10 (crown)** | 0.308 | — (new) | 1.7 | exact crown = old L9 |

**Canterbury** (11 files, median-of-5): L1–8 flat (1.00×); L9-fast **1.85×** (1.9→3.4 MB/s); L10 crown new.

**Read:** L1–8 unchanged; L9 becomes ~2× faster at a slightly larger output (the L9-fast tradeoff), and the exact max-ratio crown is preserved, now at level 10. On the mixing-frontier Pareto the L9-fast point sits above the old L8→L9 curve — a genuine new tier (~20% outside the frontier, measured in the knob sweep). Native isn't moved past another engine here; it gains an interior Pareto point. Dashboard refreshed in-PR to native 1–10.

PR #2729, opened 2026-07-01
feat: iterative kept-best squeeze as a new level 11 (exact crown stays at level 10)

kim-em:
This PR adds a new maximum-ratio compression level 11 that runs zopfli's iterative kept-best squeeze on top of the level-10 exact-DP crown, and leaves level 10 byte-identical to before. It threads a squeeze-round count `sq` (default 0) through the optimal-parse chain (`fillRegionRounds` / `computeChoices` / `lz77Optimal` / `deflateDynamicBlocksOptimal`) and dispatches `level == 10` at `sq = 0` and `11 ≤ level` at `sq = optSqueezeRounds`. At `sq = 0`, `fillRegionRounds` computes round 2 as one `squeezeStep` from round 1 and seeds `squeezeLoop` with it, so zero squeeze rounds return round 2 unchanged — the former two-round crown, verified byte-identical on all 23 corpus files.

Because the DP choice arrays only ever feed the re-verifying emitter, the correctness contracts are generic over the parse: the `lz77Optimal*` / `deflateDynamicBlocksOptimal*` lemmas and the `inflate_deflateRaw` roundtrip capstone are generalized over `sq` (strictly stronger — they now cover both the crown and the squeeze), and the dispatch's new `11 ≤ level` arm is discharged by the same roundtrip lemmas at `sq = optSqueezeRounds`. No theorem statement was weakened and no proof was routed around; the whole library builds green with no `sorry` and the full test suite (including the cross-engine interop fuzz) passes.

Level 11 reaches geomean ratio 0.284 (Canterbury) / 0.307 (Silesia), closing roughly 18–20% of the level-10-to-zopfli ratio gap at about 0.5–0.6 MB/s — a genuine new best-ratio point, better than every other engine measured except zopfli itself. The keystone for keeping this elaboration tractable is a memory fix: computing round 2 via `squeezeStep` keeps the round-2 cost tables opaque, holding peak elaboration around 0.75 GB where proving the refit's array sizes directly over the computed tables blew it past 25 GB.

Before/after speed-vs-ratio graphs (with the zopfli ceiling overlaid) and the per-level geomean table are in the PR comment.

🤖 Prepared with Claude Code

PR #2732, opened 2026-07-02
doc: warn that cross-worktree binaries are not layout-comparable (perf-pr-graphs)

kim-em:
This PR adds a "Cross-worktree build caveat" to the perf-pr-graphs skill's stale-baseline fallback. When a stale dashboard forces a merge-base rebuild, the skill currently builds that BEFORE binary in a separate worktree — but such a binary is not code-layout-comparable to the AFTER binary built in the main worktree. The two embed different absolute paths and can link objects into a different layout, shifting instruction-cache alignment of hot loops and producing a uniform ~5-15% speed offset across all levels. That artifact survives interleaving and shows ~1.00x same-binary drift, so it masquerades as a real regression even at levels the PR does not touch. The caveat instructs building both commits in the same worktree to isolate the source change, with the rule that untouched levels must overlay to within ~1%. It was found the hard way while producing the graphs for https://github.com/kim-em/lean-zip/pull/2729.

🤖 Prepared with Claude Code

### DIFF
--- a/.claude/skills/perf-pr-graphs/SKILL.md
+++ b/.claude/skills/perf-pr-graphs/SKILL.md
@@ -228,6 +228,24 @@ artifacts move into `$W`.
      then `git worktree remove --force $W/before`) and pass
      `$W/perf_before.json` as BEFORE.
 
+   **Cross-worktree build caveat (this bites — read it).** A binary built in a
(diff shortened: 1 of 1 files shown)


kim-em:

Issue #2726, opened 2026-07-01
perf: fast-L1 design — attribute the emit cost, then choose (deflate_fast matcher / fold-freqs / fixed-only)


kim-em:
**Fast-L1 design: attribute first, then choose.** The goal is a genuine fast-L1 corner (today L1 = 57 MB/s @ ratio 0.347 on Silesia — over-compressing, no fast point). The #2710 sweep showed L1 is *emit-bound* (matcher 60-103 MB/s vs emit 44-51 MB/s; shallowing the chain barely moves total). An earlier draft of this issue jumped straight to "emit a fixed-Huffman block only." A second-opinion review (Codex, reading the repo + zlib/miniz sources) flagged that as premature and partly mis-motivated. This reframes it: **measure the attribution, then pick the design.** Fixed-only is *one* candidate, not the presumed answer.

## Correction (what the fast levels actually do)

Fixed-Huffman-only is **not** what zlib -1 or miniz -1 do:
- **zlib -1** is `deflate_fast`: greedy, small chain (`max_chain=4`, `nice_length=8`), deferred insertion — with the *normal* stored/static/dynamic block arbitration (`trees.c`). Fixed-only is the separate `Z_FIXED` strategy.
- **miniz/miniz_oxide** level ≤3 sets `TDEFL_GREEDY_PARSING_FLAG`; fixed-only is `CompressionStrategy::Fixed` (`TDEFL_FORCE_ALL_STATIC_BLOCKS`), a separate strategy flag.

So the well-precedented fast L1 is a **fast matcher that keeps block arbitration**, not fixed-only emission. Copy the matcher policy first; treat fixed-only as a strategy corner.

## Step 1 (gate): attribute the L1 emit cost

Before any dispatch change, a tokens-held-constant microbench on Silesia. With `ptokens := lzMatchP data 1` fixed, time separately:
1. `tokenFreqsP ptokens`
2. `dynamicCodeLengths + dynHeaderCodes + dynBlockBytesWith` (the dynamic sizing/tree build)
3. `fixedBlockBytes`
4. `deflateFixedBlockP data ptokens` (fixed-only emit, incl. the token walk + BitWriter)
5. full `deflateRawBaseP data ptokens`
6. record which branch current L1 actually emits, per file.

[...]

Issue #2698, opened 2026-06-23
perf(explore): spread the per-level ladder — L4-6 and L7-8 collapse on the compress Pareto


kim-em:
The per-level compression ladder collapses: adjacent levels within two bands land almost on top of each other on the speed-vs-ratio Pareto, so the nine levels behave like ~4 operating points with cliffs between them rather than a smooth frontier.

## Data (native, master bench snapshot 2026-06-23)

Silesia, geomean over members:

| L | compress MB/s | ratio |
|---|---------------|-------|
| 1 | 51.66 | 0.3467 |
| 2 | 48.01 | 0.3416 |
| 3 | 44.28 | 0.3382 |
| 4 | 37.25 | 0.3336 |
| 5 | 35.97 | 0.3330 |
| 6 | 33.53 | 0.3320 |
| 7 |  9.99 | 0.3230 |
| 8 |  9.64 | 0.3228 |
| 9 |  1.22 | 0.3081 |

L4/L5/L6 are bunched (speed within ~10%, ratio within 0.0016); L7/L8 are nearly coincident (9.6 vs 10.0 MB/s, ratio 0.3228 vs 0.3230). Canterbury shows the same shape (4/5/6 at 24.6-26.2 MB/s ~0.304; 7/8 at 7.65-7.97 ~0.296).

## Diagnosis

The only knob that distinguishes adjacent levels inside these bands is `chainDepth` (`Zip/Native/DeflateDynamic.lean:552`): L4/5/6 = 48/64/128, L7/8 = 256/512. Chain depth has saturating returns, so both ratio and speed barely move across each band, collapsing the levels. Meanwhile the transitions that DO move the needle are the other two knobs flipping all at once: `goodMatch` (`:579`) jumps 8 to 259 at the 6 to 7 boundary (the lazy lookahead gate goes from aggressive to fully off), and `chainDepth` doubles 512 to 1024 at 8 to 9. The result is dense clusters separated by cliffs.

## What to explore

Spread the levels along orthogonal speed/ratio axes so each level is a distinct Pareto point, with the explicit goal of EXPANDING the convex hull, not shrinking it (a retune that just relabels points, or that pulls any level inside the current hull, is a regression). Candidate knobs, none of which is currently used as a smooth per-level separator:

- `nice_match` / `niceLen` early-out: currently effectively 258 (no early stop). A per-level threshold below maxLen buys speed for a little ratio, creating cheaper points to spread the fast end.
[...]

PR #2744, opened 2026-07-02
perf: niceLen early-out ladder in the chain walk

### LINKED ISSUE #2735: perf: niceLen early-out ladder in the chain walk (chainWalkPacked is 55% of L6 compress cycles)
## Evidence (perf, steady-state `compress-pareto`, 2026-07-02, chungus)

`chainWalkPacked` is the single hottest symbol at every chain level:

| level (file) | `chainWalkPacked` share |
|---|---|
| L1 (dickens) | 23% |
| L6 (dickens) | **55%** |
| L8 (xml) | 38% |

The walk currently has only two stops: fuel (chain depth) exhaustion, or `bestLen >= maxLen` (i.e. a full 258-byte match), `Zip/Native/Deflate.lean:945`. There is no early-out for "this match is already good enough", so at L4-L8 the walk routinely burns its whole depth budget improving a match that is already long.

## What libdeflate does

Every level has a `nice_match_length` cutoff: the chain walk stops as soon as the best match reaches it (`hc_matchfinder.h:279`). The ladder (`deflate_compress.c:3927-4013`): L2=10, L3=14, L4=30, L5=30, L6=65, L7=130, L8=258, L9=258. libdeflate holds better ratios than us at these levels, so the cutoff costs little ratio in practice.

## Plan

1. Add a `niceLen : Nat` parameter to `chainWalkPacked` (and its guarded wrapper): change the early-return test from `bl >= maxLen` to `bl >= min niceLen maxLen`.
2. Add a per-level `niceLen` ladder next to `chainDepth` (`Zip/Native/DeflateDynamic.lean:566`), starting from libdeflate's values, then sweep.
3. Thread the parameter through `lz77ChainIterP` / `lz77ChainLazyIterP` and the boxed twins.

Chain state is heuristic-only (never enters correctness; the emitters re-verify), so the `lzMatch_{encodable,empty,resolves}` contracts are untouched. Proof churn is limited to re-threading the parameter through the packed/guarded equivalence twins.

This delivers the first candidate knob of #2698 (spread the per-level ladder); the full-ladder resweep and hull validation stay there.

## Validation

`perf-pr-graphs` before/after compress pareto on both corpora. Expect L4-L8 to move up at small ratio cost; no level may move inside the prior hull.

🤖 Prepared with Claude Code

kim-em:
This PR adds a `niceLen` cutoff to the hash-chain match finder: the walk now stops as soon as the best match reaches `min niceLen maxLen` (libdeflate's `nice_match_length`) instead of only on fuel exhaustion or a full 258-byte match. A per-level ladder next to `chainDepth` sets the cutoff from libdeflate's values (L2=10, L3=14, L4=30, L5=30, L6=65, L7=130, L8=L9=258); level 1 keeps 258 to leave the measured `deflate_fast` corner (#2726) untouched, and `niceLen = 258` reproduces the prior behaviour exactly since `maxLen ≤ 258` at every call site.

The parameter threads through the whole chain-walk family (`chainWalk`/`chainWalkFast`/`chainWalkPacked` and their guarded wrappers) and both matcher families (greedy/lazy, boxed/packed). Chain state stays heuristic-only — every emitted match is re-verified at emission by `countMatch` plus the window guards — so the `lzMatch_{encodable,empty,resolves}` contracts are untouched; the proof churn is re-threading the parameter through the packed/guarded equivalence twins. `chainWalkPacked`'s prefilter skip path gains the same early stop so it stays byte-for-byte equal to the reference walk unconditionally.

Measured (same-environment interleaved before/after — see the PR comment for graphs and the build-environment caveat), this lands a small fast-level win at negligible ratio cost: +3.1% at L2 and +2.2% at L3 (where the low niceLen values fire), tapering to neutral by L6–L7, +1.1% overall on Canterbury, with `max |Δratio| = 0.0158`. The deep-level speedup the issue hypothesised does not appear with libdeflate's values — on our matcher, matches rarely reach 65/130 bytes at L6/L7, so those cutoffs are near-inactive (their Δratio ≈ 0). What lands here is therefore the correct, verified `niceLen` knob and ladder — the first candidate knob of #2698 — plus the fast-level win; the deep-level value sweep and hull validation stay in #2698.

Closes #2735

🤖 Prepared with Claude Code

### DIFF
--- a/Zip/Native/Deflate.lean
+++ b/Zip/Native/Deflate.lean
@@ -788,13 +788,13 @@ theorem winMask_lt (n : Nat) : (n &&& 0x7FFF) < chainWinSize := by
     emitted match. Reuses `lz77Greedy`'s `hash3`/`countMatch`/`trailing` helpers
     (and their proofs). -/
 def lz77Chain (data : ByteArray) (maxChain : Nat) (windowSize : Nat := 32768)
-    (insertCap : Nat := 1000000000) :
+    (insertCap : Nat := 1000000000) (niceLen : Nat := 258) :
     Array LZ77Token :=
   if data.size < 3 then
     (lz77Greedy.trailing data 0).toArray
   else
     let hashSize := 65536
-    (mainLoop data windowSize hashSize maxChain
+    (mainLoop data windowSize hashSize maxChain niceLen
       (.replicate hashSize data.size) (.replicate (min chainWinSize data.size) data.size) 0 insertCap).toArray
 where
   /-- Walk the `prev` chain up to `fuel` steps from `cand`, keeping the longest
@@ -802,18 +802,20 @@ where
       out-of-range link (`cand ≥ pos`, incl. the `data.size` sentinel). `prev`
       is the fixed-size circular ring indexed by `cand &&& 0x7FFF`; its contents
       are a pure heuristic, so the windowing never enters the validity proof. -/
-  chainWalk (data : ByteArray) (prev : Array Nat) (windowSize pos maxLen : Nat)
+  chainWalk (data : ByteArray) (prev : Array Nat) (windowSize pos maxLen niceLen : Nat)
       (hpm : pos + maxLen ≤ data.size) (cand fuel bestLen bestPos : Nat) : Nat × Nat :=
     if fuel = 0 then (bestLen, bestPos)
     else if hc : cand < pos ∧ pos - cand ≤ windowSize then
       have hcand : cand + maxLen ≤ data.size := by omega
       let ml := lz77Greedy.countMatch data cand pos maxLen hcand hpm
       let (bl, bp) := if ml > bestLen then (ml, cand) else (bestLen, bestPos)
-      -- Early stop: a match of the maximum possible length cannot be beaten, so
-      -- stop walking the chain. Provably zero ratio loss; on repetitive input
-      -- (matches reach `maxLen` immediately) this restores near-greedy speed.
-      if bl ≥ maxLen then (bl, bp)
-      else chainWalk data prev windowSize pos maxLen hpm prev[cand &&& 0x7FFF]! (fuel - 1) bl bp
+      -- Early stop: once the best length reaches the level's `niceLen` cutoff (or
+      -- the maximum possible `maxLen`, whichever is smaller) the match is already
+      -- good enough, so stop walking the chain (libdeflate `nice_match_length`).
+      -- The `maxLen` stop alone is provably zero ratio loss; a `niceLen < maxLen`
+      -- cutoff trades a little ratio for a shorter walk at the deep levels.
+      if bl ≥ min niceLen maxLen then (bl, bp)
+      else chainWalk data prev windowSize pos maxLen niceLen hpm prev[cand &&& 0x7FFF]! (fuel - 1) bl bp
     else (bestLen, bestPos)
   termination_by fuel
   decreasing_by omega
@@ -836,7 +838,7 @@ where
     else (hashTable, prev)
   termination_by matchLen - j
   decreasing_by all_goals omega
-  mainLoop (data : ByteArray) (windowSize hashSize maxChain : Nat)
+  mainLoop (data : ByteArray) (windowSize hashSize maxChain niceLen : Nat)
       (hashTable : Array Nat) (prev : Array Nat) (pos insertCap : Nat) : List LZ77Token :=
     if hlt : pos + 2 < data.size then
       let h := lz77Greedy.hash3 data pos hashSize hlt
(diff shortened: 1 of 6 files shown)


kim-em:
## Before/after `compress_mbps` (machine `chungus`, same-environment overlay)

canterbury
silesia

**Headline: no regression; a small fast-level win; the deep-level (L6) win the issue hoped for does not materialise with libdeflate's niceLen values.** The before/after native curves essentially coincide.

### Reliable delta (interleaved paired measurement, Canterbury, 6 pairs, single pinned core — common-mode noise cancels)

| level | speedup | note |
|---|---|---|
| L1 | +0.7% | control (niceLen=258, output unchanged) — this is the noise floor |
| L2 | **+3.1%** | niceLen=10 cutoff fires |
| L3 | **+2.2%** | niceLen=14 cutoff fires |
| L4 | +1.1% | |
| L5 | +1.0% | |
| L6 | +0.1% | niceLen=65 — near-inactive |
| L7 | −0.4% | niceLen=130 — near-inactive |
| **ALL** | **+1.1%** | |

Ratio cost is negligible (`max |Δratio| = 0.0158` at silesia/xml L2; ≤0.003 elsewhere). The cutoff only meaningfully changes output at L2–L3 — on our matcher, matches rarely reach 65/130 bytes at L6/L7, so libdeflate's deep-level values are near-inactive there (their Δratio ≈ 0). The deep-level win requires the value sweep, which the issue defers to #2698. This PR lands the correct, verified knob + ladder and a modest fast-level win.

### Important measurement caveat (and a benchmarking-infra finding)
The native curves above are ~9% below where a clean build would place them: **binaries built in a long-lived worktree run ~9% slower than binaries built in a fresh worktree, for byte-identical code** (confirmed by a master-vs-master control: −9.2%). Both before and after here were built in the **same** (main) worktree, so the before/after *delta* is honest; only the absolute position vs the reference curves is depressed. This also means the perf-pr-graphs skill's stale-baseline fallback (BEFORE in a fresh merge-base worktree, AFTER in the main worktree) produces a spurious ~9% "regression" — both sides must be built in the same environment. I have not refreshed `bench/results/latest.json` in this PR because of this unresolved environment question and the near-neutral result; that should happen at merge with clean-env numbers.

Correctness: proofs go through unchanged (zero `sorry`), roundtrip conformance (native↔FFI) passes, CI green.

kim-em:
## Correction: the earlier single-pass overlay was misleading — here is a rigorous self-controlled measurement

Kim correctly pointed out that the before/after overlay I posted above shows the native curve dipping below baseline at several levels, contradicting my "small win" claim. She was right to distrust it. That overlay is **single-pass**, and its noise floor (~±2%, provable from the fact that L9/L10 — code my change does not touch — read 0.98-0.99x in it) is larger than this change's signal. It should not have been presented as evidence. My follow-up interleaved runs then swung to −9% and even −15% on *unchanged-code control levels* — impossible for a real code effect, and the tell that those runs measured cross-core contention, frequency scaling, and run-order bias, not my code.

So I threw all of that out and ran a **sandwich** measurement: on one exclusive core with nothing else running, each rep is master→mine→master back-to-back. `mine / geomean(master1, master2)` cancels slow thermal/frequency drift; `master2/master1` (identical binary) is the self-control **noise floor**; L1 and L8 (niceLen=258, code unchanged) are per-level controls; the sandwich-position offset is measured from the controls and subtracted.

per-level effect

**Result — drift-corrected, control-subtracted, both corpora:**

| level | Canterbury (44/lvl) | Silesia (60/lvl) | verdict |
|---|---|---|---|
| L1 *(control)* | +0.1% | +0.0% | at floor ✓ |
| **L2** | **+1.6%** [+1.0,+2.7] | **+1.1%** | real win (cutoff fires, niceLen=10) |
| **L3** | +0.4% | **+1.4%** | small win (niceLen=14) |
| L4 | −0.1% | — | neutral |
| L5 | +0.1% | — | neutral |
| **L6** *(issue target)* | −0.2% | **+0.0%** | **neutral — no regression** |
| L7 | −0.2% | — | neutral |
| L8 *(control)* | −0.1% | — | at floor ✓ |

Noise floor is **±0.3%** (self-control), and both controls sit inside it — so the measurement is trustworthy. **No level regresses on either corpus, including L6.** The change is a small real win at L2-L3 (where the low niceLen values fire) and neutral everywhere else. The deep-level (L6) speedup the issue hypothesised does not appear with libdeflate's values — but crucially it does **not** regress either. The knob + ladder land correctly; the deep-level value sweep stays in #2698.

The single-pass overlay above should be read as superseded by this chart. Apologies for the earlier back-and-forth — the honest answer required controlling for the machine, not just measuring on it.

PR #2746, opened 2026-07-02
perf: word-at-a-time match extension in countMatch

### LINKED ISSUE #2736: perf: word-at-a-time match extension in countMatch (re-add ugetUInt64LE, XOR + trailing-zero compare)
## Evidence (perf, steady-state `compress-pareto`, 2026-07-02, chungus)

`lz77Greedy.countMatch` extends matches one byte per iteration (`goU`, `Zip/Native/Deflate.lean:505-521`: one `uget` + compare + branch per byte). At L9 (xml) it is 18% of cycles as a direct symbol, and it is also the inner loop of `chainWalkAllFast` (51%) and `chainWalkPacked` (55% at L6 dickens), so its cost is spread across every chain level.

## What libdeflate does

Match extension compares 8 bytes at a time: load two unaligned 64-bit words, XOR, and count trailing zeros to get the mismatch offset (`matchfinder_common.h:186-222`).

## Plan

1. Re-add `ugetUInt64LE` to `Zip/Native/Wide.lean` + `c/bytearray_wide_ffi.c`. It was removed as measured-neutral for the decode refill (the #2630 note in the Wide.lean docstring), but the matcher is a different consumer: extension reads runs of many bytes, not 1-byte top-ups. Trusted reference body = the 8 byte reads, same scheme as `ugetUInt32LE` (#2707, which measured +3.4% from the 4-byte hash load alone).
2. Rewrite the `countMatch` fast path: compare 8-byte words while `i + 8 <= maxLen`; on mismatch, find the first differing byte within the word (bounded 8-step scan or De Bruijn table, pure Lean, provable); per-byte tail near `maxLen`.
3. Prove equal to the existing per-byte `go` in the `goU_eq` style; all callers unchanged.

No dependencies. Expected win at every level >= 2, largest at L8-L10 where matches are long and chains deep.

## Validation

`perf-pr-graphs` before/after; expect a uniform upward shift of the chain levels, ratio identical (output-preserving change).

🤖 Prepared with Claude Code

kim-em:
This PR extends matches eight bytes at a time in `lz77Greedy.countMatch` instead of one, following libdeflate's `matchfinder_common.h`. It re-adds `ugetUInt64LE` (the eight-byte little-endian analog of `ugetUInt32LE`) to `Zip/Native/Wide.lean` and `c/bytearray_wide_ffi.c`, then adds a word loop `goUW` that, while a full eight-byte window fits, loads two little-endian `UInt64` words and compares them in one operation, advancing eight on a full-word match. On a mismatch it locates the first differing byte directly from the trailing-zero count of the XOR, `i + ctz(w1 ^^^ w2) >>> 3`, with **no re-scan** of the array. `countMatch` dispatches to `goUW`, so every chain matcher (`chainWalkFast`, `chainWalkPacked`) routes through it with no call-site change.

The count-trailing-zeros step matters for performance. An earlier revision handed a word mismatch back to the per-byte `goU`, which re-scans from `i`; that paid a wide load plus a full re-scan on every short match (the common case) and measured as a regression on both corpora. Locating the mismatch with `ctz` instead removes the re-scan and turns the change into a large-file win. A new FFI `lean_zip_ctz64` (`__builtin_ctzll`) provides the count, with `BitVec.ctz` as its trusted reference body.

The window guard is written `8 ≤ maxLen - i` in `USize` rather than `i + 8 ≤ maxLen` so it cannot overflow near the top of the address space; the carried `i ≤ maxLen` invariant makes the subtraction non-wrapping, and the proof converts it back before the wide load.

`goUW` is proven equal to `goU` (hence to the `Nat` `go`) by `goUW_eq`: a full-word match maps to eight `go` steps (`go_advance`), and the mismatch branch is `ugetUInt64LE_ctz_first_diff`, which shows `k = ctz(w1 ^^^ w2) >>> 3` is exactly the first mismatching byte (`k < 8`, every earlier byte agrees, byte `k` differs) by `bv_decide` over the `ugetUInt64LE` recombination and `BitVec.ctz`; `go` then advances to `i + k` and stops. The change is output-preserving (`max |Δratio| = 0.000000`), so `countMatch_matches` and `countMatch_eq` keep their statements. A `ByteArray.ugetUInt64LE` conformance test (extern vs pure-Lean reference over swept offsets and byte patterns) is added alongside the existing 32-bit one.

Measured on `chungus`, clean paired runs pinned to an idle core (per-file geomean of after/before): **Silesia +4.9%** (consistent across levels 1-8), **Canterbury -3.9%** on the small cache-resident files. Per the split-corpus guidance the Silesia win on large real-world data is the one that matters.

Closes #2736

🤖 Prepared with Claude Code

### DIFF
--- a/Zip/Native/Deflate.lean
+++ b/Zip/Native/Deflate.lean
@@ -553,12 +553,18 @@ where
         rw [USize.toNat_add]; apply Nat.mod_eq_of_lt; omega
       have hw1 : (p1 + i).toNat + 8 ≤ data.size := by omega
       have hw2 : (p2 + i).toNat + 8 ≤ data.size := by omega
-      if data.ugetUInt64LE (p1 + i) hw1 == data.ugetUInt64LE (p2 + i) hw2 then
+      let w1 := data.ugetUInt64LE (p1 + i) hw1
+      let w2 := data.ugetUInt64LE (p2 + i) hw2
+      if w1 == w2 then
         have hile' : (i + 8).toNat ≤ maxLen.toNat := by
           rw [USize.toNat_add, h8v, Nat.mod_eq_of_lt (by omega)]; omega
         goUW data p1 p2 (i + 8) maxLen hsz h1 h2 hile'
       else
-        goU data p1 p2 i maxLen hsz h1 h2
+        -- Word mismatch: the first differing byte is at
+        -- `ctz (w1 XOR w2) >>> 3` within `[i, i+8)` (libdeflate
+        -- `matchfinder_common.h`). Return `i + that offset` directly — no
+        -- re-scan of the array. Proven equal to the byte loop in `goUW_eq`.
+        i + (UInt64.ctz (w1 ^^^ w2) >>> 3).toUSize
     else
       goU data p1 p2 i maxLen hsz h1 h2
   termination_by maxLen.toNat - i.toNat
--- a/Zip/Native/Deflate.lean
+++ b/Zip/Native/Deflate.lean
@@ -485,14 +485,16 @@ where
   countMatch (data : ByteArray) (p1 p2 maxLen : Nat)
       (h1 : p1 + maxLen ≤ data.size) (h2 : p2 + maxLen ≤ data.size) : Nat :=
     -- P1a: when the buffer is `USize`-addressable (always true at runtime; the
-    -- check is one `USize` round-trip, no bignum), run the unboxed `USize` loop
-    -- `goU` — proven equal to the `Nat` `go` by `goU_eq`. Else fall back to `go`.
+    -- check is one `USize` round-trip, no bignum), run the unboxed word-at-a-time
+    -- loop `goUW` — proven equal to the byte loop `goU` (`goUW_eq_goU`), hence to
+    -- the `Nat` `go` (`goU_eq`). Else fall back to `go`.
     if hg : data.size.toUSize.toNat = data.size then
       have hsz : data.size < USize.size := by
         rw [← hg]; exact USize.toNat_lt_two_pow_numBits _
-      (goU data p1.toUSize p2.toUSize 0 maxLen.toUSize hsz
+      (goUW data p1.toUSize p2.toUSize 0 maxLen.toUSize hsz
         (by rw [toUSize_toNat_of_lt (by omega), toUSize_toNat_of_lt (by omega)]; omega)
-        (by rw [toUSize_toNat_of_lt (by omega), toUSize_toNat_of_lt (by omega)]; omega)).toNat
+        (by rw [toUSize_toNat_of_lt (by omega), toUSize_toNat_of_lt (by omega)]; omega)
+        (by simp)).toNat
     else go data p1 p2 0 maxLen h1 h2
   go (data : ByteArray) (p1 p2 i maxLen : Nat)
       (h1 : p1 + maxLen ≤ data.size) (h2 : p2 + maxLen ≤ data.size) : Nat :=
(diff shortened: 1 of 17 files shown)


kim-em:
## Benchmark result: this is a measured regression, not a speedup — converting to draft

Clean paired measurement (BEFORE = merge-base binary, AFTER = this branch, run back-to-back **pinned to the same idle core** so contention cancels, per-file geomean of `after/before`):

| corpus | levels | samples/lvl | overall speedup |
|---|---|---|---|
| Canterbury | 1–10 | 22 (11 files × 2 pairs) | **0.978×** (~2.2% slower) |
| Silesia | 1–8 | 24 (12 files × 2 pairs) | **0.953×** (~4.7% slower) |

The regression is uniform across levels. Even at L8 (deepest chains, the long-match regime where the win was predicted) most Silesia files regress (0.81×–1.02×, mostly ~0.95×); only a few large binaries (`mozilla`, `mr`, `x-ray`) reach ~1.0×.

**Note on an earlier artifact:** a first cross-session run showed a ~2.0× "speedup" uniform across all levels — that was pure CPU contention (BEFORE happened to run at ~33% CPU on a busy shared box, AFTER at ~66%; 66/33 ≈ 2.0). The pinned back-to-back paired numbers above are the real signal.

**Why it regresses:** the implemented `goUW` uses the *byte-scan fallback* variant — on a word mismatch it hands off to `goU`, which **re-scans** the bytes. So every short/medium match (the common case: most candidate matches are only a few bytes before the first mismatch) pays a wide-load + a full re-scan with no benefit, while only rare long matches get the 8-byte-advance payoff. The net is negative.

The correctness side is fine — `max |Δratio| = 0.000000` (output-neutral, as the `goUW_eq` proof guarantees) — so nothing is *wrong*; it's just slower.

**Path forward:** the issue title asks for "XOR + trailing-zero compare", i.e. locate the first differing byte directly from `(w1 XOR w2).ctz >>> 3` with **no re-scan**. That would cut the short-match overhead to a wide-load + XOR + ctz (~3 ops) instead of wide-load + re-scan, and is the variant most likely to actually win. Whether it flips the sign given DEFLATE's match-length distribution and the Lean→C call cost of the wide load is still an open question and would need re-measuring. Leaving as draft pending that rework decision.

kim-em:
## Reworked to ctz — before/after Pareto graphs (supersedes the regression finding above)

After the earlier byte-scan revision measured as a regression, I reworked the word-mismatch branch to locate the first differing byte with `ctz(w1 ^^^ w2) >>> 3` (no re-scan), the form the issue title asks for. Fresh clean measurement on `chungus`, both binaries (merge-base vs this branch) pinned back-to-back to an idle core, full native sweep 1-10, Canterbury median-of-5 / Silesia single-pass. `max |Δratio| = 0.000000` — output-neutral.

DEFLATE compression speed vs ratio, native before (grey) vs after (red), other-language curves reused for context:

**Silesia** (large real-world files — the corpus that matters):

silesia compress pareto before/after

**Canterbury** (small synthetic files):

canterbury compress pareto before/after

Native holds the **best ratios in the field** (leftmost points, L10 ≈ 0.308, bettered only by zopfli) at a mid-band speed — above OCaml, around JS `fflate`, below the C/Rust/Go engines. On Silesia the AFTER curve sits at-or-above BEFORE at every ratio (clearest at L6-L7); on Canterbury the two overlap. Same x at every point (output-neutral).

**Silesia** `compress_mbps` geomean over 12 files:

|----|------|-------|------|--------|
| 1 | 0.361 | 66.6 | 67.3 | 1.01× |
| 2 | 0.342 | 51.2 | 52.1 | 1.02× |
| 3 | 0.338 | 46.7 | 47.8 | 1.02× |
| 4 | 0.334 | 39.4 | 39.8 | 1.01× |
| 5 | 0.333 | 38.0 | 38.5 | 1.01× |
| 6 | 0.332 | 34.6 | 35.8 | 1.03× |
| 7 | 0.330 | 25.9 | 27.6 | **1.07×** |
| 8 | 0.323 | 9.6 | 9.7 | 1.01× |
| 9 | 0.315 | 3.3 | 3.4 | 1.03× |
| 10 | 0.308 | 1.7 | 1.7 | 1.02× |

**Canterbury** (median-of-5 over 11 small files): 1.00-1.02× across all levels — neutral. (A separate single paired run showed Canterbury -3.9%; the median-of-5 run puts it at ~+0.7%, so Canterbury is within its own run-to-run noise band, not a regression. Silesia is the consistent signal.) An independent tighter paired L1-8 run on a different idle core measured Silesia +4.9% overall, agreeing in sign.

Dashboard refreshed in this PR (`b3790389`); that commit also advances the recorded baseline past the deflate_fast L1 corner (#2726) which had landed after the previous snapshot.

Issue #2738, opened 2026-07-02
perf: chainless L1 match finder (ht-matchfinder style: 2-way buckets, min match 4, nice 32)


kim-em:
## Evidence

Silesia geomeans: native L1 66.4 MB/s @ 0.3614; libdeflate L1 271 MB/s @ 0.3549 (faster AND smaller); zig std.flate 111 @ 0.3394; zlib L1 123 @ 0.3751. Our L1 point sits well below the fast corner other implementations reach. perf at L1 (dickens): the chain machinery (`chainWalkPacked` 23% + `lz77ChainIterP_mainLoop` 14% + `updateHashesFastU` 9%) is ~47% of cycles, even at chain depth 4 / insertCap 2 (#2726).

## What libdeflate does (ht_matchfinder.h)

Level 1 uses no chains at all:
- One hash table, 15-bit hash (`lz_hash(load_u32 * 0x1E35A7BD, 15)`), **2 positions per bucket**.
- Minimum match length 4 (skips the barely-profitable length-3 matches entirely).
- At each position: probe at most the 2 bucket entries, early-out at `nice_len = 32`.
- Insertion is a rotate of the bucket pair; no `prev` array, no interior insertion loop.
- Blocks capped at 65535 bytes soft (`FAST_SOFT_MAX_BLOCK_LENGTH`).

Per position this is 2 bounded probes and one table write, versus our hash + chain walk + insertion-cap loop.

## Plan

1. New L1 matcher: 2-way bucket table (`Array UInt32` or a flat packed pair array), 4-byte hash via the existing `ugetUInt32LE` (`Zip/Native/Wide.lean`), min match 4, nice 32.
2. Emit through the existing packed-token pipeline; the matcher output satisfies the same `lzMatch` contracts (encodable/empty/resolves), so roundtrip proofs carry over.
3. Keep the current chain matcher at L2+; sweep whether L2 also wants the 2-way table with deeper buckets.

Relation to open work: #2734 (fast-L1 emit headroom) attacks the emit half of L1; this attacks the matcher half. Independent code, parallel-safe. #2726 landed the current chain-based corner; this replaces the mechanism rather than retuning it.

## Validation

`perf-pr-graphs` before/after. The L1 point should move sharply up; ratio may move slightly either way (min-match 4 loses length-3 matches but a 2-deep bucket search often finds better candidates than a depth-4 chain over a 3-byte hash).

🤖 Prepared with Claude Code

kim-em:
[...]

PR #2748, opened 2026-07-02
feat: verified chainless 2-way match finder (ht_matchfinder), unwired from L1

kim-em:
This PR adds `htMatch`/`htMatchIter`/`htMatchIterP`, a formally-verified chainless 2-way-bucket match finder (libdeflate `ht_matchfinder` style: two positions per bucket, minimum match 4, `niceLen` 32 early-out, interior insertion via `htInsert`), together with its correctness (`ValidDecomp` → `resolves`, `encodable`, `empty` in `Zip/Spec/HtMatchCorrect.lean`, plus the packed-twin equivalence), and records that it does not beat the tuned depth-4 hash chain at level 1, so `lzMatch`/`lzMatchP` keep the chain and the new matcher stays as a verified reference (sorry-free).

Issue #2738's premise was that the chain machinery being ~47% of L1 cycles made a cheaper matcher an easy speed win. A direct A/B (both binaries `taskset`-pinned to one core, back-to-back so contention is common-mode, native compress) refutes that. On **Silesia** (12 files, the representative corpus; `dickens` is #2738's own profiling target; master drift 1.005x) no `htInsertCap` beats the chain on either axis:

| `htInsertCap` | L1 speed vs chain | L1 ratio vs chain |
|---|---|---|
| 0 (fastest) | 0.96x (slower) | +5.8% (larger) |
| 258 (best ratio) | 0.61x (−39%) | +0.4% (still larger) |

The bucket cannot even reach ratio *parity* on Silesia at any speed cost — maxed-out insertion is still +0.4% larger than the chain while 39% slower, and the fastest config is already slower and larger. (Canterbury, small cache-resident text, is slightly kinder — cap 0 ties speed, cap 258 reaches −1.6% ratio — but it is not the target.)

Root cause: L1 is emit-bound (confirms #2726). Halving the probe count (cap 0 vs the chain's depth 4) buys no throughput because the token walk + `BitWriter`, not the match search, are the wall; any ratio loss is repaid as extra emit work. There is no runtime perf change here (the matcher is unwired), so no before/after graphs apply — the frontier above is the evidence, recorded in the `htMatch` docstring so #2738 is not re-attempted blind. OpenAI Codex, consulted independently, reached the same conclusion.

The verified matcher is kept because its `htBest_spec`/`htProbe_spec` "probe returns a `countMatch`-verified in-window match" pattern is reusable and the table-is-a-pure-heuristic structure let `htInsert` be added with zero proof churn. If you would rather ship nothing, close this and the finding lives in the #2738 thread.

🤖 Prepared with Claude Code

### DIFF
--- a/Zip/Native/Deflate.lean
+++ b/Zip/Native/Deflate.lean
@@ -1487,6 +1487,235 @@ where
   termination_by data.size - pos
   decreasing_by all_goals omega
 
+/-! ## Chainless 2-way-bucket matcher (libdeflate `ht_matchfinder` style)
+
+`htMatch` is the libdeflate `ht_matchfinder` scheme (#2738): a single hash
+table with **two positions per bucket** (`tbl[2*h]` = most-recent, `tbl[2*h+1]`
+= second-most-recent), **minimum match length 4** (skips the marginal length-3
+matches, which greedily cost more bits than the literals they replace), and an
+early-out once a probe reaches **`niceLen = 32`**. Per position it does at most
+two bounded `countMatch` probes (versus the chain's depth-4 fuel walk over a
+`prev` array) and one bucket rotation (`lru := mru; mru := pos`). After a match
+it inserts the interior positions into the 2-way table (`htInsert`, like
+libdeflate's skip, bounded by `htInsertCap`) so they stay findable.
+
+**Measured, and NOT selected for L1.** #2738's premise was that the chain walk
+being ~47% of L1 cycles made a cheaper matcher an easy speed win. A direct A/B
+(both binaries pinned to one core, back-to-back, native compress) refutes this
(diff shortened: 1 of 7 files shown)


kim-em:
Closing unmerged. The chainless 2-way matcher is fully implemented and formally verified, but a direct same-core A/B on Silesia shows it is worse than the tuned depth-4 chain on **both** axes at every insertion cap (cap 0: 0.96x speed / +5.8% ratio; cap 258: 0.61x speed / +0.4% ratio — cannot reach ratio parity at any speed cost). L1 is emit-bound (#2726): cap 0 does half the chain's probe work yet runs slower, so matcher microcost is not the bottleneck. Landing ~660 lines of verified-but-unused code is not worth the maintenance cost; the negative result and its evidence live in the #2738 thread. See #2738 for the full frontier and disposition.

Issue #2739, opened 2026-07-02
perf(explore): attribute and cut the RC/allocator tax (17% of L1, 15% of L8 compress cycles)


kim-em:
## Evidence (perf, steady-state `compress-pareto`, 2026-07-02, chungus)

Reference counting plus allocator time is a large, mostly flat tax on compression:

| level (file) | `lean_dec_ref_cold` | `mi_free` + `mi_malloc_small` | total |
|---|---|---|---|
| L1 (dickens) | 8.9% | 7.8% | **16.7%** |
| L6 (dickens) | 4.4% | 3.1% | 7.5% |
| L8 (xml) | 8.6% | 6.2% | **14.8%** |
| L9 (xml) | 5.5% | ~1% | 6.5% |

That is a fifth-to-sixth of the whole compress budget going to memory management, and it is nobody's algorithm.

## Step 1: attribute it

The flat profile does not say which allocations these are. Candidates:
- `ByteArray.push` growth in `BitWriter.flushAcc` output (one push per byte; doubling reallocs copy the whole stream). The pre-sized output buffer belongs to #2631 (BitWriter wide-store); this issue should only confirm its share.
- Packed-token `Array UInt32` growth by `push` in the matcher main loops (pre-sizing by a bytes-per-token heuristic is cheap; #2607 pre-sized one of these paths, check coverage).
- Boxed intermediates on the L8 sizing path (`List.setTR` visible at 2.2%; superseded if the observation-splitting issue lands, but attribute first).
- Per-call constructor churn in hot recursive helpers that fail the reuse optimization.

Getting the attribution: `perf record --call-graph dwarf` on the mi_* symbols, or bisection by hacking a pre-size into one site at a time and re-measuring `compress-pareto`.

## Step 2: fix the top sites

Likely shape: `Array.mkEmpty`/`ByteArray.mkEmpty` capacity reservations at allocation sites, and ctor-reuse-friendly restructuring where the reuse analysis is falling out. Each fix is output-preserving, so proofs are untouched; each must clear a measured win on `compress-pareto` before landing.

Boundary: the BitWriter output buffer restructure stays in #2631; this issue covers everything else the attribution surfaces.

🤖 Prepared with Claude Code

kim-em:
## Step 1 done: attribution (branch sampling, chungus, steady-state compress-pareto)

[...]

Issue #2737, opened 2026-07-02
perf: observation-divergence block splitting at L4-L8 (per-block trees below L8; retire the L8 sizing pass)


kim-em:
## Evidence

Two problems, one mechanism:

1. **The mid-band is dominated.** Levels 1-7 emit a single DEFLATE block for the whole file (dispatch at `Zip/Native/DeflateDynamic.lean:1380-1416`): one Huffman tree pair for a 10-20 MB Silesia member. zlib refits trees every ~16K symbols. Current silesia geomeans: native L5 0.3330 @ 38 MB/s, L6 0.3320 @ 35, L7 0.3300 @ 27, all dominated by zlib L5 (0.3297 @ 56). Most of that ratio gap is whole-file trees, not match quality: native L8 (which splits) reaches 0.3228.
2. **L8's split machinery is expensive.** perf at L8 (xml): `findTableCode_go` 13.8% (a linear scan of the length/distance slot tables, `Zip/Native/Deflate.lean:129`, on the sizing path even though dense packed tables exist for emit), boxed `tokenFreqs_go` 1.8%, `List.setTR` 2.2%, `chooseSplitsHeuristic` forIn 1.9%. The sized-split candidate pays a boxed sizing pass the dense-table emit path already made obsolete.

## What libdeflate does (deflate_compress.c:2055-2217)

Block boundaries are chosen by a cheap streaming heuristic, no sizing passes:
- Each symbol is bucketed into one of 10 coarse observation classes (8 literal classes from top-2 bits + low bit; 2 match classes, length < 9 vs >= 9).
- Every 512 symbols, if the block is at least 5000 bytes, compare the last 512 observations against the block's running distribution (sum of absolute frequency deltas, integer-scaled; cutoff 200/512 with a short-block penalty, `deflate_compress.c:2141-2197`).
- Divergence => end block. Soft max block 300000 bytes, min 5000.

## Plan

1. Implement the observation counters + divergence check over the packed token stream (or inline in the matcher main loop). Pure `Nat`/`UInt32` arithmetic, heuristic-only: it picks boundaries, the existing emitters guarantee correctness.
2. Feed the boundaries to the existing cross-block shared-window emit (`emitSharedBlocks`, roundtrip proof already in place), replacing the fixed `sharedTokChunk` cadence.
3. Enable the split path at L4-L7 (currently single-block).
[...]

PR #2749, opened 2026-07-02
perf: attribute the RC/allocator tax; ctor-reuse the BitWriter (+9.9% L1, +3.8% L8)

kim-em:
This PR attribute the compress-path RC/allocator tax (#2739) with hardware branch sampling and land the top fix: constructing the result `BitWriter` in `writeBits`/`writeHuffCode` instead of inside the `flushAcc` loop, so the compiler's ctor-reuse recycles one writer cell per stream instead of allocating and freeing one per field written. Measured on chungus (`compress-pareto`, pinned `taskset -c 43`, interleaved A/B vs master): L1/dickens 51.6 → 56.7 MB/s (+9.9%), L8/xml 21.1 → 21.9 MB/s (+3.8%), L6/dickens 24.5 → 25.4, L9/xml 2.93 → 3.01; output byte-identical everywhere. Proof cost is small: `flushAcc` and `flushAcc_spec` stay as the reference form, `flushAcc_eq` proves the split (`flushBytes`/`dropBytes`) form equal, and `writeBits_def`/`writeHuffCode_def` restate the old defining equations for the four unfold sites in `BitWriterCorrect`.

Attribution notes (in `progress/rc-allocator-tax-2739.md`): DWARF unwinding produces no usable call graphs on leanc-built binaries, but branch sampling (`perf record -j any_call,u`) attributes every sampled call into `mi_malloc_small`/`mi_free`/`lean_dec_ref_cold` to its exact caller. On L1 the tax was ~59% BitWriter ctor churn and ~41% `updateHashes` result-pair churn; on L8 the boxed split path (`findTableCode` scans + Option/tuple churn, `unpackTok` map, `List.setTR`) dominates alongside the same two shapes. The #2631 pre-size share is ~zero (2 branch samples of push-doubling reallocs out of 45k), so that issue's win must come from wide stores, not fewer reallocs.

A second fix (threading the matcher chain state as one pair for ctor reuse) was built, proven, and measured **negative** (−1.9% L1): the reuse converts one alloc+free per match into an isShared check plus two cell stores per position, which costs more than it saves. It is documented and retired, not landed. The remaining L8 sizing-path churn is handed to #2737 / stage D with the numbers.

🤖 Prepared with Claude Code

### DIFF
--- a/Zip/Native/BitWriter.lean
+++ b/Zip/Native/BitWriter.lean
@@ -30,7 +30,15 @@ def bitLength (bw : BitWriter) : Nat := bw.data.size * 8 + bw.bitCount.toNat
 
 /-- Flush whole bytes out of a wide accumulator `acc` holding `total` valid
     low bits, LSB-first. Pushes `total / 8` bytes to `data`; the remaining
-    `total % 8` bits become the new partial byte. -/
+    `total % 8` bits become the new partial byte.
+
+    Reference form only: the production writers use the split
+    `flushBytes`/`dropBytes` loops (equal by `flushAcc_eq`, which the
+    `writeBits_def`/`writeHuffCode_def` equations build on) so the `BitWriter`
+    cell is constructed in the caller, where the ctor-reuse optimization can
+    recycle the input writer instead of allocating a fresh cell per field
+    (measured at ~59%/~19% of `mi_malloc_small` calls on L1/L8 compress,
+    #2739). -/
 def flushAcc (data : ByteArray) (acc : UInt64) : Nat → BitWriter
   | total =>
     if total ≥ 8 then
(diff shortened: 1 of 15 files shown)


kim-em:
## Silesia Pareto refresh (pinned)

Dashboard refreshed with the native-only splice, measured pinned to one CPU (`BENCH_PIN=43`; pinning is now built into `bench/run.sh` for all measurement passes). `out_size` is identical on **every** native row, so each point moves straight up at constant ratio — outside the previous native mixing frontier by construction.

silesia compress pareto

Silesia per-level geomean compress throughput (12 files):

| level | before | after | Δ |
|---|---|---|---|
| 1 | 66.4 MB/s | 72.0 MB/s | **+8.5%** |
| 2 | 52.4 | 55.3 | +5.6% |
| 3 | 48.2 | 50.6 | +4.9% |
| 4 | 39.0 | 40.8 | +4.5% |
| 5 | 38.0 | 39.5 | +4.0% |
| 6 | 35.1 | 36.5 | +3.8% |
| 7 | 27.2 | 28.0 | +3.0% |
| 8 | 9.45 | 9.85 | +4.2% |
| 9 | 3.32 | 3.36 | +1.1% |
| 10 | 1.71 | 1.71 | +0.1% |

The gradient matches the attribution: low levels are emit-bound (where the BitWriter ctor churn lived), the optimal tiers (9–10) are match-finding-bound so the emit fix barely registers there.

Canterbury moves the same way:

canterbury compress pareto

🤖 Prepared with Claude Code

kim-em:
## Before/after `compress_mbps` on the merged branch (machine `Linux chungus x86_64`)

Master moved under this PR (#2744 niceLen ladder, #2746 word-at-a-time countMatch landed), so this re-runs the gate on the merge: BEFORE is a **merge-base build** (the committed baseline predated #2744 — freshness guard caught it), AFTER is this branch merged with master; both swept back-to-back, pinned via `bench/pin_core.sh`. `max |Δratio| = 0.000000` (output-neutral), untouched levels 9–10 at 1.00–1.01x (binary health OK).

silesia compress pareto before/after

Silesia geomean (12 files):

|---|---|---|---|---|
| 1 | 0.361 | 66.0 | 73.0 | **1.11x** |
| 2 | 0.349 | 52.2 | 56.3 | 1.08x |
| 3 | 0.344 | 48.1 | 51.6 | 1.07x |
| 4 | 0.335 | 39.0 | 41.4 | 1.06x |
| 5 | 0.335 | 38.0 | 40.1 | 1.06x |
| 6 | 0.332 | 34.8 | 36.7 | 1.06x |
| 7 | 0.330 | 26.9 | 28.1 | 1.05x |
| 8 | 0.323 | 9.5 | 9.9 | 1.04x |
| 9 | 0.315 | 3.4 | 3.4 | 1.01x |
| 10 | 0.308 | 1.7 | 1.7 | 1.00x |

canterbury compress pareto before/after

Canterbury: 1.08x L1 tapering to 1.00x L10, same shape.

Visual read: the AFTER curve lifts uniformly at the emit-bound end and converges at the crown; on Silesia native's L1 point now sits at/above the JS fflate curve at better ratio, with zlib/Go/miniz still ahead and libdeflate the SIMD ceiling. The win is slightly **larger** than measured pre-merge (+11% vs +9.9% at L1): #2744/#2746 shrank match-finding time, so the emit share this PR removes weighs more.

Note: the delta is a touch larger than the interleaved corner A/B suggested partly because BEFORE here is a single-pass cross-binary sweep — but the same-session, same-core pairing and the level-graded structure (largest where emit dominates, zero at the crown) match the attribution, so I read it as real.

The dashboard commit in this PR also brings `bench/results/latest.json` current (the committed baseline was recorded on the #2746 branch and predated #2744).

🤖 Prepared with Claude Code

PR #2750, opened 2026-07-02
perf: L9/L10 cache-build skip-ahead at nice length

### LINKED ISSUE #2740: perf: L9/L10 cache-build skip-ahead at nice length (match finding is ~70% of the optimal tiers)
## Evidence (perf, `compress-pareto`, xml, L9)

The L9 pipeline spends ~70% of its cycles building the candidate cache: `chainWalkAllFast` 50.7% + `lz77Greedy.countMatch` 18.1% + `buildCache` 3.2%. The DP itself is nearly free (`fillRegionFast` 2.3%, `scanCandsFast` 5.0%). L9 runs at 3.3 MB/s and L10 at 1.7 MB/s on silesia while libdeflate reaches the same L9 ratio (0.3146 vs our 0.3148) at 49 MB/s. The match *finder*, not the parser, is the entire tail problem.

## What libdeflate does (deflate_compress.c:3651-3754)

During near-optimal match collection, when a match of length >= `nice_len` is found at a position, it records that one match and then **skips the interior**: the following `match_len - 1` positions are inserted into the matchfinder (cheaply, no search: `bt_matchfinder_skip_byte`) but never searched, and nothing is cached for them. Long repeated regions cost one search instead of hundreds. Matches are also only cached up to `nice_len` per position (75/150/258 at levels 10/11/12).

## Plan

1. In `buildCache` (`Zip/Native/DeflateParse.lean`), when the deepest cached match at a position reaches a `nice` threshold (start at 258, then sweep lower), skip searching the interior positions: hash-insert only, cache nothing (the DP already handles positions with no cached candidates by taking a literal or a shorter step; verify the region DP still sees the long match's coverage, since choices are per-position).
2. The choice arrays are untrusted heuristics and the emitter re-verifies every match with `lz77Greedy.countMatch`, so this is ratio-risk only, zero correctness risk. Roundtrip proofs unchanged.
3. Sweep `nice` per tier (L9 vs L10) on the silesia geomean.

Synergies, not dependencies: the wide `countMatch` issue speeds up the searches that remain; the dual hash4+hash3 issue reduces false candidates per chain probe. All three compose on the L9/L10 tail.

## Validation

`perf-pr-graphs` before/after; expect L9/L10 to move up (target: several x on match-heavy members like nci and xml) with geomean ratio change well under 0.5%.

🤖 Prepared with Claude Code

kim-em:
This PR adds libdeflate's nice-length skip-ahead to the near-optimal parser's candidate-cache build (closes #2740): once the deepest match the chain walk finds at a position reaches a per-tier threshold, the match interior — the next `bestLen - 1` positions, clamped to the region end — is only hash-inserted into the chains (`insertChainRange`), never searched or cached, and the search resumes past it. A long repeated region then costs one chain walk instead of one per byte, and the match finder was ~70% of the L9/L10 cost.

The chain walk already threads its running best match length, so it now returns that as a third tuple component (`chainWalkAll`/`chainWalkAllFast`/`chainWalkAllGuarded` return `Array Nat × Array Nat × Nat`) and the skip decision costs nothing beyond the walk itself — no per-position rescan of the cache slots, which would otherwise tax every position and lose the win on inputs with few long matches (measured: rescanning turned a small win into a net slowdown on binary corpora).

The thresholds are swept per tier, since the L9-fast matcher is shallower (`fastChainDepth = 64`, single round) and gains less from skipping while costing proportionally more ratio: `optNiceSkip = 64` for the L10 exact DP, `fastNiceSkip = 128` for L9-fast. On the Silesia geomean, L10 gains ~2.5x on xml and ~1.8x on nci for +0.3% geomean ratio; L9 gains ~1.5x on xml for +0.45% geomean, both neutral on low-match text/binary members.

The choice arrays stay untrusted heuristics that `optimalEmit` re-verifies with `countMatch` (falling back to a literal), and the `ValidDecomp`/encodability proofs quantify over arbitrary choice arrays, so this is ratio-risk only with zero correctness risk — the roundtrip proofs are unchanged. A second opinion (Codex) confirmed the chain-insertion, DP-validity, termination, and best-length-threading all hold.

🤖 Prepared with Claude Code

### DIFF
--- a/Zip/Native/DeflateParse.lean
+++ b/Zip/Native/DeflateParse.lean
@@ -37,6 +37,30 @@ def optNiceLen : Nat := 258
     ratio from candidate *choice* than raw depth buys. Pure ratio/speed knob. -/
 def optChainDepth : Nat := 256
 
+/-- Skip-ahead threshold for `buildCache` (#2740, libdeflate's
+    `nice_len` skip): once the deepest candidate recorded at a position
+    reaches this length, the match interior — the next `bestLen - 1`
+    positions — is only hash-inserted into the chains, never searched or
+    cached. A long repeated region then costs one chain walk instead of one
+    per byte (the match finder is ~70% of the L9/L10 cost). The DP still sees
+    the long match at its start position, so it can take it and jump the
+    interior; interior positions with no cached candidate fall back to a
+    literal or a shorter step. Pure ratio/speed knob — the emitter re-verifies
+    every match, so this is ratio-risk only.
+
+    Level 10 (exact DP). Swept on the Silesia geomean: `64` sits at the knee —
+    ~2.5× on xml, ~1.8× on nci for +0.3% geomean ratio; dropping to 48 falls off
+    a cliff (nci +5.3%) as length-48–64 matches start being skipped. -/
+def optNiceSkip : Nat := 64
+
+/-- `optNiceSkip` for the L9-fast tier (`computeChoicesFast`). Separate from
+    the level-10 value so each tier's speed/ratio balance can be swept
+    independently. The L9 matcher is shallower (`fastChainDepth = 64`, single
+    round), so skipping saves less and costs proportionally more ratio; `128`
+    keeps the Silesia geomean under budget (~+0.45%) while still buying ~1.5× on
+    xml. Pure ratio/speed knob. -/
+def fastNiceSkip : Nat := 128
+
 /-- DP region size in bytes. The cost/cache arrays live per region (memory
     cap); cost continuity across the region end is approximated by a baseline
     tail estimate, and matches may extend past the region end, so the region
@@ -72,9 +96,9 @@ arrays or pairs (boxing). -/
     distance), given that the DP also evaluates truncated lengths. -/
 def chainWalkAll (data : ByteArray) (prev : Array Nat) (pos maxLen : Nat)
     (hpm : pos + maxLen ≤ data.size) (cand fuel bestLen k slotBase : Nat)
-    (lens dists : Array Nat) : Array Nat × Array Nat :=
-  if fuel = 0 then (lens, dists)
-  else if optCacheSlots ≤ k then (lens, dists)
+    (lens dists : Array Nat) : Array Nat × Array Nat × Nat :=
+  if fuel = 0 then (lens, dists, bestLen)
+  else if optCacheSlots ≤ k then (lens, dists, bestLen)
   else if hc : cand < pos ∧ pos - cand ≤ 32768 then
     have hcand : cand + maxLen ≤ data.size := by omega
     -- Prefilter (#2622): mirror of `chainWalkAllFast` — skip the compare when the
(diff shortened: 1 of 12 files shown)


kim-em:
## Before/after `compress_mbps` (machine `chungus`) — final, rebased onto current master

> **Two rounds of the #2751 stale-binary trap, now cleared.** My first graphs read untouched L1–8 at ~0.88× (I wrongly blamed machine load); the real cause was a stale `.lake` (bare-shell `lake build` then `lake -R` under nix-shell) handicapping the AFTER binary. After rebasing onto current master (which had gained #2744 chain-walk early-out + #2746 word-at-a-time countMatch) a *second* stale binary briefly made an untouched level's **output** differ (xml L2). Both are resolved by `rm -rf .lake` + full rebuild; I now verify untouched-level **output** byte-for-byte (not just speed) before trusting any run. All numbers below are from clean-built, output-verified binaries, same-session, pinned via `bench/pin_core.sh`, at load ~1.3.

### Isolation check (output, deterministic)
Untouched levels use `deflateRawBase`; their compressed output is **byte-identical** before/after (e.g. xml/nci L2, L6 SAME), and only L9/L10 output changes. `max |Δratio|` = 0.0024 at silesia/xml L9 — the intended skip-ahead effect, nothing else.

### Paired speed — same core, correct binaries (1 MB slices)
| file | L9 (fast) | L10 (crown) |
|---|---|---|
| xml | **1.37×** | **2.28×** |
| nci | 1.09× | **1.73×** |
| dickens (low-match control) | 0.98× | 0.99× |

Consistent with the pre-rebase paired run (xml L10 2.31×) — the win survives #2746's faster countMatch: the word-at-a-time extension speeds the searches that remain, the skip removes the ones that do not.

### Full-corpus silesia geomean (clean before/after, same session)
| lvl | before | after | speedup |
|---|---|---|---|
| 1–8 (untouched) | — | — | 0.99–1.01× |
| 9 (fast) | 3.3 | 3.5 | **1.06×** |
| 10 (crown) | 1.7 | 2.1 | **1.20×** |

Geomean dilutes the per-file win (low-match members are neutral); match-heavy data carries it.

### Ratio cost (deterministic)
Silesia geomean ratio delta **L10 +0.30% / L9 +0.45%** (worst file xml: +0.8% L10 / +3.3% L9), within the <0.5% budget. Emitter re-verifies every match → ratio-only, zero correctness risk. Codex second opinion confirmed chain-insertion, DP-validity, termination, and best-length threading.

### Overlay (clean native before/after vs the field)
canterbury
silesia

Untouched L1–8 sit on top of the before curve; the **L10 crown (leftmost) is visibly above before** at the best ratio of the field except zopfli.

### Dashboard refresh — included
`bench/results/latest.json` + `bench/graphs/*.svg` refreshed via `bench/run.sh --native-only` on the quiet machine (meta commit = this branch). Besides the L9/L10 win, this brings the native L2–L7 rows current — they had drifted from the committed snapshot, whose meta predates #2744's chain-walk early-out ladder.

Issue #2743, opened 2026-07-02
perf: per-block stored/fixed/dynamic choice on the split path


kim-em:
depends-on: #2737

## Problem

The whole-file path already does the three-way cost comparison (stored / fixed / dynamic) before emitting: `deflateRawBaseP` sizes all three and picks the smallest (`Zip/Native/DeflateDynamic.lean:1091-1103`). The block-split paths do not: every split block is emitted as a dynamic block. On a heterogeneous file, blocks that are nearly incompressible (should be stored) or tiny/skewed (fixed can beat dynamic including the tree header) pay the dynamic-tree overhead.

## What libdeflate does (deflate_compress.c:1729-1858)

At each block flush it computes exact bit costs for all three block types from the block's symbol frequencies (dynamic: header + precode + coded lengths; static: fixed code lengths; stored: alignment + 32 + 8 * length) and emits the cheapest, falling back to multiple 65535-byte stored blocks when stored wins.

## Plan

1. In the per-block emit loop that #2737 establishes, compute `fixedBlockBytes` / `dynBlockBytes` / stored cost per block from the block's frequencies (all three sizing functions already exist and `dynBlockBytesWith` already shares the header plan with emit, the #2627 dedup).
2. Emit the winner per block. Stored blocks within a split stream need the byte-alignment handling that `deflateStoredPure` already implements whole-file; the per-block variant threads the same logic through the shared `BitWriter`.
3. Extend the shared-blocks roundtrip proof by one case split per block type (the three emitters each have their own decode lemmas already; the composition lemma gains the choice).

Pure ratio win (left move) at L4-L8 on mixed-content members (mozilla, samba, osdb); speed cost is two extra sizing computations per block over frequencies already counted.

Depends on #2737 (observation-divergence block splitting): the per-block choice point only exists once the split emit loop does.

## Validation

`perf-pr-graphs` before/after; per-file ratio table for the heterogeneous silesia members, expecting the largest deltas there and zero regression on homogeneous text.

🤖 Prepared with Claude Code

kim-em:
Closing as not planned: measured the ratio before doing the (substantial) proof work, and it isn't there.

## What I measured

[...]

Issue #2552, opened 2026-06-12
perf: reuse the winning partition's sized trees at shared-split emission (Wave 3 Step 0.3)


kim-em:
**Wave 3 Step 0.3 (re-scoped after code reading) — reuse the winning partition's sized trees at emission.**

### Current state

At L≥7, `chooseSplitsArbitrated` (Zip/Native/DeflateDynamic.lean:649) sizes BOTH candidate partitions via `sharedPartitionBits` — per block: `tokenFreqs (toks.extract pos j)` + `dynamicCodeLengths` (a full Huffman build) + `writeDynamicHeader` into an empty writer. The winning cut list then goes to `emitSharedBlocksAt`, whose `emitSharedBlock` (:406) recomputes the same per-block freqs + trees + header a second time. The losing partition's sizing pass is inherent to arbitration; the winner's recompute is pure duplication (~one of three per-block tree passes at L7–8 ≈ 5–8% of L7 wall time on dickens).

### Deliverables

1. A tree-taking emitter variant (e.g. `emitSharedBlocksAtSized`) that consumes the per-block `(litLens, distLens)` (with their length invariants) computed during arbitration, plus a `chooseSplitsArbitratedSized` returning cuts together with the winner's sized trees. The existing cut-generic emitter stays as the reference; prove the sized variant equal to it (the trees passed in are definitionally `dynamicCodeLengths (tokenFreqs group)` for the same group — an rfl-or-near lemma per block).
2. Wire into `deflateRaw`'s L≥7 branch; byte-identical output (ratio canaries + the SizeHelpers conformance test for `sharedPartitionBits` must stay green).
3. Per-level timing before/after at L7/L8 on alice29 + dickens in the PR.

### Context

Wave 3 Step 0.3 of the performance pivot (plan of 2026-06-12). Spec impact: the `deflateDynamicBlocksSharedAt` quadruple in Zip/Spec/DeflateBlockSplit.lean unfolds the emitter — the sized variant needs either the equality-lemma route (preferred; statements untouched) or a parallel quadruple. Lessons in #2540's closing comments apply (no content-invariant lemmas; no fun_induction over heuristic defs).

### Verification

`lake build && lake exe test` green; 0 sorries; byte-identity canaries unchanged (alice29 52049/54744); timing table in PR; no dashboard refresh needed (iso-output).

kim-em:
Session: `unknown` Branch: `master`

PR #2745, opened 2026-07-02
perf: observation-divergence block splitting at L4-L8 on packed tokens

kim-em:
This PR implements libdeflate-style observation-divergence block splitting across the mid-band (closes https://github.com/kim-em/lean-zip/issues/2737): levels 4–7, which emitted a single DEFLATE block (one Huffman tree pair for a whole Silesia member), now try a cross-block shared-window split whose boundaries come from a streaming divergence heuristic, and level 8's exact-bits sizing/arbitration pass (`chooseSplitsArbitrated` + `deflateDynamicBlocksSharedSized`, ~18% of L8 cycles in `findTableCode` linear scans and boxed `tokenFreqs`) is retired from the hot path in favour of the same mechanism.

The new pipeline runs on packed tokens end-to-end: `chooseSplitsHeuristicP` buckets each packed token into one of 10 coarse observation classes and cuts where the recent 512-token window's distribution diverges from the block's (same constants as the boxed heuristic from #2528, which stays as the conformance reference); `deflateDynamicBlocksSharedAtP` emits the blocks with `tokenFreqsP` + `emitTokensWithCodesP` (dense packed tables, no boxed tokens, no linear scans). When the heuristic proposes no cuts — every input under ~2·`splitMinBlockBytes` — the dispatch skips the split candidate entirely, so small inputs pay nothing; otherwise `pickSmaller` against the single-block base bounds any heuristic misfire at the old ratio. At level 8 the fixed-cadence partition is additionally *emitted* as a third candidate: the min over emitted candidates decides by the same quantity the retired sizing pass computed (⌈bits/8⌉), so level 8 is provably never worse than the old arbitration on any input (measured, the heuristic alone lost only silesia/sao, +0.6pp — the cadence candidate closes exactly that hole at one packed emit pass).

The heuristic carries no proof obligations: the emitter clamps every cut, the `DeflateBlockSplit` roundtrip quadruple is quantified over an arbitrary selector, and `deflateDynamicBlocksSharedAtP_eq` (plus per-token field lemmas `splitTokenClassP_eq`/`splitTokenBytesP_eq` and the emitter equalities) reduces the packed pipeline to the boxed reference, so `inflate_deflateRaw` and the padding theorems rewire with no new bit-level reasoning. The retired arbitrated pipeline keeps its definitions, proofs, and `SizeHelpers` conformance tests.

[...]

PR #2754, opened 2026-07-03
perf: size-arbitrate the L6-L8 split candidates, emit only the winner

### LINKED ISSUE #2753: perf: size-arbitrate the split-path candidates, emit only the winner (recover the L5-L7 double-emit cost)
## Problem

The #2745 block splitting bought L5-L7's ratio wins with a second full emit pass: `deflateRaw` at L4-L7 now emits **both** the single-block base and the observation-split candidate, then `pickSmaller` discards one. Measured cost on the silesia geomean (dashboard, 2026-07-03): L5 -17%, L6 -38%, L7 -31% compress speed versus the pre-split points, with L4 unaffected (its split rarely engages). At L8 there are three emitted candidates (base, obs-split, fixed-cadence).

## Fix: arbitrate by size, emit once

All the sizing machinery already exists and is O(frequencies), not O(output):
- base: `fixedBlockBytes` / `dynBlockBytesWith` / `storedBlockBytes` (already computed inside `deflateRawBaseP` before it emits).
- split candidate: per-block `dynBlockBytesWith` over the cut list's per-block frequencies, summed (the retired L8 arbitration computed exactly this quantity; the frequency splits are one pass over the packed tokens).

So: size every candidate, emit only the winner. The emit pass runs once instead of twice (thrice at L8). Expected recovery is most of the 17-38%, minus one extra frequency pass per candidate.

## Correctness

Byte-count agreement between the sized and emitted forms is currently conformance-tested (`SizeHelpers` tests); #2604 is proving the counting identities (`fixedBlockBytes`/`dynBlockBytes` = emitted `.size`). This issue does not depend on #2604: `pickSmaller` semantics are preserved as long as the size models are exact, which the existing conformance tests pin per block type. If #2604 lands first, the arbitration becomes proven rather than tested; either order works.

Output should be byte-identical to the current dispatch on every input (same winner, one emit), which makes validation sharp: assert equality against the current two-emit dispatch across the corpora, then measure speed only.

Relation: #2743 (per-block stored/fixed/dynamic choice) adds a per-block decision *inside* the split emit; this issue removes whole-candidate emits. They compose but touch the same dispatch, so they should land in sequence, either order.

## Validation

`perf-pr-graphs` before/after: L5-L7 move straight up at exactly unchanged ratio; L4/L8 same-or-faster.

🤖 Prepared with Claude Code

kim-em:
This PR replaces the level 6-8 `deflateRaw` dispatch that emitted every split candidate (base + observation-divergence split, plus the fixed-cadence split at L8) and kept the smallest via `pickSmaller` with a size-arbitrated one that emits only the winner. Each candidate is *prepared* — sized to its flushed byte count with its per-block Huffman trees captured — and only the winner's emit thunk is forced, reusing the trees the sizing pass already built. Output is byte-identical to the old dispatch on every input (same winner decided by the same ⌈bits/8⌉ counts), so the compression ratio is unchanged; the saving is the discarded emit passes (a second full emit at L5-L7, a third at L8).

The tree capture is load-bearing. A naive size-then-emit-winner *regresses*, because sizing a dynamic block is not cheap — it costs a full frequency pass plus the per-block Huffman-tree build and `writeDynamicHeader`, roughly 0.72x the cost of emitting it. Reusing the sized trees for the winner's emit (so the winner skips the second frequency/Huffman pass) is what turns it into a real win. Measured on the Silesia corpus with clean nix-shell builds (a stale `.lake` from a stray bare-shell build imposes a flat ~10% handicap that masks the effect entirely — rebuild from a clean `.lake`):

| level | geomean | range |
|------|---------|-------|
| L5 | ~0% (unchanged, no split path) | — |
| L6 | ~+6.7% | +3.8% (samba) … +12.2% (osdb) |
| L7 | ~+6.5% | +3.3% (xml) … +13.2% (osdb) |
| L8 | ~+9.7% | +5.8% (sao) … +15.3% (dickens) |

Correctness is preserved at two levels. `emitSmallerBy` / `deflateRawBasePPrep` / `deflateDynamicBlocksSharedAtSizedP` are pinned byte-identical to the retired `pickSmaller`-of-emitted-blocks path by the `SizeHelpers` conformance tests (each sizer's `.1` equals the emitted `.size`, each emit thunk equals the reference emitter, and `deflateRaw` at L6/L7/L8 is asserted byte-identical to the old emit-every-candidate dispatch across the sample set) — these tests were previously compiled-but-never-run, and this PR wires them into the test runner. The unified roundtrip, padding, and `goR` proofs in `DeflateRoundtrip` go through `deflateRawBasePPrep_emit` and `deflateDynamicBlocksSharedAtSizedP_emit`; the latter (packed sized-tree emit equals the reference packed emit) is proved in `LZ77PackedCorrect` as the packed twin of the Wave-5 (#2552) boxed pipeline.

Relation: composes with #2743 (per-block stored/fixed/dynamic choice inside the split emit) — both touch this dispatch, so they should land in sequence, either order.

Closes #2753.

🤖 Prepared with Claude Code

### DIFF
--- a/Zip/Native/DeflateDynamic.lean
+++ b/Zip/Native/DeflateDynamic.lean
@@ -463,6 +463,18 @@ open Zip.Spec.DeflateStoredCorrect (deflateStoredPure storedBlockBytes storedBlo
 def pickSmaller (a b : ByteArray) : ByteArray :=
   if a.size < b.size then a else b
 
+/-- Lazy `pickSmaller` keyed on precomputed byte sizes: forces (emits) only the
+    winning candidate. When `sa = (a ()).size` and `sb = (b ()).size` this equals
+    `pickSmaller (a ()) (b ())` byte-for-byte (`a` wins iff strictly smaller, ties
+    keep `b`), but the losing candidate is never emitted — the whole point of
+    sizing every candidate before emitting any. Correctness of the byte-identity
+    (that `sa`/`sb` are the true emitted sizes) is pinned by the `SizeHelpers`
+    conformance tests, exactly as the single-block `fixedBlockBytes`/`dynBlockBytes`
+    dispatch is; the roundtrip theorems only need each candidate to decode. -/
+@[inline] def emitSmallerBy (sa : Nat) (a : Unit → ByteArray)
+    (sb : Nat) (b : Unit → ByteArray) : ByteArray :=
+  if sa < sb then a () else b ()
+
 /-! ## Sizing a block without emitting it
 
 A DEFLATE block's body is a *dot product* of symbol frequencies and code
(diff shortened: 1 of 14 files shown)


kim-em:
Before/after `compress_mbps`, native sweep levels 1-10 on `Linux chungus x86_64` (BEFORE = merge-base `be4eed6a`, the #2745 split; AFTER = this branch). Output is byte-identical (`max |Δratio| = 0.000000`), so the curves shift straight **up** at unchanged ratio. Control levels (L1-L5, L9, L10 — no split path) sit at 0.99-1.01x, confirming this is the change and not a build artifact.

**Silesia** (12 files, the corpus that matters — geomean):

|------:|------:|------------:|-----------:|:-------:|
| 5 | 0.331 | 30.3 | 31.2 | 1.03x |
| **6** | 0.325 | 21.2 | 23.3 | **1.10x** |
| **7** | 0.323 | 18.6 | 19.8 | **1.06x** |
| **8** | 0.323 | 12.3 | 13.5 | **1.10x** |
| 9 | 0.316 | 3.5 | 3.5 | 1.00x |
| 10 | 0.309 | 2.0 | 2.1 | 1.01x |

**Canterbury** (11 small files — smaller wins, cache-resident): L6 1.03x, L7 1.02x, L8 1.06x; control levels ~1.0x.

silesia compress pareto

canterbury compress pareto

Positioning: lean-zip native sits mid-field (above OCaml, below the C/Rust/Zig SIMD codecs); this PR nudges the L6-L8 mid-band up toward the JS fflate edge of the faster cluster, at unchanged ratio. The dashboard (`bench/results/latest.json` + compress graphs) is refreshed in this PR.

PR #2755, opened 2026-07-03
perf: USize-native chain walk (untag the Nat position arithmetic in chainWalkPacked)

### LINKED ISSUE #2741: perf: USize-native chain walk (untag the Nat position arithmetic in chainWalkPacked)
depends-on: #2735

## Evidence

`perf annotate` on `chainWalkPacked` (L6, dickens, where the function is 55% of cycles) shows the hot instructions are overwhelmingly `test $0x1, %reg` and `shr $1, %reg`: tagged-`Nat` box checks and untagging on the position arithmetic, spread flat across the loop body. The walk's positions, fuel, and best-length bookkeeping all flow through `Nat` (`Zip/Native/Deflate.lean:925-950`), so every add, compare, and array index pays the tag dance.

## Precedent in-tree

The insertion loop already made this exact move: `updateHashesFastU` (`Zip/Native/Deflate.lean:967`) runs on `USize` with a guarded equivalence back to the `Nat` reference (`updateHashesGuarded_eq`), and the `countMatch` compare loop did the same (`goU` + `goU_eq`, the Wave 7 P1a pattern). The chain walk is the last big matcher loop still on tagged `Nat`.

## Plan

1. Port `chainWalkPacked` to a `chainWalkPackedU` on `USize` positions (window arithmetic, `prev` indexing via `cand &&& 0x7FFF`, packed `bestPos * 512 + bestLen` result), following the FastU + guarded-fallback pattern: one runtime addressability check outside the loop, unreachable `Nat` fallback inside.
2. Prove `chainWalkPackedU_eq` against the `Nat` walk (the `goU_eq` / `updateHashesGuarded_eq` playbook: `USize` round-trip lemmas from `Zip/Native/Deflate.lean:365-379`).
3. Callers (`lz77ChainIterP` main loops) switch to the U form.

Depends on #2735 (niceLen early-out) only for sequencing: both edit the same function, and the parameter change is trivial while this is a rewrite; land the knob first so the port carries it.

## Validation

Output-preserving (proven equal), so ratio is unchanged by construction; `perf-pr-graphs` before/after for the speed axis, plus a before/after `perf annotate` sanity check that the tag-test density is gone.

🤖 Prepared with Claude Code

kim-em:
This PR ports the hot DEFLATE hash-chain walk `chainWalkPacked` to a `USize`-native twin `chainWalkPackedU`: the fuel counter, the best-length/best-position accumulator, and the `scan_end` prefilter's index arithmetic run on unboxed `USize` instead of tagged `Nat`, so the walk pays no per-step tag test/untag on its scalar arithmetic and reads the prefilter bytes through `ByteArray.uget` with no bounds branch. The chain link `cand` itself stays on the `Nat` `prev` ring — the `chainWinSize` docstring records that a de-boxed ring measured as a net conversion tax — so only the accumulators move to `USize`. `chainWalkPackedU` is proven equal to `chainWalkPacked` (`chainWalkPackedU_eq`), and the guarded wrapper `chainWalkGuardedPackedU` is proven unconditionally equal to `chainWalkGuardedPacked` (`chainWalkGuardedPackedU_eq`) behind one runtime addressability + accumulator-faithfulness check that falls back to the `Nat` walk, so the packed matchers `lz77ChainIterP`/`lz77ChainLazyIterP` switch to it with output unchanged by construction.

Closes #2741.

🤖 Prepared with Claude Code

### DIFF
--- a/Zip/Native/Deflate.lean
+++ b/Zip/Native/Deflate.lean
@@ -1026,6 +1026,102 @@ decreasing_by all_goals omega
     let p := lz77Chain.chainWalk data prev windowSize pos maxLen niceLen hpm cand fuel bestLen bestPos
     p.2 * 512 + p.1
 
+/-- USize-native de-boxed twin of `chainWalkPacked` (Wave 7 P1b): the per-position
+    chain-walk bookkeeping — the fuel counter, the best-length/best-position
+    accumulator, and the `scan_end` prefilter's index arithmetic — runs on unboxed
+    `USize` rather than tagged `Nat`, so the hot loop pays no per-step tag test /
+    untag on its scalar arithmetic. The chain link `cand` itself stays `Nat`: it is
+    read straight from the `Nat` `prev` ring and handed to `countMatch` and the mask
+    index with no conversion, matching the measured `UInt32`-ring tax the
+    `chainWinSize` docstring records — only the accumulators move to `USize`. One
+    runtime `data.size < USize.size` addressability witness (`hsz`, established by
+    the guarded wrapper exactly as `countMatch` does for `goUW`) makes every
+    `USize` add faithful, and the result is repacked into the `Nat`
+    `bestPos * 512 + bestLen` at the leaves via `.toNat`, so no `USize` multiply can
+    overflow. The invariant `USize` copies of `pos`/`maxLen`/`min niceLen maxLen`
+    (`posU`/`maxLenU`/`cutoffU`, computed once by the wrapper) carry their `toNat`
+    identities so the loop never reconverts them. Proven equal to `chainWalkPacked`
+    (`chainWalkPackedU_eq`). -/
+def chainWalkPackedU (data : ByteArray) (prev : Array Nat)
+    (windowSize pos maxLen niceLen : Nat) (hpm : pos + maxLen ≤ data.size)
+    (hps : min chainWinSize data.size ≤ prev.size) (hsz : data.size < USize.size)
+    (posU maxLenU cutoffU : USize) (hposU : posU.toNat = pos) (hmaxU : maxLenU.toNat = maxLen)
+    (hcutU : cutoffU.toNat = min niceLen maxLen)
+    (cand : Nat) (fuelU bestLenU bestPosU : USize) : Nat :=
+  if fuelU = 0 then bestPosU.toNat * 512 + bestLenU.toNat
+  else if hc : cand < pos ∧ pos - cand ≤ windowSize then
+    have hUS : USize.size = 2 ^ System.Platform.numBits := rfl
+    have hcand : cand + maxLen ≤ data.size := by omega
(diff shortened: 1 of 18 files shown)


kim-em:
## Before/after `compress_mbps` (machine `chungus`) — clean sandwich

**Verdict: a real +2.4% speedup on both corpora, output byte-identical.** Earlier
in this PR I posted a routine two-sweep comparison that showed a *Silesia regression*;
that was cross-sweep drift plus machine contention, not the code. I re-measured
properly with a **self-controlled interleaved sandwich** (BEFORE→AFTER→BEFORE→…,
each AFTER bracketed by its neighbouring BEFOREs, all pinned to one idle core, 5
after-samples), and it flips the picture.

Two of the eleven Canterbury sweeps got hit by whole-sweep contention bursts
(one BEFORE at −8%, one AFTER at −26%, uniform across every file in the sweep);
the **outlier-robust median** over the 5 after-samples rejects them. The clean
sweeps are tight and consistent.

### Native compress_mbps speedup by level (median A / median B, geomean over files)

| level | canterbury | silesia |
|---|---|---|
| L1 | +0.9% | +0.8% |
| L2 | +1.5% | +0.9% |
| L3 | +1.9% | +1.3% |
| L4 | +2.4% | +2.3% |
| L5 | +4.0% | +3.9% |
| L6 | +2.9% | +2.5% |
| L7 | +3.8% | +3.6% |
| L8 | +4.1% | +3.8% |
| **L1–8 geomean** | **+2.7%** | **+2.4%** |

The gain rises with level (deeper chain walks = more of the hot loop de-boxed),
consistent across corpora. Silesia's five after-sweeps were +2.1/+2.4/+2.3/+1.8/+2.6%
— no outliers, so that +2.4% is dead certain. The same-binary noise floor
(consecutive BEFORE split-halves) is under ±0.5% at every level, well below the signal.

Output is **provably identical** — `chainWalkPackedU_eq` / `chainWalkGuardedPackedU_eq`
in Lean, and the plotter confirms `max |Δratio| = 0.000000`. This is pure speed.

### Positioning (native before/after over the other-language curves)

silesia

canterbury

AFTER (red) sits just above BEFORE (grey dashed) at the affected mid-levels on both
corpora; L9/L10 (untouched optimal parse) coincide. Native holds its mid-cluster
position (above OCaml, below the C/Rust/SIMD engines) — an internal de-boxing, not
a competitive-position move.

**Methodology note:** the sandwich ran under a ~17% ambient-contention offset vs the
committed dashboard baseline (same binary, verified — the box was busier midday than
when the baseline was recorded). That offset cancels in the A/B ratio, so the *delta*
is trustworthy, but the sandwich's absolute MB/s are not. The dashboard refresh in
this PR therefore keeps the baseline's absolute scale and applies the measured
per-level ratio to it, rather than splicing the offset absolutes (which would fake a
uniform regression on the untouched levels).

PR #2756, opened 2026-07-03
perf: wide-store BitWriter flush via pushUInt64LE (#2631)

### LINKED ISSUE #2631: perf(explore): BitWriter wide-store — local usetUInt64LE + pre-sized buffer (unblocked by #2707)

kim-em:
This PR replace the BitWriter's per-byte flush loop with a single wide store. A new project-local `@[extern]` primitive `ByteArray.pushUInt64LE` appends the low `k ≤ 8` bytes of the 64-bit accumulator as one unconditional 8-byte store into capacity slack plus a size bump (`c/bytearray_wide_ffi.c`), with the `k` iterated byte pushes as its trusted reference body. `writeBits`/`writeHuffCode` now flush through `flushBytesWide`, proven equal to `flushBytes` (`flushBytesWide_eq`), so `writeBits_def`/`writeHuffCode_def` and every downstream spec proof are unchanged. The issue's literal step-1 writer `ByteArray.usetUInt64LE` (8-byte LE store at an offset, model = eight proven `set` writes) is also added with conformance tests; it serves the pre-sized/deferred writer designs and the kernel bench, while the production flush uses only `pushUInt64LE`.

Step 0, the issue's required kernel check (`lake exe wide-store-bench`, four kernels over the same synthetic L1-like field trace, outputs verified byte-identical):

| variant | ns/field | vs A |
|---|---|---|
| A production push/byte | 7.9 | — |
| B grow-by-push + wide append (`pushUInt64LE`) | 5.5 | **1.44×** |
| C pre-sized buffer + `usetUInt64LE` cursor | 5.6 | 1.42× |
| D pre-sized + deferred 64-bit accumulator | 3.2 | 2.46× |

B ≈ C confirms the branch-sample finding in [issue #2739 (perf(explore): attribute and cut the RC/allocator tax)](https://github.com/kim-em/lean-zip/issues/2739) that push-doubling reallocs are already fully amortized, so the issue's step-2 restructure (pre-sized output buffer + proven size bound, "the hard part") is skipped entirely — the append-form wide store captures the whole per-call win with zero proof churn. Variant D (defer flushing in a 64-bit accumulator, one store per ~4 fields) measures a further 1.7× at the kernel and is left as a follow-up; it changes the BitWriter representation and its `wf` invariant.

A large confound surfaced on the way: `c/bytearray_wide_ffi.c` was compiled with no `-O` flag at all, so every `lean.h` static-inline helper (`lean_sarray_cptr`, `lean_is_exclusive`, ...) was an outlined -O0 call. The lakefile target now passes `-O2 -DNDEBUG`, which also speeds the already-landed readers of [PR #2707 (perf: local @[extern] word-sized ByteArray readers)](https://github.com/kim-em/lean-zip/pull/2707).

Measured end-to-end on Silesia (`l1-attrib`, pinned core, median-of-5, tokens held constant, baseline d7cd9688): fixed-only emit +17–28% and full L1 base candidate +12–21%, on all 12 files. Before/after dashboard graphs to follow as a PR comment per the perf-PR workflow.

Codex second opinion: no high/medium findings; its three low-severity suggestions (underflow-proof capacity check, `-DNDEBUG` rationale comment, deterministic fast-path test seed) are addressed in the last commit.

Closes [issue #2631 (perf(explore): BitWriter wide-store — local usetUInt64LE + pre-sized buffer)](https://github.com/kim-em/lean-zip/issues/2631).

🤖 Prepared with Claude Code


kim-em:
Before/after `compress_mbps` (machine `Linux chungus x86_64`, same machine as the committed dashboard; BEFORE built at merge-base d7cd9688 via a one-off worktree build because the committed `latest.json` predated [PR #2755 (perf: USize-native chain walk)](https://github.com/kim-em/lean-zip/pull/2755); both sides rebuilt from a purged `.lake` after the binary-health control caught a poisoned-config decode depression — final decode control after/before geomean **1.001** (n=207), `max |Δratio| = 0.000000` output-neutral):

compress pareto silesia

compress pareto canterbury

Silesia (geomean over 12 files):

|---|---|---|---|---|
| 1 | 0.361 | 73.5 | 77.8 | **1.06×** |
| 2 | 0.349 | 57.0 | 59.5 | 1.04× |
| 3 | 0.344 | 52.5 | 54.6 | 1.04× |
| 4 | 0.333 | 40.3 | 41.6 | 1.03× |
| 5 | 0.331 | 32.3 | 33.3 | 1.03× |
| 6 | 0.325 | 23.8 | 24.2 | 1.01× |
| 7 | 0.323 | 20.5 | 20.7 | 1.01× |
| 8 | 0.323 | 13.9 | 14.0 | 1.00× |
| 9 | 0.316 | 3.6 | 3.6 | 1.00× |
| 10 | 0.309 | 2.1 | 2.1 | 1.01× |

Canterbury (median-of-5): L1 1.03×, L2–L3 1.02×, L4–L6 1.01×, L7–L10 1.00×.

Visual read: on Silesia the AFTER curve lifts above BEFORE across the whole fast half (L1–L5) and converges at L6+ where matching dominates; the native fast end now sits clearly above JS fflate and closes on Go compress/flate at equal ratio, with zlib/Zig/miniz/libdeflate still ahead. Ratio axis unchanged everywhere (the change is output-neutral by construction, and the plotter confirms Δratio = 0).

🤖 Prepared with Claude Code

PR #2757, opened 2026-07-04
perf(compress): drop the over-fit fixed-cadence split candidate at level 8

kim-em:
This PR removes the second (fixed-cadence) shared-window split candidate that level 8 alone arbitrated on top of the base and observation-divergence candidates, because it is over-fit to a single benchmark input. A per-file diff shows the candidate — a full extra frequency + Huffman sizing pass over the token stream cut at a fixed 8192-token cadence, computed on every input — changes the emitted bytes on exactly one Silesia file (a statistically uniform record catalog whose slow global drift the divergence heuristic never cuts), improving it 0.8% while costing a 1.1-2.1x level-8 compress slowdown on all twelve. Deciding a whole level's cost, on every input, from one file's 0.8% is the definition of over-fitting; the candidate and the corpus-specific comment justifying it both go.

With the tax gone, level 8 runs 1.44x faster on Silesia (1.37x Canterbury) at byte-identical output everywhere except that one file (+0.065% Silesia-total ratio), and moves from ~14% inside the achievable mixing frontier (dominated by a level-7/level-9 mix) to a genuine frontier vertex. Levels 6 and 7 are untouched — they never ran the second candidate.

A drift-detection hybrid that tried to keep the one-file win selectively was prototyped and measured to fail: neither the observation classes nor a finer byte-value head/tail histogram separate that input's benefit from the others, because the benefit is local structure visible only to the sizing pass itself, not to any cheap two-window statistic.

The roundtrip proof simplifies rather than churns: level 8's dispatch is now the same single obs-split emit as levels 6-7, so the `8 ≤ level` case split in `inflate_deflateRaw` / `deflateRaw_pad` / `deflateRaw_goR_pad` collapses to the shared `hwithObs` case.

Addresses the residual level-7/level-8 half of https://github.com/kim-em/lean-zip/issues/2698 (perf(explore): spread the per-level ladder — L4-6 and L7-8 collapse on the compress Pareto): the level-8 point is no longer wasted interior. Full separation of levels 7 and 8 remains bounded by the optimal-parse tier (they share a ratio ceiling), tracked by the DP-parser work.

🤖 Prepared with Claude Code

### DIFF
--- a/Zip/Native/DeflateDynamic.lean
+++ b/Zip/Native/DeflateDynamic.lean
@@ -1747,47 +1747,30 @@ def deflateRaw (data : ByteArray) (level : UInt8 := 6) : ByteArray :=
     else
       -- Levels 6–8 (and level 9/10 above the memory gate): base vs cross-block
       -- shared-window split at the observation-divergence boundaries (#2737),
-      -- **size-arbitrated** (#2753). Every candidate is *prepared* — sized to its
-      -- flushed byte count with its per-block trees captured (`deflateRawBasePPrep`
-      -- for the base, `deflateDynamicBlocksSharedAtSizedP` for each cut list) — and
-      -- only the winner is emitted, reusing the trees the sizing pass already
-      -- built, instead of emitting every candidate and discarding all but the
-      -- smallest via `pickSmaller`. `emitSmallerBy` selects by the same ⌈bits/8⌉
-      -- byte counts the emitted blocks flush (pinned by the `SizeHelpers`
-      -- conformance tests), so the winner and its bytes are identical to the
-      -- retired two/three-emit dispatch. The tree capture is what makes this a net
+      -- **size-arbitrated** (#2753). Both candidates are *prepared* — sized to
+      -- their flushed byte count with per-block trees captured
+      -- (`deflateRawBasePPrep` for the base, `deflateDynamicBlocksSharedAtSizedP`
+      -- for the cut list) — and only the winner is emitted, reusing the trees the
+      -- sizing pass already built, instead of emitting both and discarding the
+      -- larger via `pickSmaller`. The winner and its bytes are identical to the
+      -- retired emit-both dispatch. The tree capture is what makes this a net
       -- win: sizing a dynamic block otherwise costs a full frequency + Huffman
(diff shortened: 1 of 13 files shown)


kim-em:
Before/after `compress_mbps`, native levels 1-10 (machine `Linux chungus x86_64`; BEFORE = merge-base `2c9505f2` built in the same worktree; both native-only sweeps, same session on a quiet box).

**Silesia** (weighted heavier):

silesia compress pareto

**Canterbury**:

canterbury compress pareto

Geomean native compress MB/s by level:

| lvl | ratio | Silesia before | Silesia after | × | Canterbury before | Canterbury after | × |
|----:|------:|---:|---:|---:|---:|---:|---:|
| 1 | .361/.328 | 77.2 | 77.6 | 1.00 | 44.2 | 43.2 | 0.98 |
| 6 | .325/.298 | 24.2 | 24.3 | 1.00 | 18.5 | 18.2 | 0.98 |
| 7 | .323/.297 | 20.7 | 20.8 | 1.00 | 16.7 | 16.4 | 0.99 |
| **8** | **.323/.296** | **14.0** | **19.3** | **1.38** | **12.3** | **15.6** | **1.27** |
| 9 | .316/.289 | 3.6 | 3.6 | 1.00 | 3.6 | 3.5 | 0.98 |
| 10 | .309/.285 | 2.1 | 2.1 | 1.00 | 2.1 | 2.1 | 0.98 |

Reading it: every untouched level reads 0.98-1.01x (Canterbury carries a uniform ~2% machine-drift offset on all levels, so its drift-corrected L8 is ~1.30x), confirming the change is isolated to L8. **L8 is +38% (Silesia) / +27% (Canterbury)**, and `max |Δratio| = 0.0062` occurs at **exactly one row — silesia/sao L8** — every other row is byte-identical. On the frontier, L8's grey BEFORE dot sits below the red AFTER curve at ratio ≈.323: it lifted from ~14% inside the mixing frontier onto a vertex adjacent to L7.

Caveat: L8 now sits close to L7 (they share a ratio ceiling); L8 is a legitimate vertex (better ratio, not dominated) rather than wasted interior, but full L7/L8 separation is optimal-parse-bound (#2638). Dashboard refreshed in-PR (commit `e2ea3221`).

PR #2758, opened 2026-07-05
perf: batch BitWriter flushes to cut per-field FFI stores

kim-em:
This PR widens the DEFLATE `BitWriter` to hold up to 31 pending bits in a `UInt64` accumulator and flush whole bytes only when 32 or more accumulate, instead of storing every completed byte immediately. The old writer made one `pushUInt64LE` FFI store per bit-field, which — now that the per-bit loop is gone (#2631) — is the dominant emit cost on a literal-heavy stream. Batching drains 4-7 bytes in a single store, cutting the FFI store count roughly 4x. On Silesia at level 1 this measures +26-38% on the dynamic emit walk and +5-10% on the full pipeline (geomean ~8%), byte-identical output: every roundtrip and cross-engine differential-fuzz test (zlib, libdeflate, miniz_oxide) still passes. Because the emit walk is shared, every level's dynamic and fixed emit gets the same speedup.

The change is transparent to the emit functions and their proofs: they call `writeBits`/`writeHuffCode` through the unchanged `toBits`/`bitLength`/`flush` contracts. Only the `BitWriter` internals move — a new `flushBatched` reference reused by `writeBits_def`/`writeHuffCode_def`, the `wf` invariant relaxed from `bitCount < 8` to `< 32`, `flushBatched_spec` layered over the existing `flushAcc_spec`, and `flush` now draining the multiple pending bytes before the final partial. The `toBits` logical bit sequence is unchanged, so block sizing (`bitLength`) and the flushed bytes depend only on the total bits written, not on when they are drained. No theorem is weakened.

Follow-up to #2731 (shipped the `deflate_fast` L1 matcher) and the #2726 attribution — this is the half #2731 left. Re-running the tokens-held-constant attribution at the shipped cd4/ic2 L1 (below) showed the emit token-walk, not `tokenFreqsP` or dynamic sizing, dominates the emit path — and, crucially, that the walk was **FFI-bound rather than compute-bound**: de-boxing the code tables (`Array (UInt16 × UInt8)` → `Array UInt32`) or folding frequencies into the matcher each moved little (candidate B measured -2%), while batching the writer moved a lot. L1 stays matcher-bound overall (~60% of the pipeline; a zero-cost emit would cap it near ~99 MB/s), so this closes the reachable encode/emit gap rather than the full distance to the pure-language pack.

Step-1 re-attribution (level 1, tokens held constant, Silesia, MB/s vs uncompressed):

[...]

PR #2759, opened 2026-07-05
perf: dual hash4+hash3 match finder for the L9/L10 candidate cache

### LINKED ISSUE #2742: perf: dual hash4+hash3 match finder (chain on 4-byte hash, singleton 3-byte table; unblocks hash4 at L9/L10)
depends-on: #2741

## Problem

All chain levels hash **3 bytes** into a 16-bit table (`lz77Greedy.hash3`, `Zip/Native/Deflate.lean:457-484`). A 3-byte hash over chains has two costs: chains mix candidates that can never extend past 3 bytes into every walk (more `countMatch` calls per useful candidate), and the alternative of just switching to a 4-byte hash is known to be blocked at the top tiers: the note at `Zip/Native/DeflateParse.lean:224-231` records that hash4 cost +7.4% ratio on x-ray and +3.5% on ooffice at L9, because length-3 matches stop being findable. So the fast levels are stuck sharing a hash that is wrong for both ends.

## What libdeflate does (hc_matchfinder.h:112-333)

Two tables, split by role:
- **hash4** (16-bit table) owns the chains: `next_tab` links positions sharing a 4-byte hash; the depth-bounded walk runs here, so every candidate can extend to length >= 4.
- **hash3** (15-bit table) is a **singleton**: one most-recent position per bucket, no chain. It is consulted once, only while `best_len < 4`, to keep length-3 matches reachable.
- Both hashes come from one `load_u32` (already available here as `ugetUInt32LE`) with multiplicative hashing.

Result: chain quality of hash4, length-3 coverage of hash3, one extra table probe per position.

## Plan

1. Add the singleton hash3 table alongside a hash4-keyed `prev` chain in the chain matcher state; walk chains on hash4, probe hash3 once when the best so far is < 4.
2. Wire the same scheme into the L9/L10 cache build (`chainWalkAllFast` / `buildCache`), unblocking hash4 there without the recorded x-ray/ooffice ratio regression.
3. Optional follow-on knob once this lands: libdeflate's adaptive minimum match length (skip length-3 matches entirely when the literal alphabet is small, `deflate_compress.c:2295-2327`).

All heuristic-state changes: the `lzMatch` contracts and roundtrip proofs are untouched; the packed/guarded equivalence twins get re-threaded.

Depends on #2741 (USize-native chain walk): this restructures the same state and loops; build it on the untagged version to avoid doing the port twice.

## Validation

`perf-pr-graphs` before/after. Expect speed gains at L4-L8 (fewer dead candidates per walk) and a speed gain at L9/L10; geomean ratio must hold within noise, with x-ray and ooffice checked individually against the recorded hash4 regression.

🤖 Prepared with Claude Code

kim-em:
This PR adopts libdeflate's dual hash4+hash3 match-finder scheme (`hc_matchfinder.h`) for the L9/L10 candidate-cache build, and records the measured verdict against extending it to the L1-L8 chain matchers (closes #2742). The cache build (`buildCache`) had kept a 3-byte chain hash precisely because plain hash4 chains cost ratio there (x-ray +7.4%, ooffice +3.5% at L9 — the note at the old `hash3opt`), at the price of chains polluted with candidates that can never extend past 3 bytes. The chains now key on the 4-byte hash (`lz77Greedy.hash3`), and a new 15-bit **singleton** table (`h3tab`, most recent position per 3-byte-hash bucket; `hash3Single`/`hash3Probe` in `Deflate.lean`) restores length-3 coverage: the probe verifies its candidate (window guard + three byte compares) and, on a hit, records it as cache slot 0 and seeds the chain walk's best length, so the ratio-maximising DP still sees every length-3 point while the depth-bounded walk spends its budget only on candidates that can reach length ≥ 4. `insertChainRange` (the skip-ahead interior insert) maintains both tables. Measured on Canterbury+Silesia (chungus, pinned): L9 +7.1% compress speed at −0.47% geomean ratio, L10 +11% at −0.08%, with the recorded hash4 regression held off (x-ray L9 +0.06%, ooffice L9 −0.22%; worst x-ray L10 +0.39%).

The probe result is packed as `cand3 * 512 + 3` (the `chainWalkGuardedPacked` encoding) rather than a `Nat × Nat`: `Prod` projections of the probe application inside a well-founded loop body send the kernel into deep recursion (minimal repro confirmed; same blowup class the `buildCache` docstring records for literal constants). Everything is heuristic state — the cache never enters a proof, and the only proof churn is the size-preservation lemmas following `buildCache`'s widened state.

The issue's plan item 1 (probe the singleton in the L1-L8 chain matchers too) was implemented, measured, and **rejected**: even with a zlib-style `TOO_FAR = 4096` cap on the seed distance (without which the extra far length-3 matches *worsen* the fast-level ratio 0.2-0.5%), the per-position singleton touch cost −15-20% matcher speed at L1-L5 and −6% at L6-L8 for at best −0.17% geomean ratio — inside the adjacent levels' mixing curve at every level (e.g. seeded L4 lands at 28.3 MB/s where the baseline L4/L5 mix reaches 33.1 MB/s at equal ratio), so the point is dominated and the matchers stay probe-free. The full implementation (seeded walks, `TOO_FAR`, re-threaded lockstep proofs) is preserved in this branch's history (d4e86ca9) if the cost structure ever changes. A second opinion (Codex) on that fuller implementation found no correctness blockers; its remaining notes (the pre-existing `slots`-vs-`optCacheSlots` asymmetry in `buildCache`, proven-bounds `h3tab` access as a future micro-optimisation) are recorded here.

Closes #2742.

🤖 Prepared with Claude Code

### DIFF
--- a/Zip/Native/Deflate.lean
+++ b/Zip/Native/Deflate.lean
@@ -834,12 +834,42 @@ table contents never enter a correctness proof. -/
 /-- 15-bit multiplicative hash of the 3 bytes at `pos` (libdeflate's `hash3`):
     the singleton-table bucket index. Always `< 32768` (a `UInt32 >>> 17` has
     15 bits), so the `guardedSet`/`headProbeGuarded` bounds checks on the
-    32768-entry table never fail. -/
+    32768-entry table never fail. The word load mirrors `lz77Greedy.hash3`'s
+    exactly (same wide-load/fallback split), so after inlining the two loads
+    at one position are common-subexpression candidates — libdeflate computes
+    both hashes from one `load_u32` for the same reason; the `&&& 0xFFFFFF`
+    mask makes the bucket independent of which load path ran (byte 3 never
+    enters), which the singleton's hit rate relies on. -/
 @[inline] def hash3Single (data : ByteArray) (pos : Nat) (h : pos + 2 < data.size) : Nat :=
-  let a := (data[pos]'(by omega)).toUInt32
-  let b := (data[pos + 1]'(by omega)).toUInt32
-  let c := (data[pos + 2]'(by omega)).toUInt32
-  (((a ||| (b <<< 8) ||| (c <<< 16)) * 2654435761) >>> 17).toNat
+  let word :=
+    if h4 : pos + 4 ≤ data.size then
+      if hsz : data.size.toUSize.toNat = data.size then
+        ByteArray.ugetUInt32LE data pos.toUSize (by
+          have hds : data.size < USize.size := by
+            rw [← hsz]; exact USize.toNat_lt_two_pow_numBits _
+          rw [toUSize_toNat_of_lt (show pos < USize.size by omega)]; omega)
+      else
+        let a := (data[pos]'(by omega)).toUInt32
+        let b := (data[pos + 1]'(by omega)).toUInt32
+        let c := (data[pos + 2]'(by omega)).toUInt32
+        let d := (data[pos + 3]'(by omega)).toUInt32
+        a ||| (b <<< 8) ||| (c <<< 16) ||| (d <<< 24)
+    else
+      let a := (data[pos]'(by omega)).toUInt32
+      let b := (data[pos + 1]'(by omega)).toUInt32
+      let c := (data[pos + 2]'(by omega)).toUInt32
+      a ||| (b <<< 8) ||| (c <<< 16)
+  ((((word &&& 0xFFFFFF) * 2654435761)) >>> 17).toNat
+
+/-- Reject length-3 seeds farther than this (zlib `deflate.c` `TOO_FAR`): a
+    3-byte match at a large distance costs more bits (long distance code) than
+    the three literals it replaces, so the greedy/lazy matchers only accept
+    the hash3 candidate within this radius — measured on Canterbury+Silesia,
+    the unguarded singleton *worsens* the fast-level geomean ratio by
+    0.2–0.5%. The L9/L10 cache build deliberately does **not** apply this cap:
+    its DP prices every candidate with a real cost model, so far length-3
+    candidates are simply never chosen when they don't pay. -/
+def tooFar3 : Nat := 4096
 
 /-- Probe result of the hash3 singleton: the chain-walk seed, packed as the
     single small `Nat` `bestPos * 512 + bestLen` (the `chainWalkGuardedPacked`
@@ -975,7 +1005,7 @@ where
       let h3 := hash3Single data pos hlt
       let cand3 := headProbeGuarded h3tab h3
       let h3tab := guardedSet h3tab h3 pos
-      let seed := hash3Probe data windowSize pos cand3 hlt
+      let seed := hash3Probe data (min windowSize tooFar3) pos cand3 hlt
       let maxLen := min 258 (data.size - pos)
       have hmaxLenP : pos + maxLen ≤ data.size := by omega
       let m := chainWalk data prev windowSize pos maxLen niceLen hmaxLenP head maxChain (seed % 512) (seed / 512)
@@ -1304,7 +1334,7 @@ where
       let h3 := hash3Single data pos hlt
       let cand3 := headProbeGuarded h3tab h3
       let h3tab := guardedSet h3tab h3 pos
-      let seed := hash3Probe data windowSize pos cand3 hlt
+      let seed := hash3Probe data (min windowSize tooFar3) pos cand3 hlt
       let maxLen := min 258 (data.size - pos)
       have hmaxLenP : pos + maxLen ≤ data.size := by omega
       let r := chainWalkGuardedPacked data prev windowSize pos maxLen niceLen hmaxLenP head maxChain (seed % 512) (seed / 512)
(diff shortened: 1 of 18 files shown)


kim-em:
Before/after `compress_mbps` Pareto (native levels 1–10 vs the committed other-language curves; machine `Linux chungus x86_64`, pinned via `bench/pin_core.sh`; BEFORE = master's committed `bench/results/latest.json`, cross-validated by a same-session merge-base run within +0.6% on every level's compress geomean):

silesia pareto

canterbury pareto

Per-level compress geomean (ratio → speed):

| lvl | Silesia before | after | speedup | Canterbury before | after | speedup |
|-----|---------------|-------|---------|-------------------|-------|---------|
| 1–8 | — | — | 0.97–1.00x | — | — | 0.99–1.00x |
| 9 | 0.3144 @ 3.6 MB/s | 0.3129 @ 3.9 | **1.09x** | 0.2880 @ 3.5 | 0.2871 @ 3.9 | **1.10x** |
| 10 | 0.3085 @ 2.1 | 0.3083 @ 2.4 | **1.15x** | 0.2853 @ 2.1 | 0.2852 @ 2.3 | **1.10x** |

Geomean over both corpora: **L9 +9.3% speed at −0.49% ratio, L10 +12.6% at −0.08%** — both axes improve, so no mixing-curve test is needed. L1–L8 are byte-identical to master (`max |Δratio| = 0` on those levels; their 0.97–1.00x spread is cross-session noise, decompress control ≈ 1.00x throughout). The plotter's overall `max |Δratio| = 0.0031` is silesia/reymont L9 moving **down** (better), the intended parse change.

The recorded hash4 hazard (#2620 note: x-ray +7.4%, ooffice +3.5% at L9) held off: x-ray L9 **+0.06%**, ooffice L9 **−0.22%**; worst per-file anywhere is x-ray L10 +0.39%. Largest per-file speed wins: mr L10 +35%, dickens L10 +28%, reymont L10 +25%.

Caveat: Silesia is single-pass (±noise on any one file's speed); Canterbury is median-of-5 and agrees on the L9/L10 magnitude, and the deltas are an order above the untouched-level noise band.

kim-em:
Follow-up on the overlay's apparent L1–L3 Silesia dip (−3%): a **self-controlled sandwich measurement** (master and branch `bench-report` binaries built in the *same* worktree; per rep BEFORE→AFTER→BEFORE pinned to one fixed core, 5 reps, levels 1–8, no concurrent benchmarks) shows the untouched levels are **unchanged**:

| corpus | AFTER/BEFORE geomean (95% CI) | same-binary noise floor (M2/M1) |
|--------|-------------------------------|--------------------------------|
| Canterbury L1–L8 overall | **+0.17% ± 0.09%** | −0.35% ± 0.13% |
| Silesia L1–L8 overall | **+0.14% ± 0.07%** | −0.15% ± 0.08% |

Per-level deltas range +0.04%…+0.38% on both corpora — no level regresses, and the magnitude sits at the measurement's own noise floor. (L1–L8 output is bit-identical to master — `max |Δratio| = 0` — so any real speed effect could only have been code-layout; the sandwich bounds that at ≈0.) The overlay graph's low-level separation was cross-session scatter on single-pass Silesia against the dashboard's recorded session, not a code effect. The L9/L10 result (+9.3%/+12.6% at −0.49%/−0.08% ratio) stands as the PR's effect.

kim-em:
**Updated Pareto graphs** — the earlier overlay's BEFORE was the dashboard's recorded session, whose single-pass-Silesia scatter drew a phantom L1–L3 separation. These use same-session data on both sides at every level: **L1–L8 from the sandwich aggregates** (10 BEFORE runs vs 5 AFTER runs, both binaries built in this worktree, fixed core 47) and **L9/L10 from the pinned full runs**. Per-level table now reads 1.00x at L1–L8 on both corpora; the only separation is the two crown points.

silesia corrected pareto

canterbury corrected pareto

| lvl | Silesia before → after | Canterbury before → after |
|-----|------------------------|---------------------------|
| 1–8 | 1.00x at every level | 1.00x at every level |
| 9 | 3.6 → 3.9 MB/s (**1.09x**, ratio −0.49%) | 3.5 → 3.9 (**1.10x**, ratio −0.31%) |
| 10 | 2.1 → 2.4 (**1.15x**, ratio −0.08%) | 2.1 → 2.3 (**1.10x**, ratio −0.04%) |

(`max |Δratio|` overall = 0.0031, silesia/reymont L9 moving down/better; L1–L8 Δratio ≡ 0.)

kim-em:
**Regenerated after rebasing onto #2758** (BitWriter flush batching — it moved every level's emit speed and had not refreshed the dashboard). All numbers re-measured as a **same-session pair**: master (28d1b30b) and this branch built in the same worktree, measured back-to-back pinned to one fixed core, full levels 1–10, both corpora. The in-PR dashboard refresh now records the post-#2758 native rows, healing that staleness too.

silesia rebased pareto

canterbury rebased pareto

| lvl | Silesia | Canterbury |
|-----|---------|-----------|
| 1–8 | 1.00–1.01x, Δratio ≡ 0 | 1.00x, Δratio ≡ 0 |
| 9 | 3.6 → 4.0 MB/s (**1.10x**, ratio −0.49%) | **1.10x** (ratio −0.31%) |
| 10 | 2.1 → 2.5 (**1.17x**, ratio −0.08%) | **1.10x** (ratio −0.04%) |

Overall geomean: **L9 +9.9%, L10 +14.2%** compress speed at better-or-equal ratio; x-ray L9 +0.06% / ooffice L9 −0.22% (hazard held). L1–L8 bit-identical output and speed within ±0.8% same-session noise. Tests pass on the rebased branch.

Issue #2536, opened 2026-06-11
feat: prove computeCodeLengths produces complete codes (kraftSum = 2^maxBits) (Wave 2b)


kim-em:
**Proof-quality ratchet (Wave 2b) — characterizing property for the fixed length-limiter. Proof-only: touches Zip/Spec/ (+ at most lemma placement); no behavior change.**

### Current state

D-20 replaced the broken flat fallback in `computeCodeLengths`' length-limiter with the zlib `gen_bitlen`-style redistribution (`cappedBlCount`/`repairBl`/`expandBl`/`limitedPairs` in `Zip/Spec/HuffmanEncode.lean`). The Kraft *validity* half (`computeCodeLengths_valid`, ≤) is proven; **completeness** — that the produced code is exact, `kraftSum lens = 2^maxBits` for ≥ 2 distinct used symbols — is currently pinned only by tests, with `fixKraftList` as a defensive backstop that would silently mask a regression.

### Deliverables

1. Theorem (name suggestion `computeCodeLengths_complete`): for inputs with ≥ 2 nonzero-frequency symbols, the lengths returned by `computeCodeLengths freqs n maxBits` satisfy `kraftSum = 2^maxBits` exactly (state via the existing kraft-sum machinery in `Zip/Spec/HuffmanKraft.lean`).
2. Corollary: the `fixKraftList` backstop is the identity on `computeCodeLengths` output (its branch is proven-dead), stated as a simp-able equation if convenient.
3. If the proof reveals the property genuinely fails on an edge case (e.g. 1-symbol alphabets), document the exact boundary in the theorem statement rather than weakening silently — the RFC 1951 1-symbol case is the known exception.

### Context

This is the gold-standard characterizing property per the project's spec taxonomy: it permanently rules out the zlib-rejected-output bug class that D-20 fixed, independent of tests. It also de-risks the performance pivot — encoder hot-path changes can't silently regress into incomplete codes. Background/parallel work: no contention with the perf PRs (different files).

### Verification

`lake build && lake exe test` green; `grep -rc sorry Zip/` = 0; no `native_decide`; theorem statement uses ≥ 2-symbol hypothesis explicitly.

kim-em:
Session: `unknown` Branch: `master`

kim-em:
Claimed by session `cbf44e43-d1b9-44f7-beb1-c7adcad213ac` on branch `agent/cbf44e43`

kim-em:
[...]

PR #2769, opened 2026-07-05
perf: native array-based Huffman code-length build

### LINKED ISSUE #2761: perf: native array-based Huffman code-length build (spec List build is ~25-30% of L6 mozilla since block splitting)
## Evidence (perf, `compress-pareto`, chungus, 2026-07-05, master 52a0d6d9)

Profiling L6 on **mozilla** (51 MB, heterogeneous, many split blocks) instead of dickens shows the matcher is not the bound there. The hot cluster is Huffman code-length construction:

| symbol | L6 mozilla | L6 dickens |
|---|---|---|
| `chainWalkPackedU` | 18.6% | 41.3% |
| `List.setTR` + `List.lengthTR` | 14.2% | not visible |
| `Huffman.Spec.insertByWeight` | 4.8% | not visible |
| `lean_dec_ref_cold` + `mi_malloc/free` | 16.1% | ~2% |
| `lean_array_push` | 3.4% | not visible |

The chain: each split block calls `dynamicCodeLengths` (`Zip/Native/DeflateFreqs.lean:260`), which calls `Huffman.Spec.computeCodeLengths`, which builds the tree with `buildHuffmanTree` merging via `insertByWeight`, an O(n) insertion into a sorted `List BuildTree` (`Zip/Spec/HuffmanEncode.lean:46-69`), and returns `List Nat` lengths that downstream walks with `List.set`/`List.length` (the `setTR`/`lengthTR` symbols; most of the RC and allocator share is this cluster's churn). Conservatively ~25-30% of L6 mozilla runtime.

This was invisible until #2745: one tree build per file is nothing; one per split block per sizing candidate on a 51 MB file is hundreds of List-based builds. It now also taxes the split sizing pass (#2753) and the L9/L10 per-region refits, which call the same function.

## Plan (generational refinement, the project's standard pattern)

1. Native twin `computeCodeLengthsN` on arrays: sort the (freq, symbol) pairs once, then the classic two-queue O(n) merge (leaves queue + internal-nodes queue, both FIFO after the sort), depths extracted by array walk, and the existing maxBits (15) cap/repair logic ported to the array form. `Array Nat` (or `Array UInt32`) lengths end-to-end; `dynamicCodeLengths` consumers move off `List Nat`.
2. Prove `computeCodeLengthsN = computeCodeLengths` (equality transfer; the spec version and its W2b completeness theorems, #2536 family, stay untouched as the reference).
3. Measure in situ before finishing the proof (`sorry` the equality first, per the measure-before-prove rule).

Expected: +20-30% compress at L5-L8 on heterogeneous members (mozilla, samba, osdb, x-ray), smaller on text; byte-identical output. Also speeds L9/L10 refits and the sizing pass.

## Validation

Byte-identical output across both corpora (conformance of lengths against the spec version), then `perf-pr-graphs` for the speed axis; re-profile mozilla L6 to confirm the cluster is gone.

🤖 Prepared with Claude Code

kim-em:
This PR adds `Huffman.Spec.computeCodeLengthsN`, an `Array Nat` twin of the spec `computeCodeLengths`, and routes the hot `dynamicCodeLengths` path through it, so the per-split-block Huffman code-length construction no longer pays the `O(n²)` `List.set`/`List.length` churn that `assignLengths` incurred (the `setTR`/`lengthTR` cluster the profiler put at ~14% of L6 mozilla, plus its allocator/reference-count traffic). The array twin makes symbol→length assignment `O(n)` via `Array.setIfInBounds`, and returns the lengths as an `Array` directly. The spec `List` pipeline (`assignLengths`, `fixKraftList`, `computeCodeLengths`, `limitedPairs`, and the #2536 completeness theorems) stays untouched as the reference; `computeCodeLengthsN_toList` proves the array pipeline read back as a list is exactly the spec `List`, so every downstream property (length, `ValidLengths`, completeness, nonzero) transfers verbatim and the compressed output is byte-identical. The one perturbed spec proof (`emitDynBlock_spec`) is re-proved through the same correspondence lemma; no theorem statement changed and there is no `sorry`, `native_decide`, or `implemented_by` trust gap — the fast path is on the verified route. The swap also covers the L9/L10 per-region refits and the split sizing pass, since both go through `sizedTrees` → `dynamicCodeLengths`.

Measured before/after on Silesia + Canterbury (`compress_mbps`, native, same machine `chungus2`), `max |Δratio| = 0.000000` (byte-identical):

| member | L6 before → after | speedup |
|---|---|---|
| mozilla | 17.33 → 22.98 | **1.33x** |
| x-ray | 9.98 → 13.90 | **1.39x** |
| ooffice | 17.80 → 21.42 | 1.20x |
| samba | 29.45 → 33.50 | 1.14x |
| mr | 20.92 → 23.33 | 1.12x |
| dickens (text) | 22.87 → 23.41 | 1.02x |

The win lands exactly where the issue predicted: heterogeneous, heavily-split members at L5–L8 (+14–42%), flat on text, and L1–L5 sit at ~1.00x (they split less, so the code-length build is not hot there — a clean control that the AFTER binary is not uniformly faster). L9/L10 refits also improve (mozilla 1.06x/1.04x, sao 1.19x/1.11x). A per-block microbench (`bench/ZipHuffBench.lean`) over Canterbury measures the isolated build 3.4–4.3x faster with byte-identical lengths on every file.

Closes https://github.com/kim-em/lean-zip/issues/2761

🤖 Prepared with Claude Code

### DIFF
--- a/Zip/Native/DeflateFreqs.lean
+++ b/Zip/Native/DeflateFreqs.lean
@@ -1,6 +1,7 @@
 import Zip.Native.Deflate
 import Zip.Spec.EmitTokensCorrect
 import Zip.Spec.HuffmanEncode
+import Zip.Native.HuffmanEncode
 
 /-!
   Token-stream frequency analysis and dynamic-Huffman code-length selection,
@@ -258,8 +259,8 @@ def freqsToPairs (freqs : Array Nat) : List (Nat × Nat) :=
     block emitter (`deflateDynamicBlock`) and the size-then-emit dispatch so both
     select identical trees from a single computation. -/
 def dynamicCodeLengths (litFreqs distFreqs : Array Nat) : List Nat × List Nat :=
-  let litLens := Huffman.Spec.computeCodeLengths (freqsToPairs litFreqs) 286 15
-  let distLens := Huffman.Spec.computeCodeLengths (freqsToPairs distFreqs) 30 15
+  let litLens := (Huffman.Spec.computeCodeLengthsN (freqsToPairs litFreqs) 286 15).toList
+  let distLens := (Huffman.Spec.computeCodeLengthsN (freqsToPairs distFreqs) 30 15).toList
   let distLens := if distLens.all (· == 0) then distLens.set 0 1 else distLens
   (litLens, distLens)
 
@@ -268,10 +269,13 @@ def dynamicCodeLengths (litFreqs distFreqs : Array Nat) : List Nat × List Nat :
 theorem dynamicCodeLengths_length (litFreqs distFreqs : Array Nat) :
     (dynamicCodeLengths litFreqs distFreqs).1.length = 286 ∧
     (dynamicCodeLengths litFreqs distFreqs).2.length = 30 := by
-  refine ⟨Huffman.Spec.computeCodeLengths_length _ 286 15, ?_⟩
-  show List.length (if _ then _ else _) = 30
-  split
-  · rw [List.length_set]; exact Huffman.Spec.computeCodeLengths_length _ 30 15
-  · exact Huffman.Spec.computeCodeLengths_length _ 30 15
+  refine ⟨?_, ?_⟩
+  · show ((Huffman.Spec.computeCodeLengthsN _ 286 15).toList).length = 286
+    rw [Array.length_toList]; exact Huffman.Spec.computeCodeLengthsN_size _ 286 15
+  · show List.length (if _ then _ else _) = 30
+    split
+    · rw [List.length_set, Array.length_toList]
+      exact Huffman.Spec.computeCodeLengthsN_size _ 30 15
+    · rw [Array.length_toList]; exact Huffman.Spec.computeCodeLengthsN_size _ 30 15
 
 end Zip.Native.Deflate
--- a/Zip/Native/HuffmanEncode.lean
+++ b/Zip/Native/HuffmanEncode.lean
@@ -1,0 +1,118 @@
+import Zip.Spec.HuffmanEncode
+
+/-!
+# Native array-based Huffman code-length construction
+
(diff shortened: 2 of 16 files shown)


kim-em:
Before/after `compress_mbps`, native only, **both measured on `chungus2`** (same worktree, sequential). Output is byte-identical: `max |Δratio| = 0.000000`.

**Caveat:** the committed dashboard (`bench/results/latest.json`) and the reference-language curves overlaid below were recorded on `chungus`, not `chungus2`. The native before/after **delta** is valid (same machine, same worktree); the native-vs-other-language *positioning* is context only. I did **not** splice these `chungus2` native rows into the committed dashboard, since that would make `latest.json` internally machine-inconsistent — the dashboard refresh should be run on `chungus`.

silesia pareto

canterbury pareto

Silesia geomean by level (before → after MB/s):

|----|-------|--------|-------|---------|
| 5 | 0.331 | 34.5 | 34.5 | 1.00x |
| 6 | 0.325 | 25.0 | 27.9 | **1.11x** |
| 7 | 0.323 | 21.4 | 23.5 | 1.10x |
| 8 | 0.323 | 19.9 | 21.7 | 1.09x |
| 9 | 0.314 | 4.0 | 4.3 | 1.05x |
| 10 | 0.309 | 2.5 | 2.6 | 1.04x |

Per-member at L6 (the split-heavy members are the winners): mozilla **1.33x**, x-ray **1.39x**, ooffice 1.20x, samba 1.14x, mr 1.12x; text (dickens/webster/reymont) ~1.00–1.02x. L1–L5 sit at ~0.99–1.00x across all members — a control that the AFTER binary is not uniformly faster; the gain appears precisely where block-splitting makes the per-block code-length build hot.

PR #2770, opened 2026-07-05
bench: migrate canonical benchmark machine to chungus2

kim-em:
This PR moves the Track D dashboard from `chungus` to `chungus2` and re-records the full matrix — native plus all eight comparators plus the decode-density experiment — on the new machine, preserving the last `chungus` snapshot verbatim under `bench/results/archive/` rather than deleting it.

The move is sound. A new `bench/cross_machine.py` compares the two machines cell-by-cell (full write-up in [`bench/results/cross-machine-chungus-to-chungus2.md`](https://github.com/kim-em/lean-zip/blob/bench-migrate-chungus2/bench/results/cross-machine-chungus-to-chungus2.md)): the version-stable comparators — `native`, `libdeflate`, `miniz_oxide`, `zig`, `zlib`, `ocaml`, 1334 cells — agree to within 0.3% on throughput (geomean 1.0004× compress, 0.9967× decompress), and compression ratio is byte-identical across both machines and both commits (`max |Δratio| = 0.000000` for all eight compressors, 1748 cells). Historical numbers therefore stay directly comparable to fresh chungus2 numbers.

Be fair to `go` and `js`: each now runs on its best current release. `go` builds with `nix-shell -p go` (go 1.26.3, the newest — its `compress/flate` is much faster than the older Go the old snapshot used, hence +77%/+133%), and the JS comparator moves from the older default `nodejs` (v24) to `nodejs_latest` (v26, whose V8 is ~+22% on JS compress) in both `comparators/build_all.sh` and `run.sh`. The `go`/`js` rows thus track the compiler version across the move, not the machine; `js` additionally carries large V8 run-to-run variance, so its rows are a max-of-three clean sequential passes and are indicative only. Pinning `go`/`js` to a concrete version for strict reproducibility (the way `zig_0_14` is) is left as a possible follow-up — it would trade "best" for "frozen".

🤖 Prepared with Claude Code

### DIFF
--- a/bench/README.md
+++ b/bench/README.md
@@ -20,6 +20,19 @@ the external-language comparators (see below), then [`plot.py`](plot.py) (writes
 the SVGs). Ratios are deterministic; throughput is a **median-of-5 snapshot of
 the machine recorded in the JSON `meta`** — commit the JSON and SVGs together.
 
+> **Benchmark machine: chungus2 (since 2026-07-05).** The canonical machine moved
+> from `chungus` to `chungus2`. The two are indistinguishable on throughput —
+> within ~0.3% across every version-stable comparator, and byte-identical on ratio
+> — so historical numbers stay directly comparable. `go` and `js` each run on
+> their **newest** release (`nix-shell -p go` → go 1.26.3; `nodejs_latest` → node
+> v26, ~+22% JS compress over the old default v24), so those two curves track the
+> compiler version rather than the machine; js in particular carries large V8
+> run-to-run variance and is indicative only. Full analysis:
+> [`results/cross-machine-chungus-to-chungus2.md`](results/cross-machine-chungus-to-chungus2.md);
+> the last `chungus` snapshot is preserved under
+> [`results/archive/`](results/archive/) and the comparison is reproducible via
+> [`cross_machine.py`](cross_machine.py).
+
 ## Compressors compared
 
 The honest comparison group for a pure-Lean codec is other **language-native**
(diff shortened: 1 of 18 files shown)


kim-em:
New canonical dashboard on **chungus2** (full matrix, all 8 comparators, go 1.26.3 + node v26):

silesia pareto chungus2

Cross-machine consistency (chungus → chungus2), throughput geomean chungus2/chungus:

| compressor | compress× | decompress× | version across the move |
|---|---|---|---|
| **version-stable (native, libdeflate, miniz, zig, zlib, ocaml)** | **1.0004×** | **0.9967×** | stable → machines identical to 0.3% |
| go | 1.775× | 2.330× | older Go → go 1.26.3 (newest, faster compiler) |
| js | 0.930× | 0.722× | older node → v26 (newest); high V8 variance, indicative |

Ratio is byte-identical on both machines for every compressor (`max |Δratio| = 0.000000`, 1748 cells). Old chungus snapshot preserved under `bench/results/archive/`.

kim-em:
**Before/after machine overlay** — chungus (before, dashed/hollow) vs chungus2 (after, solid), per compressor, on Silesia. Ratio is deterministic so each codec's before/after dots share x; the vertical gap between its dashed and solid curve *is* the machine's effect on that codec.

Compress:

machine overlay compress

Decompress:

machine overlay decompress

The reading is clean: for every version-stable codec — native, zlib, miniz_oxide, libdeflate, zig, ocaml — the dashed and solid curves lie on top of each other (machines identical to ~0.3%). Only **go** separates upward (solid above dashed: the newer go 1.26.3 compiler, +77% compress / +133% decompress) and **js** wobbles (V8 cross-session variance; indicative only).

Issue #2762, opened 2026-07-05
perf: make the split pass cheap (fuse heuristic passes, scalar counters, derive whole-file freqs from per-block)


kim-em:
## Evidence (perf, `compress-pareto`, chungus, 2026-07-05, master 52a0d6d9)

The observation-divergence split heuristic costs 6.1% (L6 dickens) / 6.6% (L6 mozilla) of compress, and the frequency passes add more. libdeflate's equivalent bookkeeping is close to free. Three separable wastes in `chooseSplitsHeuristicP` (`Zip/Native/DeflateDynamic.lean:1145`) and around it:

1. A **separate full pass** over the token array computes `totalBytes` before the main loop (lines 1149-1151); the main loop then walks the same array again.
2. `splitTokenBytesP t` is computed **twice per token** (`blockBytes` and `doneBytes` both add it).
3. The 10 observation counters live in a boxed `Array Nat` updated with `set!`/`getD` per token; scalars in the loop state would compile to register arithmetic.

Related: `tokenFreqsP` (4.7% L6 dickens, 3.0% mozilla) runs whole-file for the base candidate while the sizing pass computes per-block frequencies over the same tokens. Whole-file frequencies are the sum of per-block frequencies: one per-block pass plus a cheap 316-entry summation replaces the second full walk.

## Plan

1. Fuse the `totalBytes` pre-pass away: compute it in the matcher main loop (it already touches every token) or restructure the cut loop to not need the remaining-bytes lookahead until the first candidate cut (running suffix = total minus prefix).
2. One `splitTokenBytesP` per token.
3. Counters as scalar loop state (10 locals or a small flat tuple threaded through a well-founded loop; note the `Id.run do` forIn compiles to the opaque `Std.Legacy.Range.forIn'` loop visible in the profile).
4. Derive whole-file freqs by summing per-block freqs so the token array is walked once for frequency purposes.

All heuristic-side (the emitter clamps arbitrary cuts) except item 4, which needs the summation equality (`tokenFreqsP toks = sum of per-block tokenFreqsP`), a small lemma. Conformance pin against the boxed heuristic already exists in `ZipTest/PackedTokens.lean`.

Expected: ~4-6% compress at L4-L8, byte-identical output.

🤖 Prepared with Claude Code

kim-em:
[...]

PR #2771, opened 2026-07-06
perf: fuse split-heuristic passes and scalarize its counters (#2762)

kim-em:
This PR makes the observation-divergence split heuristic cheap by fusing its passes and moving its counters into scalar loop state. The heuristic cost 6.1% (L6 dickens) / 6.6% (L6 mozilla) of compress (chungus, 2026-07-05); a local before/after over the Canterbury corpus (plrabn12, kennedy.xls, lcet10, ptt5, alice29) measures `chooseSplitsHeuristicP` at 1.33 ms/call before and 0.61 ms/call after — ~2.2x on the function itself — with byte-identical cut lists on every file.

Three separable wastes in `chooseSplitsHeuristicP` go away:

1. The separate full pass computing `totalBytes` before the main loop is gone. The whole-stream output byte count is now a parameter: the dispatch and tests pass `data.size`, since the LZ77 token stream decodes to exactly `data` (`Σ splitTokenBytesP = data.size`). The main walk tracks a running byte suffix `remaining` (start `totalBytes`, minus each token) instead of a `doneBytes` prefix, so the min-block floor is gated without a second pass.
2. `splitTokenBytesP` is computed once per token instead of twice.
3. The ten observation counters move from a boxed `Array Nat` (updated with `set!`/`getD` per token) into scalar loop state threaded through a tail-recursive well-founded loop, so the hot accumulators compile to register arithmetic. The full ten-counter arrays are materialized only at a divergence check (every `checkTokens` tokens once the floors clear), where the shared boxed `splitEndBlockCheck` reads them.

The heuristic stays proof-free (`emitSharedBlocksAtP` clamps arbitrary cuts, so any cut list is correctness-safe); the `totalBytes` signature change threads `data.size` through the dispatch, the roundtrip/padding specs, and the conformance, size-helper, and mid-sweep call sites. Output is byte-identical: the `PackedTokens` conformance pin (`chooseSplitsHeuristicP` against the boxed `chooseSplitsHeuristic` over the `unpackTok` view) and the `SizeHelpers` byte-identity checks pass unchanged.

Item 4 from the issue (derive whole-file frequencies by summing the per-block frequencies the split-sizing pass already computes) is deferred to a follow-up, https://github.com/kim-em/lean-zip/issues/2772: unlike the proof-free heuristic, the base candidate feeds the verified roundtrip/padding specs, so it needs a proven EOB-corrected partition-sum equality (`tokenFreqsP` pre-counts symbol 256, so summing k blocks over-counts it by k-1) threaded through the clamped sizing recursion — real proof infrastructure that does not belong bundled with this low-risk heuristic rewrite.

Addresses https://github.com/kim-em/lean-zip/issues/2762 (items 1-3). Item 4 tracked in https://github.com/kim-em/lean-zip/issues/2772.

🤖 Prepared with Claude Code

### DIFF
--- a/Zip/Native/DeflateDynamic.lean
+++ b/Zip/Native/DeflateDynamic.lean
@@ -1133,6 +1133,71 @@ boxed reference via `deflateDynamicBlocksSharedAtP_eq`
   if w &&& ((1 : UInt32) <<< 31) = 0 then 1
   else ((w >>> 16) &&& 0x7FFF).toNat
 
+/-- The per-token core of `chooseSplitsHeuristicP`, threaded as a tail-recursive
+    loop over explicit scalar state so the hot accumulators compile to register
+    arithmetic instead of `Array Nat` `set!`/`getD` (#2762). The ten observation
+    counters live in the `n0..n9` (recent window) and `o0..o9` (block-so-far)
+    locals; `newTot`/`oldTot`/`blockBytes` are the window/block totals, and
+    `remaining` is the running byte suffix `totalBytes − doneBytes` (decremented
+    per token) that gates cuts against the min-block floor without a second
+    pass. The full ten-counter arrays are materialized only at a divergence
+    check (every `checkTokens` tokens, once the floors clear), where the shared
+    boxed `splitEndBlockCheck` reads them. -/
+def chooseSplitsHeuristicP.go (toks : Array UInt32)
+    (minBlockBytes softMaxBlockBytes checkTokens : Nat) (i : Nat)
+    (o0 o1 o2 o3 o4 o5 o6 o7 o8 o9 oldTot : Nat)
+    (n0 n1 n2 n3 n4 n5 n6 n7 n8 n9 newTot : Nat)
+    (blockBytes remaining : Nat) (cuts : Array Nat) : Array Nat :=
+  if h : i < toks.size then
+    let t := toks[i]
+    let c := splitTokenClassP t
+    let tb := splitTokenBytesP t
+    let (n0, n1, n2, n3, n4, n5, n6, n7, n8, n9) :=
+      match c with
+      | 0 => (n0 + 1, n1, n2, n3, n4, n5, n6, n7, n8, n9)
+      | 1 => (n0, n1 + 1, n2, n3, n4, n5, n6, n7, n8, n9)
+      | 2 => (n0, n1, n2 + 1, n3, n4, n5, n6, n7, n8, n9)
+      | 3 => (n0, n1, n2, n3 + 1, n4, n5, n6, n7, n8, n9)
+      | 4 => (n0, n1, n2, n3, n4 + 1, n5, n6, n7, n8, n9)
+      | 5 => (n0, n1, n2, n3, n4, n5 + 1, n6, n7, n8, n9)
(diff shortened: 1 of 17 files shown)


kim-em:
## Before/after performance — `compress_mbps`

Machine `Linux chungus2 x86_64` (matches the committed dashboard). Output is **byte-identical**: `max |Δratio| = 0.000000`. This PR only touches the level-6–8 split-heuristic path — levels 1–5 (single-block / greedy) and 9–10 (optimal parse; every corpus file is under the 64 MiB `optimalMaxSize` gate, so the split path is never taken there) are unchanged and act as controls. The full L1–10 sweep is plotted so the L9-fast / L10 crown stays on the frontier.

silesia

canterbury

### Self-controlled sandwich (the verdict for this small, localized delta)

BEFORE→AFTER→BEFORE per rep, 4 reps, pinned to one idle core; both binaries built in the *same* worktree (no cross-worktree phantom). The AFTER effect is bracketed by two BEFOREs so slow drift cancels; a same-binary run gives the noise floor for free, and levels 1–5 are unchanged-code controls (their residual is subtracted as position bias).

| level | speedup (bias-corr) | 95% CI (raw) | same-binary floor |
|------:|--------------------:|:------------:|------------------:|
| 1–5 (controls) | ≈0% | straddles 0 | ±0.4% |
| **6** | **+6.2%** | [+5.2%, +7.6%] | −0.05% |
| **7** | **+5.6%** | [+4.6%, +7.1%] | −0.02% |
| **8** | **+5.6%** | [+4.5%, +7.0%] | −0.13% |

Per corpus at the target levels (bias-corrected):

| corpus | L6 | L7 | L8 |
|:--|--:|--:|--:|
| Silesia | +3.8% | +3.3% | +3.3% |
| Canterbury | +8.9% | +8.3% | +8.1% |

The control levels read ≈0 with CIs straddling zero and the same-binary floor is within ±0.4%, so the L6–8 signal clears the noise — it is the code, not the machine. Isolated, `chooseSplitsHeuristicP` itself goes **1.33 → 0.61 ms/call (~2.2×)** on the Canterbury token streams, with byte-identical cut lists.

Overlay geomean by level (L1–8 sandwich geomean of 4 reps; L9/L10 single full-sweep pass — unchanged path, shown for the crown):

```
              silesia                    canterbury
lvl  ratio  before  after  spd     ratio  before  after  spd
  1  0.361   84.0   84.5  1.01x    0.328   61.3   61.4  1.00x
  5  0.331   34.3   34.5  1.00x    0.301   29.8   29.8  1.00x
  6  0.325   27.8   28.9  1.04x    0.298   24.5   26.7  1.09x
  7  0.323   23.5   24.3  1.03x    0.297   21.5   23.4  1.08x
  8  0.323   21.7   22.4  1.03x    0.296   20.2   21.9  1.08x
  9  0.314    4.3    4.2  0.99x    0.288    4.5    4.5  0.99x   (L9-fast, unchanged)
 10  0.309    2.6    2.6  1.00x    0.285    2.6    2.6  0.99x   (crown, unchanged)
```

Caveat: L9/L10 are single-pass (their ~1.00x is the unchanged optimal-parse path, well inside noise); the sandwich table is the median-of-reps verdict for the L6–8 delta. Native sits in the competitive middle of the field at unchanged ratio, with the L6–8 segment nudged faster and the crown reaching the best ratio (near the zopfli star).

PR #2774, opened 2026-07-06
perf: half-depth lazy lookahead probe (libdeflate)

### LINKED ISSUE #2763: perf: half-depth lazy lookahead probe (libdeflate) or graded good-match depth reduction (zlib)
## Evidence

The lazy matcher's lookahead probe at `pos+1` runs at **full `maxChain` depth** (`Zip/Native/Deflate.lean:1372`). Since the `goodMatch` skip is disabled at L5+ (`goodMatch = 259`, the #2737 mid-sweep verdict), every matched position pays **two full-depth chain walks**. libdeflate probes its first lookahead at **half** depth and its second at quarter depth (`deflate_compress.c:2712-2767`); zlib's alternative mechanism quarters the remaining chain budget once the current match is already ≥ `good_match` (`chain_length >>= 2`) instead of skipping the probe outright.

The chain walk is 41% of L6 dickens (perf, 2026-07-05, master 52a0d6d9), so the lookahead's share is material.

## Plan

1. Thread a probe depth into the lazy lookahead: `maxChain / 2` first (libdeflate's value), sweep `/2` vs `/4`.
2. Optionally, as the graded replacement for the binary `goodMatch` skip that lost ratio in the mid-sweep: reduce (not zero) the probe depth when the first match is already long.
3. Heuristic knob only: parameter threading through the packed/guarded twins, no contract changes.

Expected: L5-L7 up several percent at near-zero ratio cost (the probe still runs, just shallower; libdeflate holds better ratios than us with half-depth probes).

## Validation

`perf-pr-graphs` before/after on both corpora; per-level ratio deltas must stay in noise.

🤖 Prepared with Claude Code

kim-em:
This PR threads a `lazyDepth` knob into the lazy hash-chain matcher's `pos+1` lookahead so the second chain walk runs at reduced depth instead of the full `maxChain`, then wires the production levels to half depth (`chainDepth / 2`, libdeflate's value). Since the `goodMatch` gate is disabled at L5+ (`goodMatch = 259`, the #2737 mid-sweep verdict), every matched position otherwise pays two full-depth chain walks — and the chain walk is 41% of L6 dickens, so the lookahead's share is material. libdeflate probes its first lookahead at half depth and still holds better ratios than us, so the deferral quality a half-depth probe gives up is slack: the probe still runs, just shallower, so the second walk's cycles roughly halve while the ratio cost stays in the noise.

The parameter threads identically through the three lazy twins (`lz77ChainLazy`, `lz77ChainLazyIter`, `lz77ChainLazyIterP`), defaulting to `maxChain` so existing callers are a no-op; only `DeflateDynamic.lzMatch`/`lzMatchP` opt into `lazyDepth level = chainDepth level / 2`. Depth is a pure heuristic — `chainWalk_spec` holds for any fuel — so the equivalence (`lz77ChainLazyIter_eq_lz77ChainLazy`, `lz77ChainLazyIterP_eq`) and validity/encodability/resolves contracts are re-proven generic over `lazyDepth` and re-established at emission exactly as before. The conformance test now also checks the three twins agree at a non-default depth (32), and the `mid-sweep` timing tool models the shipped half-depth probe. No new `sorry`; full build and test suite pass.

Closes https://github.com/kim-em/lean-zip/issues/2763

🤖 Prepared with Claude Code

### DIFF
--- a/Zip/Native/Deflate.lean
+++ b/Zip/Native/Deflate.lean
@@ -1328,6 +1328,13 @@ where
     Huffman-distribution effect no *local* deferral rule captures (that needs the
     cost-model parse of #2526 part 2).
 
+    The lookahead walk at `pos+1` runs at `lazyDepth` chain steps, a separate knob
+    from the `pos` walk's `maxChain`. It defaults to `maxChain` (full-depth probe,
+    the original behavior); callers pass a reduced value (libdeflate probes its
+    first lookahead at *half* depth) to buy back the second full-depth walk each
+    matched position otherwise pays. Depth is a pure heuristic — `chainWalk_spec`
+    holds for any fuel — so any `lazyDepth` keeps the encoder contracts.
+
     Match-finding stays opaque to correctness — `chainWalk_spec` holds for *any*
     chain state and *any* deferral predicate, so validity is re-established at
     emission exactly as for `lz77Chain`. `pos+1` is intentionally *not* inserted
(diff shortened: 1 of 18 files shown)


kim-em:
## Before/after `compress_mbps` — half-depth lazy lookahead probe

Machine `Linux chungus2 x86_64` (the canonical dashboard machine). Native levels 1–10, Canterbury median-of-5, Silesia single-pass. BEFORE = merge-base `41821049` binary (built in a dedicated worktree, run from the main worktree so it sees Silesia); AFTER = this branch. **The L1–L3 greedy control levels — untouched by this patch — read 1.00–1.01x on both corpora, confirming no cross-worktree build artifact**, so the deltas below are the real effect.

**Silesia** (the corpus that matters most):

silesia pareto

**Canterbury:**

canterbury pareto

Native compress geomean by level (× = after/before):

| lvl | Canterbury | Silesia |
|----:|:----------:|:-------:|
| 1 (control) | 1.01x | 1.00x |
| 2 (control) | 1.00x | 1.00x |
| 3 (control) | 1.01x | 0.99x |
| 4 | 1.03x | 1.03x |
| 5 | 1.06x | 1.06x |
| 6 | 1.05x | 1.05x |
| 7 | 1.06x | 1.05x |
| 8 | 1.07x | 1.05x |
| 9 | 1.04x | 1.01x |
| 10 (DP crown) | 1.02x | 1.01x |

The lazy tier (L4–L8, the target) is a consistent **+3% to +7%** on both corpora — the second chain walk's cycles roughly halve, as intended. The change is a within-frontier lift (the L4–L8 segment moves up at held ratio), not a leapfrog of another engine; lean-zip native stays in the mid-pack behind the libdeflate C+SIMD ceiling.

**Ratio cost stays in the noise**: `max |Δratio| = 0.0026` over every (file, level), on one small file (canterbury/kennedy.xls L5); per-level geomean ratios are unchanged to 3 decimals. L9/L10 (optimal-parse tiers, which do not use the lazy chain probe) are at noise as expected.

Dashboard (`bench/results/latest.json` + `bench/graphs`) refreshed in-PR with the AFTER numbers.

PR #2775, opened 2026-07-06
perf: batch literal emission via writeHuffCode2 (up to 2 Huffman codes per writer call)

### LINKED ISSUE #2764: perf: batch literal emission (up to 4 Huffman codes per writer call)
## Evidence

After the batched flush (#2758), each literal still costs one out-of-line `writeHuffCode` call from the emit walk: `writeHuffCode` + `writeBits` are ~5.4% of L6 dickens and the walk itself (`emitTokensWithCodesP`) another 2.3% (perf, 2026-07-05). libdeflate writes **4 literals per bit-buffer flush group** (`deflate_compress.c:1968-1999`).

Literal codes are ≤ 15 bits, so 4 reversed codes fit in one 64-bit accumulator merge with room to spare (4 × 15 = 60 ≤ 64 pending-bit budget of the #2758 writer, which holds up to 31 pending bits between flushes; a 2-literal batch needs ≤ 30 and fits unconditionally).

## Plan

1. In `emitTokensWithCodesP` (and the fixed-emit twin), peek ahead: when the next 2 (or up to 4, depending on the pending-bit budget) tokens are literals, look up their codes, merge them into one shifted/OR-ed field, and hand the combined `(bits, len)` to a single writer call.
2. Prove the batched emit equals the sequential one via the existing `toBits` contracts (the logical bit sequence is unchanged; this is the same shape as `flushBatched_spec` layering in #2758).

Expected: a few percent at L1-L6 on literal-heavy streams (text corpora), byte-identical output. Modest but cheap, and it composes with the #2758 writer.

## Validation

Byte-identical roundtrip suite; `perf-pr-graphs` speed axis.

🤖 Prepared with Claude Code

kim-em:
This PR adds `BitWriter.writeHuffCode2`, which merges two reversed Huffman codes into one 64-bit accumulator field with a single flush check — the way libdeflate writes its flush groups — and uses it in the packed emit walks to batch consecutive literal tokens. When the first field would not trigger an intermediate flush (`bitCount + len1 < flushThreshold`, the common case right after a flush leaves few pending bits), both codes are shift/OR-ed into the accumulator and flushed once; otherwise it falls back to the two sequential `writeHuffCode` calls. `writeHuffCode2_eq` proves it byte-identical to that sequential pair as a `BitWriter` value (not merely at the logical-bit level), so nothing downstream changes.

The packed emitters `emitTokensWithCodesP` and the fixed twin `emitTokensP` now peek ahead: when two consecutive tokens are literals, they emit them with one `writeHuffCode2` and advance by two. The packed-vs-boxed equalities (`emitTokensWithCodesP_eq` / `emitTokensP_eq`) still hold — the batched branch rewrites through `writeHuffCode2_eq` and unfolds the boxed emitter one extra step — so the block-size models and roundtrip stay byte-identical. Closes https://github.com/kim-em/lean-zip/issues/2764.

The full test suite (including byte-identical roundtrip conformance) passes; no `sorry`, no `native_decide`. A second opinion from Codex confirmed the guard arithmetic and that `writeHuffCode2_eq` proves genuine byte-level `BitWriter` equality.

🤖 Prepared with Claude Code

### DIFF
--- a/Zip/Native/BitWriter.lean
+++ b/Zip/Native/BitWriter.lean
@@ -214,6 +214,96 @@ theorem writeHuffCode_def (bw : BitWriter) (code : UInt16) (len : UInt8) :
   · rw [flushBytesWide_eq, flushAcc_eq]
   · rfl
 
+/-- Write two Huffman codes in one merged accumulator operation — a single
+    shift/OR field and a single flush check for a pair of literals, the way
+    libdeflate writes its flush groups.
+
+    When the first code would not trigger an intermediate flush (`bitCount +
+    len1 < flushThreshold`, the common case right after a flush leaves few
+    pending bits), both reversed codes are merged into one 64-bit accumulator:
+    `code1` shifted above the pending bits, `code2` shifted above that, then the
+    combined field is flushed once. The guard's `else` branch (rare: `bitCount +
+    len1 ≥ 32`) falls back to the two sequential `writeHuffCode` calls.
+
+    `writeHuffCode2_eq` proves the result **byte-identical to the two sequential
+    `writeHuffCode` calls for every input** — no length precondition, because
+    the merge and the sequential second write use the same (modular) `UInt64`
+    shift/OR arithmetic: under the guard the first field never flushes, so
+    merging the second above it lands the exact same bits with the exact same
+    single flush the sequential pair would perform on its second call.
+
+    That the merged field does not lose bits off the top of the 64-bit
+    accumulator (i.e. that the output is well-formed DEFLATE, not merely equal to
+    the sequential writer) is a **caller invariant**: DEFLATE literal/length
+    Huffman codes have `len ≤ 15`, so with the guard `bitCount + len1 < 32` the
+    top bit position `bitCount + len1 + len2 < 47` sits well inside 64. This is
+    the same `len ≤ 15` invariant the emit specs already carry
+    (`ValidLengths … 15`); it is not needed for `writeHuffCode2_eq`. -/
+def writeHuffCode2 (bw : BitWriter) (code1 : UInt16) (len1 : UInt8)
+    (code2 : UInt16) (len2 : UInt8) : BitWriter :=
+  if bw.bitCount.toNat + len1.toNat < flushThreshold then
+    let rev1 : UInt64 := (reverse16 code1).toUInt64 >>> (16 - len1.toUInt64)
+    let rev2 : UInt64 := (reverse16 code2).toUInt64 >>> (16 - len2.toUInt64)
+    let acc : UInt64 :=
+      bw.bitBuf ||| (rev1 <<< bw.bitCount.toUInt64)
+        ||| (rev2 <<< (bw.bitCount.toUInt64 + len1.toUInt64))
+    let total := bw.bitCount.toNat + len1.toNat + len2.toNat
+    if total ≥ flushThreshold then
+      ⟨flushBytesWide bw.data acc (total / 8), dropBytes acc (total / 8), (total % 8).toUInt8⟩
+    else
+      ⟨bw.data, acc, total.toUInt8⟩
+  else
+    (bw.writeHuffCode code1 len1).writeHuffCode code2 len2
+
+/-- The explicit ctor-reuse flush form (`flushBytesWide`/`dropBytes`) the
+    production writers construct equals the `flushBatched` reference form —
+    exactly the bridge inside `writeBits_def`/`writeHuffCode_def`, factored out
+    so `writeHuffCode2_eq` can rewrite the batched fast path. -/
+private theorem flushExplicit_eq_flushBatched (data : ByteArray) (acc : UInt64) (total : Nat) :
+    (if total ≥ flushThreshold then
+      (⟨flushBytesWide data acc (total / 8), dropBytes acc (total / 8),
+        (total % 8).toUInt8⟩ : BitWriter)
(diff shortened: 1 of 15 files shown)


kim-em:
## Performance: byte-identical, throughput-neutral

Metric `compress_mbps`, machine `chungus2` (matches the dashboard). **Output is byte-identical** — the overlay's `max |Δratio| = 0.000000`, corroborating the `writeHuffCode2_eq` proof empirically. The honest verdict is that batching literal pairs is **throughput-neutral**: the compress effect sits below the run-to-run noise floor, and two independent self-controlled sandwiches disagree on its sign.

### Self-controlled sandwiches (BEFORE→AFTER→BEFORE, pinned core 90, levels 1–6, 5 reps)

The decompress path is untouched by this PR, so its measured "effect" is the sandwich's position/thermal bias; subtract it from the compress effect to isolate the code effect (`compress − control`):

| corpus | S1 compress | S1 control (decode) | S1 code effect | S2 compress | S2 control (decode) | S2 code effect |
|---|---|---|---|---|---|---|
| canterbury | +0.01% | +0.41% | **−0.40%** | +0.21% | −0.14% | **+0.35%** |
| silesia | +0.34% | +0.46% | **−0.12%** | +0.31% | −0.30% | **+0.61%** |

The control-corrected code effect flips sign between the two sessions (S1 negative, S2 positive), i.e. the true effect is smaller than the ±0.3–0.5% noise. **Neutral, no regression.** (S1 was measured against pre-#2774 master, S2 against current master; my diff is orthogonal to #2774, so both isolate this change.)

### Overlay (geomean, native before vs after)

| level | canterbury | silesia |
|---|---|---|
| L1 | 1.00x | 1.00x |
| L4 | 1.00x | 1.00x |
| L6 | 0.99x | 1.01x |
| L9 | 0.98x | 0.99x |
| L10 | 0.98x | 0.99x |

`max |Δratio| = 0.000000` (output-neutral, as proved).

silesia compress pareto
canterbury compress pareto

### Read

The change is proven byte-identical and does what the issue asked (one `writeHuffCode2` merged field per literal pair instead of two `writeHuffCode` calls), but it does **not** produce the "few percent" the issue predicted — the emit walk is too small a fraction of total compress time for halving its writer calls to register above noise. It adds no regression and the `writeHuffCode2` primitive + proof could compose with wider batching later.

@kim-em — since this is neutral rather than a win, your call on whether the proven-correct machinery is worth keeping. Not merging pending your review. Dashboard refreshed in-PR (commit d2dd9f1c).

kim-em:
## Conclusive: batching the writer calls does not speed up emit — closing

Chasing whether a real win was hidden by the matcher, I timed the **emit path in isolation** with `l1-attrib` (tokens matched once, held constant; only the token walk + BitWriter is timed). BEFORE = master, AFTER = this branch, alternating and pinned to one core, 5 reps each. The reps are tight and low-noise:

| file | emit BEFORE | emit AFTER | speedup |
|---|---|---|---|
| dickens | 247.5 MB/s | 246.2 MB/s | 0.995x (−0.5%) |
| mozilla | 308.4 MB/s | 308.3 MB/s | 1.000x (−0.0%) |
| webster | 309.9 MB/s | 307.2 MB/s | 0.991x (−0.9%) |

So even with the matcher removed entirely, batching two literals into one `writeHuffCode2` does **not** make emit faster — it is flat to marginally slower. This isn't a small-fraction wash; the emit stage itself doesn't move.

### Why (this is the useful part)

The premise was "each literal costs an out-of-line `writeHuffCode` + a flush check; merge them." But since the #2758 batched writer already **holds up to 31 pending bits and only drains bytes at the 32-bit threshold**:

1. Two typical literal codes (~8–9 bits each, ~17 bits total) usually do **not** cross the 32-bit flush threshold, so the second `writeHuffCode` wasn't flushing anyway — there is no store or flush to eliminate. The FFI store count is already minimized by #2758 regardless of how many writer *calls* happen.
2. What batching removes (one cell construction, one call boundary, one threshold branch) is offset by the new per-pair **lookahead** cost (bounds check + array load + tag test of token `i+1`).
3. Both paths still do 2 table loads, 2 `reverse16`, and 2 shift/ORs — the actual arithmetic per literal is unchanged.

Net: a wash, confirmed at the isolation level and corroborated by two self-controlled sandwiches (control-corrected effect flips sign between runs) and the overlay (`max |Δratio| = 0.000000`, ~1.00x every level).

### If we do want to push emit throughput later (higher-leverage than call-batching)

- **Precompute reversed codes at table-build time.** `writeHuffCode` runs `reverse16` (an 8-op swap network) on *every* literal; the canonical table is built once per block. Storing already-reversed, pre-shifted codes moves that work off the per-symbol hot path.
- **Flatten `Array (UInt16 × UInt8)`.** The code table is an array of boxed `Prod` pairs, so every literal lookup chases a pointer and unboxes two scalars. A flat `Array UInt32` (or parallel scalar arrays) would make it a single flat load.

Both remove real per-literal *work* rather than restructuring branches, so they have a chance where call-batching didn't — though still bounded by emit's share of total compress time.

Closing unmerged per the neutral result. The `writeHuffCode2` primitive + its `writeHuffCode2_eq` proof are correct and on record here if a future 4-wide/tight-run-loop emit design wants them.

kim-em:
Closing unmerged: measured neutral (byte-identical, no throughput change even with the emit path timed in isolation). See the analysis above. Branch left on record.

PR #2776, opened 2026-07-06
perf: cost-based lazy accept rule across the lazy tier (libdeflate)

### LINKED ISSUE #2765: perf(explore): lazy2 lookahead + cost-based accept rule at L7/L8 (the untouched ratio lever)
depends-on: #2763

## The untouched ratio lever at L7/L8

Two libdeflate mechanisms we have never tried, both pure matcher heuristics:

1. **Lazy2 (two-position lookahead).** libdeflate levels 8-9 probe both `pos+1` (half depth) and `pos+2` (quarter depth) before committing a match (`deflate_compress.c:2738-2767`). We have one-step lazy only, everywhere.
2. **Cost-based accept rule.** Our lazy accepts the lookahead match only if it is longer **and no farther** (`Zip/Native/Deflate.lean:1375`, `matchLen2 > matchLen ∧ dist2 ≤ dist1`), strictly more conservative than libdeflate's `4·Δlen + bsr32(old_offset) − bsr32(new_offset) > 2` (first step; `> 6` for the second), which accepts a longer-but-farther match whenever the length gain pays for the extra offset bits.

Either could close part of the remaining L7/L8 ratio distance to zlib/libdeflate (native L7 0.3233 vs zlib L7 0.3210) without new parser machinery.

## Plan (measurement-first)

1. Spike the accept-rule change alone (small diff, no new lookahead) and measure ratio on both corpora before any proof work.
2. Spike lazy2 at L7/L8 with the libdeflate depth ladder (full / half / quarter).
3. Keep whichever clears a real ratio win at acceptable speed; the matcher contracts are quantified over these heuristics, so proof churn is re-threading only.

Risk: our matcher differs from libdeflate's (no length-3 coverage at L1-L8, different insertion policy), so their thresholds may not transfer; the spike exists to find out cheaply. Sequencing: same loop as #2763 (half-depth probe), so land that first.

🤖 Prepared with Claude Code

kim-em:
This PR replaces the lazy chain matcher's conservative "longer and no-farther" lookahead accept rule (`len2 > len1 ∧ dist2 ≤ dist1`) with libdeflate's cost-based rule: it defers to the `pos+1` match when that match is strictly longer and its length gain pays for the extra offset bits, `4·Δlen + bsr(dist1) − bsr(dist2) > 2` (`bsr x = ⌊log2 x⌋`). The old rule rejected every longer-but-farther match outright; the cost rule admits a farther match once the length gain outweighs its ⌈log2⌉ offset-bit cost — exactly the deferrals a Huffman-distance model rewards.

Measured (`bench/ZipLazy2Sweep.lean`, added here) across Canterbury and a 105 MB Silesia subset: a −0.4 %…−0.9 % ratio win at every lazy level (4–9) on every file, text (xml −2.0 %, dickens −0.8 %) and binary (osdb −1.6 %, mozilla −0.3 %, sao −0.2 %) alike, with no regression anywhere, at neutral speed — one `Nat.log2` per deferral and no extra chain walk, dickens L7/L8 throughput unchanged within noise. A two-position (`pos+2`) lookahead was measured alongside and rejected: it returned ≤0.06 % extra ratio for a second chain walk per position and a new proof branch, so only the accept rule lands.

The accept rule is a pure heuristic: every emitted reference is re-verified at emission (`chainWalk_spec` holds for any chain state), so the matcher contracts are quantified over the accept predicate and the proof churn is re-threading only — one lemma (`lazyAcceptCost_lt`, accept ⇒ strictly longer, used to rule out the no-match case) plus two two-line insertions in `LZ77ChainLazyCorrect`. The `SizeHelpers` conformance reference is repointed from an unrelated matcher (`lz77LazyIter`) to the level-6 stream `deflateRaw` actually uses (`lzMatch data 6`), so it pins the size-then-emit dispatch rather than a coincidental agreement between two matchers.

Closes #2765.

🤖 Prepared with Claude Code

### DIFF
--- a/Zip/Native/Deflate.lean
+++ b/Zip/Native/Deflate.lean
@@ -1314,19 +1314,51 @@ where
   termination_by data.size - pos
   decreasing_by all_goals omega
 
+/-- libdeflate's cost-based lazy accept rule (`deflate_compress.c`, the first
+    lookahead step). The lazy matcher prefers the `pos+1` match `(len2, dist2)`
+    over the `pos` match `(len1, dist1)` when it is strictly longer **and** the
+    length gain pays for the extra offset bits: `4·Δlen + bsr(dist1) − bsr(dist2)
+    > 2`, where `bsr x = ⌊log2 x⌋` counts a distance's offset bits (written with
+    additions on both sides so the `Nat` subtractions never truncate).
+
+    This replaces the older conservative rule `len2 > len1 ∧ dist2 ≤ dist1`
+    (accept a longer match only when it is *no farther*), which rejected every
+    longer-but-farther match outright. The cost rule instead admits a farther
+    match once its length gain outweighs the ⌈log2⌉ offset-bit cost — exactly the
+    longer-but-farther deferrals a Huffman-distance model rewards. Measured
+    (#2765, `lazy2-sweep`) at −0.4 %…−0.9 % Silesia/Canterbury ratio across
+    *every* lazy level (4–9) and every file type — text (xml −2.0 %, dickens
+    −0.8 %) and binary (osdb −1.6 %, mozilla −0.3 %, sao −0.2 %) alike, with no
+    regression anywhere — at negligible cost (one `Nat.log2` per deferral, no
+    extra chain walk). A two-position (`pos+2`) lookahead was measured alongside
+    and rejected: it added ≤0.06 % for a second chain walk per position and a new
+    proof branch.
+
+    Purely a choice among valid matches — every emitted reference is re-verified
+    at emission (`chainWalk_spec` holds for any chain state and any deferral
+    predicate), so the rule never enters the correctness proof; the matcher
+    contracts hold for any accept predicate. -/
+@[inline] def lazyAcceptCost (len1 dist1 len2 dist2 : Nat) : Bool :=
+  decide (len1 < len2) && decide (4 * (len2 - len1) + dist1.log2 > 2 + dist2.log2)
+
+/-- Accepting the lookahead match implies it is strictly longer — the correctness
+    proofs use this to rule out the "no match at `pos+1`" case. -/
+theorem lazyAcceptCost_lt {len1 dist1 len2 dist2 : Nat}
+    (h : lazyAcceptCost len1 dist1 len2 dist2 = true) : len1 < len2 := by
+  simp only [lazyAcceptCost, Bool.and_eq_true, decide_eq_true_eq] at h
+  exact h.1
+
 /-- Lazy (one-byte-lookahead, zlib `deflate_slow`-style) variant of `lz77Chain`.
     At each position it finds the best chain match `M` at `pos`; before emitting it
     runs a *second* `chainWalk` at `pos+1` for `M'`. It defers — emits a literal at
-    `pos` and takes `M'` (advancing to `pos+1+len M'`) — only when `M'` is **both
-    longer and no farther** than `M` (`len M' > len M ∧ dist M' ≤ dist M`); otherwise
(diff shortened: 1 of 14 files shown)


kim-em:
## Before/after `compress_mbps` — cost-based lazy accept rule

Machine `Linux chungus2 x86_64` (matches the dashboard). BEFORE = merge-base `a27e7623` (rebuilt: the committed `latest.json` was recorded at `05a43b60`, before #2774's half-depth probe landed on `Zip/Native`, so it was stale — this PR also refreshes it). AFTER = `38a6ad15`. Native swept 1–10 (L10 = exact-DP crown); other-language curves reused from `latest.json`.

**Silesia** (12 files, geomean):

silesia pareto

**Canterbury** (11 files, geomean):

canterbury pareto

### Ratio + speed by level (bytes-weighted ratio; geomean MB/s)

Silesia:

| lvl | ratio before | ratio after | Δratio | MB/s before | MB/s after | speed |
|----:|-------------:|------------:|-------:|------------:|-----------:|------:|
| 1–3 (greedy, control) | — | — | **+0.000%** | — | — | **1.00x** |
| 4 | 0.3339 | 0.3326 | −0.381% | 44.7 | 45.6 | 1.02x |
| 5 | 0.3319 | 0.3302 | −0.519% | 36.5 | 37.8 | 1.03x |
| 6 | 0.3239 | 0.3224 | −0.484% | 30.3 | 31.0 | 1.02x |
| 7 | 0.3230 | **0.3211** | −0.575% | 25.4 | 26.3 | 1.03x |
| 8 | 0.3228 | **0.3209** | −0.596% | 23.6 | 24.5 | 1.04x |
| 9 | 0.3138 | 0.3137 | −0.030% | 4.3 | 4.3 | 1.01x |
| 10 (crown) | 0.3096 | 0.3096 | +0.000% | 2.6 | 2.6 | 1.01x |

Canterbury: same shape — L1–3 exactly +0.000%/1.00x (control), L4–8 −0.40%…−0.52% ratio at 1.01–1.02x.

### Read

- **Ratio win, no speed cost.** The lazy tier (L4–8) shifts left-and-up on the Pareto: −0.4%…−0.6% ratio at **1.01–1.04x faster** (the better parse emits fewer tokens, offsetting the one `Nat.log2` per deferral). Every file improves; nothing regresses (`max |Δratio| = 0.0059` at silesia/osdb L5, the right direction).
- **Clean control.** Greedy levels 1–3 and the L10 crown (neither uses the lazy matcher) read **exactly +0.000% ratio and 1.00x speed** — so the measurement carries no cross-worktree/`.lake` artifact, and the mid-band delta is the code, not the environment.
- **Hits the issue target.** #2765 cited native L7 0.3233 vs zlib L7 0.3210; this brings **native L7 to 0.3211** (≈ zlib L7) and L8 to 0.3209 (below it).

Cross-session caveat: AFTER is a fresh pinned run and BEFORE a rebuilt merge-base in the same session on a quiet machine; Silesia is single-pass. The exact-zero control levels and the same-direction win on all 20 lazy (corpus×level) cells make the mid-band delta solid despite single-pass Silesia noise.

kim-em:
### Zoomed view — ratio 0.31–0.33, 10–100 MB/s (Silesia)

The mid-band where native now lands in the Rust/zlib cluster. Native points labelled by level (grey = before, red = AFTER).

silesia zoom

At L4–5 native (0.333/0.330 @ ~46/38 MB/s) sits right on top of `miniz_oxide` (Rust) and ahead of `zlib`; the cost-rule shift (grey→red) nudges the whole lazy tier left at equal-or-better speed. `libdeflate` (C+SIMD) remains the top-left ceiling.

PR #2777, opened 2026-07-06
perf: incompressible-stretch skip-ahead in the lazy matcher (zstd-style)

### LINKED ISSUE #2766: perf(explore): incompressible-stretch skip-ahead in the lazy matcher (zstd-style acceleration)
depends-on: #2763

## Idea (from LZ4/zstd, absent from the zlib family)

On stretches with no matches (binary members: mozilla, x-ray, sao), the lazy matcher still pays full per-position price: hash, head probe, chain walk, insert. LZ4/zstd-style acceleration: after K consecutive positions without a match, search every 2nd position, then every 3rd, ... (resetting on the next match), while still inserting skipped positions into the hash chains (or inserting sparsely; sweep).

The ratio risk is bounded by construction: skipped positions are in regions that just demonstrated they have no matches, and any missed match start is re-findable at the next searched position minus at most the stride. On text (dense matches) the K threshold rarely triggers, so text ratio should be untouched.

perf (2026-07-05, L6): mozilla spends 18.6% in the chain walk plus ~17% in per-position loop overhead (`mainLoop`, `updateHashesFastU`) at 51 MB; x-ray and sao are the slowest members per MB at mid levels. This targets exactly those.

## Plan (measurement-first)

1. Spike: add `(missStreak, stride)` to the lazy main loop, stride ladder like zstd's (`stride = 1 + missStreak >> 5`, sweep the shift), measure speed and ratio per corpus member before proofs.
2. If it clears: thread through the packed/guarded twins (heuristic-only, contracts quantified over the policy).

Expected: large speed win on low-match members at L4-L8, near-zero on text; geomean ratio cost must stay in noise.

Sequencing: same loop as #2763 (and possibly #2765); claim after those land to avoid rebase churn.

🤖 Prepared with Claude Code

kim-em:
This PR adds an LZ4/zstd-style skip-ahead to the lazy hash-chain matcher: after a run of positions with no match, the matcher advances by a growing `stride = 1 + (missStreak >>> skipShift)` instead of one byte, emitting the skipped bytes as literals and resetting the miss streak on every match. On the incompressible binary members (mozilla, x-ray) this cuts the per-position chain-walk cost the matcher otherwise pays on regions that just demonstrated they have no matches; on text and dense-match data (sao) the miss threshold rarely triggers, so their ratio is untouched. The shipped ladder enables `skipShift = 7` for the lazy tier (levels ≥ 4) and disables it below; `skipShift = 63` (the default) keeps `stride` at 1 and reproduces the pre-skip output exactly.

The stride is a pure heuristic — skipped positions are re-findable at the next searched position and every emitted reference is re-verified at emission — so it never enters the correctness proof: the encoder contracts (`ValidDecomp` → `resolveLZ77`, encodability, empty) hold for *any* `skipShift`. `skipShift` and a `missStreak` accumulator thread through all three lazy-matcher twins (`lz77ChainLazy` / `lz77ChainLazyIter` / `lz77ChainLazyIterP`, proven mutually equal). The proofs gain three bridging lemmas — `litRun_valid` (a skip-ahead literal run extends a valid decomposition, mirroring `trailing_valid` but bounded) and `pushLits_eq` / `pushLitsP_eq` (the boxed and packed accumulator push forms equal `litRun`) — plus one mirrored branch in each twin-equality and encodability proof.

Measured with the shipped matcher (`bench/skip-spike`, pinned core), single-block compression over the Silesia binary members at chain 64 (L6), size delta vs. skip-ahead off:

| member | shift 7 speed | shift 7 size | shift 6 | shift 5 | shift 4 |
|--------|--------------:|-------------:|--------:|--------:|--------:|
| mozilla (51 MB) | — | +0.055% | +0.093% | +0.186% | +0.469% |
| x-ray (8 MB) | — | +0.097% | +0.362% | +1.036% | +2.274% |
| sao (7 MB) | — | 0% | 0% | +0.001% | +0.096% |
| text (alice/lcet) | — | 0% | 0% | <0.01% | <0.03% |
| 3-member speed | **+4.3%** | | +4.9% | +8.0% | +10.8% |

At chain 256 (L7) the win is larger where the chain walk dominates more (x-ray+sao: shift 5 +11%, shift 4 +20%). `sao` barely triggers (frequent short matches keep the streak short), so it correctly self-disables. The conservative `skipShift = 7` keeps the corpus geomean ratio cost in the noise while capturing the binary-member speedup; the exact per-level ladder is open for tuning via `bench/skip-spike` and the graphs.

Tests cover the twin equalities at `skipShift ∈ {4, 6, 7, 63}` (including `prng`, where skip-ahead is active), the `skipShift = 63` identity to the default matcher, and end-to-end roundtrip of the skip-ahead token stream through native inflate.

Closes #2766

🤖 Prepared with Claude Code

### DIFF
--- a/Zip/Native/Deflate.lean
+++ b/Zip/Native/Deflate.lean
@@ -1348,6 +1348,47 @@ theorem lazyAcceptCost_lt {len1 dist1 len2 dist2 : Nat}
   simp only [lazyAcceptCost, Bool.and_eq_true, decide_eq_true_eq] at h
   exact h.1
 
+/-! ## Incompressible-stretch skip-ahead literal runs (Wave 8, #2766)
+
+The lazy matcher's skip-ahead (LZ4/zstd-style) advances `pos` by a growing
+`stride` after a run of positions with no match, emitting every skipped byte as
+a literal. `litRun`/`pushLits`/`pushLitsP` are the three representations of "emit
+the literals for `data[pos .. stop)`": a `List` prepend (`litRun`, for the
+recursive reference matcher), and two `Array`-accumulator pushes (`pushLits` for
+boxed tokens, `pushLitsP` for packed `UInt32`). They advance one byte at a time
+and terminate on `stop - pos`, so `stop = pos + 1` (the disabled `skipShift`) is
+exactly the single-literal step the matchers took before this knob existed.
+`pushLits_eq`/`pushLitsP_eq` (in `LZ77ChainLazyCorrect`/`LZ77PackedCorrect`)
+relate the push forms back to `litRun` so the twin equalities carry through. -/
+
+/-- The literals `data[pos], …, data[stop-1]` as a `List` (empty once `pos`
+    reaches `stop` or the buffer end). The prepend form used by the recursive
+    lazy matcher's skip-ahead miss branch. -/
+def litRun (data : ByteArray) (pos stop : Nat) : List LZ77Token :=
+  if h : pos < stop ∧ pos < data.size then
+    .literal (data[pos]'h.2) :: litRun data (pos + 1) stop
+  else []
+termination_by stop - pos
+
+/-- `litRun` pushed onto a boxed-token accumulator (`Array LZ77Token`). Equal to
+    `acc ++ (litRun data pos stop).toArray` (`pushLits_eq`). -/
+def pushLits (acc : Array LZ77Token) (data : ByteArray) (pos stop : Nat) :
+    Array LZ77Token :=
+  if h : pos < stop ∧ pos < data.size then
+    pushLits (acc.push (.literal (data[pos]'h.2))) data (pos + 1) stop
+  else acc
+termination_by stop - pos
+
+/-- `litRun` pushed onto a packed-token accumulator (`Array UInt32`). Each push is
+    `packTok` of the token `pushLits` pushes, so
+    `pushLitsP (acc.map packTok) = (pushLits acc).map packTok` (`pushLitsP_eq`). -/
+def pushLitsP (acc : Array UInt32) (data : ByteArray) (pos stop : Nat) :
+    Array UInt32 :=
+  if h : pos < stop ∧ pos < data.size then
+    pushLitsP (acc.push (packTok (.literal (data[pos]'h.2)))) data (pos + 1) stop
+  else acc
+termination_by stop - pos
+
 /-- Lazy (one-byte-lookahead, zlib `deflate_slow`-style) variant of `lz77Chain`.
     At each position it finds the best chain match `M` at `pos`; before emitting it
     runs a *second* `chainWalk` at `pos+1` for `M'`. It defers — emits a literal at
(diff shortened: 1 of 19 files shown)


kim-em:
## Before/after `compress_mbps` (machine `Linux chungus2 x86_64`, native levels 1–10)

BEFORE = merge-base `ac14de13` (master, no skip-ahead) built + measured in a clean worktree; AFTER = this branch `8a6bc577`, re-measured after `rm -rf .lake` to clear the stale link-flag cache (the first overlay showed a uniform ~4% depression on the **greedy** levels 1–3, which this change does not touch — a cross-worktree `.lake` artifact, now removed: Silesia L1–3 sit at 1.00×).

**Silesia** (geomean over 12 files)
silesia pareto

**Canterbury** (geomean over 11 files)
canterbury pareto

The before/after native curves sit on top of each other on both corpora: at the **corpus geomean** the change is speed- and ratio-neutral (no regression). That is expected — skip-ahead only fires on incompressible stretches, so its win is concentrated in a few binary members and diluted at the corpus scale, well below single-pass Silesia noise.

### Ratio (deterministic — the clean signal)

`max |Δratio| = 0.0007` across the whole run, at **silesia/x-ray L6** only. Per-member size delta vs. skip-ahead off:

| member | Δ output size |
|--------|--------------:|
| mozilla | +0.05% |
| x-ray | +0.10% |
| sao | 0.00% (dense short matches — never triggers) |
| dickens / webster / nci / text | 0.00% |

Text and dense-match data are untouched; the cost is a few hundredths of a percent on the binary members. Corpus geomean ratio cost is in the noise.

### Speed (controlled same-binary measurement)

The corpus overlay above cannot resolve a member-specific sub-few-percent delta at 1 rep, so the speed verdict comes from the `bench/skip-spike` sweep, which toggles **only** `skipShift` within one build (pinned core, 3 reps) — the cleanest isolation of the skip-ahead effect. Single-block compression over the Silesia binary members at chain 64 (L6):

| shift | speed vs off | mozilla size | x-ray size |
|-------|-------------:|-------------:|-----------:|
| 7 (shipped) | **+4.3%** | +0.055% | +0.10% |
| 6 | +4.9% | +0.093% | +0.36% |
| 5 | +8.0% | +0.186% | +1.04% |
| 4 | +10.8% | +0.469% | +2.27% |

At chain 256 (L7) the win grows to +11% (shift 5) / +20% (shift 4) where the chain walk dominates more. The shipped ladder uses the conservative `skipShift = 7` (levels ≥ 4), keeping the corpus geomean ratio in the noise.

**Caveats:** the overlay is a cross-session, single-pass-Silesia comparison — read it for *ratio* (deterministic) and *no-regression positioning*, not for the fine speed delta; that comes from the controlled sweep. CI is green. The exact per-level shift ladder is open for tuning (`bench/skip-spike`).

kim-em:
## Corrected graphs + a straight answer on the "slight regression"

Two fixes from the earlier comment: (1) the graphs now include the other-language curves (I'd mistakenly passed a native-only file as the reference source), and (2) I re-measured the speed properly, because you were right to be suspicious of that overlay.

### The overlay was the wrong instrument — and I over-claimed earlier

The whole-corpus before/after overlay sits neutral-to-slightly-below because skip-ahead only fires on incompressible stretches, so its effect is confined to 2-3 Silesia members and drowned by the text-heavy majority + single-pass Silesia noise. It genuinely cannot resolve this delta, and reading it as a "slight regression" is fair.

silesia pareto

### Controlled same-binary A/B (the honest number)

One build, only `skipShift` toggled (63 = off vs 7 = shipped), 7 interleaved reps, pinned core. `dickens` is the text control — its ~0.3-0.5% reading is the noise floor.

| file | matcher off→on | full L6 `deflateRaw` off→on |
|------|---------------:|----------------------------:|
| mozilla | **+6.6%** | **+2.0%** |
| x-ray | +2.0% | +0.4% |
| sao | −0.4% (noise) | −0.1% (noise) |
| dickens (text control) | +0.5% (noise) | +0.3% (noise) |

So: **it is a small real win on the incompressible binary members, not a regression** — but more modest than the "+4.3%" I first quoted. That figure was the base-path (L4/L5, no block-split) measurement; at L6 the split path runs a *second* emit pass that skip-ahead doesn't touch, diluting the matcher's +6.6% (mozilla) down to +2.0% of the full compress. mozilla's +2.0% is ~4× the noise floor; x-ray/sao/text are at the floor.

Ratio cost is unchanged from before and negligible: mozilla +0.05%, x-ray +0.10%, everything else 0 (`max |Δratio| = 0.0007`, x-ray L6 only).

### Honest call for you

At the shipped conservative `skipShift = 7` this is roughly "+2% on mozilla, neutral-to-slightly-positive elsewhere, ~0 ratio cost" — a small win for real proof complexity. If that's not worth it, reasonable options are: (a) ship a more aggressive shift (5-6) to trade a little more ratio for a clearer speed win on the binary tier, (b) enable skip-ahead only at the base-path levels L4/L5 where it isn't diluted by the split's second emit pass, or (c) hold it. Your call — I didn't want to present the earlier +4.3% as if it were the production number.

CI is green. `bench/skip-spike --ab 7 <files>` reproduces the table.

kim-em:
Closing — measured but not landing. The skip-ahead win is confined to 2 incompressible Silesia members (mozilla ~+2%, x-ray up to +5.4% at aggressive shifts) and dilutes through the L6+ block-split's second emit pass, while the larger x-ray win costs measurable ratio (+0.36–1.04%). Full write-up and the shift 5/6/7 speed-vs-ratio table recorded on #2766. The mechanism is proven and re-derivable from this branch's history if a latency tier ever wants it.

PR #2778, opened 2026-07-06
perf(#2767): merged-array chain state removes the per-position Prod (+2.2% geomean)

### LINKED ISSUE #2767: perf(explore): fuse the lazy matcher's walk+insert step (delete the per-position pair alloc and boundary RC)
depends-on: #2763

## Evidence (generated C + perf, 2026-07-05, master 52a0d6d9)

Reading `.lake/build/ir/Zip/Native/Deflate.c` for the lazy matcher's per-position orchestration:

- `updateHashesFastU___redArg` **allocates its `(hashTable, prev)` result pair on every call** (`lean_alloc_ctor` x2 in the body): one heap pair + unpack (`lean_ctor_get` x2) + free per match position, visible at 11.2% (L6 dickens) / 6.1% (mozilla).
- `lz77ChainLazyIterP_mainLoop` compiles to a 54 KB body with ~120 `lean_dec` sites, 32 `lean_inc`, and pair unpacking, at 10.5% / 6.5%. Per position it crosses four function boundaries (hash, head probe, chain walk, insert), each with RC traffic on the two table arrays.

Prior art to respect: #2749 measured a *threaded persistent pair* variant **negative** (-1.9%: `isShared` checks + two cell stores per position). This issue proposes the other shape: **fuse** the walk and the insert into one step function that takes and returns the two arrays as separate arguments (no pair), or restructure `mainLoop` so the tables never cross a helper boundary (single well-founded loop owning all state). The goal is deleting the per-position ctor/unpack and shrinking the RC choreography, not adding persistent-state checks.

## Plan (measurement-first)

1. Spike the fused step with the equality lemma `sorry`'d; confirm byte-identical output; measure in situ at L4-L8 on dickens + mozilla.
2. Read the regenerated C and confirm the `lean_alloc_ctor` pair and the boundary `inc/dec` actually disappeared (the arbiter of whether de-boxing happened).
3. If the win clears the bar, do the equality transfer; if not, record the negative result next to #2749's.

Sequencing: rewrites the same loop as #2763/#2765/#2766; claim last in that family.

🤖 Prepared with Claude Code

kim-em:
This PR resolves #2767 — deleting the per-matched-position `(hashTable, prev)` Prod that `updateHashesFastU` returns — by holding the chain state in **one `Array Nat`** laid out `prev ++ hashTable` (prev ring at offset 0, hash table after), so the interior-hash walk returns a single array with no pair while the main loop stays one self-tail-recursive `goto` loop. `lzMatchP` dispatches to the merged matcher at levels 4+, proven equal to `lz77ChainLazyIterP` with **no `sorry`**. Measured **+2.2% geomean** across all 12 Silesia files (every one positive), byte-identical output.

The issue's own proposal — *fuse* the walk and insert into a separate step function threading the two arrays as separate args — was tried first and is **measured negative**: −24…−47% on dickens and a **stack overflow** on mozilla. Splitting the single self-tail-recursive loop into two mutually-recursive functions loses whole-function refcount linearity (insert self-time tripled 9.3%→28.5% on a byte-identical body) and is not tail-call-optimized across the boundary. A `partial def` control of the unfused loop measured at baseline, pinning the cause on the split.

The *merge* shape avoids both. With `prev` at offset 0 the 34.5%-hot chain walk reads `c[cand & 0x7FFF]` unchanged (the combined array is passed straight in where `prev` was); only the once-per-position hash accesses take a `+prevSize` offset. Generated C confirms the insert walk has 0 `lean_alloc_ctor` (was 2) and returns a single array with no pair unpack.

Measured with a self-controlled sandwich (base→merged→base, both binaries built in the same worktree, pinned, min-of-N; base = geomean of the bracketing runs) — the single-pass Silesia dashboard *overlay* reads ~1.00x because a ~2% delta sits below its ±30% noise floor, so the sandwich is the trustworthy signal. **The insert loop runs proven-bounds `set`** (`updateHashesMergedFast`, reached once per matched position through the guarded dispatch `updateHashesMergedGuarded`): an earlier `guardedSet` cut with a bounds branch per insert step *regressed* insert-dense xml −2%; the proven-bounds path makes it uniform. All 12 Silesia at L6:

    dickens +3.7%  ooffice +3.0%  mozilla +2.8%  mr +2.7%   x-ray +2.6%
    osdb +2.5%     webster +2.5%  samba +2.4%    reymont +1.8%  sao +1.8%
    xml +1.0%      nci +0.1%

Proof (`Zip/Spec/LZ77MergedCorrect.lean`), all `sorry`-free: the 4 `Array` append helpers, `set_eq_set!`, `chainWalk_append` (the walk reads only `[0, prevSize)` where `c` and `prev` agree), `chainWalkGuardedPackedU_append`, `updateHashesMerged_append`, `updateHashesMergedFast_eq`/`updateHashesMergedGuarded_eq`, the `updateHashes` size-preservation lemmas, `mergedLoop_eq` (the lockstep), and `lz77ChainLazyIterPMerged_eq`. `lake exe test` green. Both negatives (this fusion and #2749's persistent-pair reuse) and the positive merge are recorded in `progress/lazy-fuse-walk-insert-2767.md`.

Note: the committed dashboard (`bench/results/latest.json`) is **not** refreshed in this PR — the machine has several concurrent benchmark jobs running, which would depress the single-pass native numbers and mislead the next PR's baseline. The dashboard refresh should be a quiet-machine follow-up; the sandwich above (self-controlled, contention-cancelling) is the perf evidence.

🤖 Prepared with Claude Code


kim-em:
**Before/after `compress_mbps`** (machine `chungus2`; overlay is single-pass Silesia so a ~2% delta sits *below* its ±30% noise floor — the curves overlap, which is expected, not evidence of no win):

silesia pareto

The trustworthy signal is the **self-controlled sandwich** (base→merged→base, same worktree, pinned, min-of-7), with the **proven-bounds** insert loop:

| file | L4 | L6 | L8 | type |
|---|---|---|---|---|
| dickens | +5.5% | +4.6% | +3.4% | text |
| xml | +1.6% | +1.9% | +1.3% | dense text |
| webster | +1.9% | +0.7% | +3.2% | text |
| mozilla | +2.1% | +0.5% | +1.0% | binary |
| sao | +2.4% | +1.7% | +3.4% | binary |

Output byte-identical everywhere (`max |Δratio| = 0.000000`). The committed code currently uses `guardedSet` (matches the existing proof) which regresses xml −2%; the proven-bounds switch above is required before shipping. WIP: `mergedLoop_eq` lockstep is one `sorry`.

kim-em:
**Final proven-bounds pareto** (`compress_mbps`, machine `chungus2`):

silesia pareto

Native (red) sits mid-pack — above JS fflate at the fast end, below zlib/Go/miniz, with libdeflate the SIMD ceiling. Before/after **overlap** on the log axis: the change is output-neutral (`max |Δratio| = 0.000000`) and the ~1–2% speed delta on the lazy tier is below single-pass Silesia's ±30% noise, so the curves land on top of each other. That's expected — the pareto shows *positioning*, not a 2% delta.

Where the delta *does* show through (the levels my change touches — L4–8 lazy tier — vs the unchanged greedy L1–3 and optimal-parse L9–10):

| tier | Silesia | Canterbury |
|---|---|---|
| L1–3 (greedy, unchanged) | 1.00x | 1.00x |
| **L4–8 (lazy, this PR)** | **1.01x** | **1.01x** |
| L9–10 (optimal parse, unchanged) | 1.00x | 1.00x |

The real magnitude is the self-controlled sandwich (base→merged→base, same worktree): **+2.2% geomean at L6, all 12 Silesia files positive** (dickens +3.7%, xml +1.0%, nci +0.1%). Single-pass understates it; the sandwich and the L4–8-only lift here agree it's a real, uniform, output-neutral win isolated to the lazy tier.

kim-em:
**Re-measured pinned to an idle core** (`taskset -c 13`, confirmed 0% busy) and refreshed the committed dashboard (`bench/results/latest.json` + graphs) in this PR. Native rows re-measured on the merged matcher, all 10 levels; reference-language rows reused.

silesia pareto

Before/after native, geomean by level (output-neutral, `max |Δratio| = 0.000000`):

| level | Silesia | Canterbury | |
|---|---|---|---|
| L1–3 (greedy, unchanged) | 1.00x | 1.00x | control ≈ 0 |
| **L4 (lazy, this PR)** | **1.02x** | **1.01x** | |
| **L5–8 (lazy, this PR)** | **1.01x** | **1.01x** | |
| L9–10 (optimal parse, unchanged) | 1.00x | 1.00x | control ≈ 0 |

The lift is isolated to the L4–8 lazy tier my change touches; the unchanged greedy (L1–3) and optimal-parse (L9–10) tiers sit at 1.00x (±0.7%), the binary-health control. Cross-session dashboard understates the same-session sandwich (+2.2% geomean at L6, all 12 files positive), but the sign, the output-neutrality, and the tier-isolation are unambiguous. Dashboard refresh committed.

PR #2789, opened 2026-07-07
perf: re-grid the L6-L8 tuning ladder — L6 chain 64→128 fixes the L6<L5 inversion

### LINKED ISSUE #2785: perf(tune): re-grid the L6-L8 tuning ladder (chainDepth/niceLen/goodMatch + block-split thresholds) after the Huffman-build + lazy-accept landings
## Re-grid the L6-L8 tuning ladder after the Huffman-build + lazy-accept landings

The per-level knob ladder ([`chainDepth`](Zip/Native/DeflateDynamic.lean#L589) / [`niceLen`](Zip/Native/DeflateDynamic.lean#L651) / [`goodMatch`](Zip/Native/DeflateDynamic.lean#L625), and the [block-split thresholds](Zip/Native/DeflateDynamic.lean#L1235)) was gridded before several cost-shifting changes landed (native Huffman build #2769, cost-based lazy accept #2776, half-depth probe #2774, fused split pass #2762). The per-file cost balance has moved, so the current knees may no longer be knees. Zero-proof-risk sweeps — pure retune of heuristics.

Specific things to re-check:

- **The `chainDepth` L6=64 < L5=128 inversion** ([here](Zip/Native/DeflateDynamic.lean#L595)) — L6 goes *shallower* than L5 because splitting compensates. Re-confirm the inversion still pays post-Huffman-rebuild.
- **`niceLen` at L8 = 258 (cutoff disabled)** ([here](Zip/Native/DeflateDynamic.lean#L658)) — a cutoff may earn its speed back now; sweep 130/180/258.
- **`goodMatch` = 259 (gate off) for L5+** ([here](Zip/Native/DeflateDynamic.lean#L627)) — re-check whether a gate at L8 buys speed for tolerable ratio.
- **Block-split `minBlockBytes` / `softMaxBlockBytes` / `checkTokens`** — re-grid per level for L6-L8.

## Plan

Run `ratio-sweep` (both a text and a binary member per the dickens-vs-mozilla lesson) across the grids, adopt the new knees, and record the re-baselined ladder. This should run *first* among the L6-8 ideas — it re-baselines everything the structural bets are measured against.

🤖 Prepared with Claude Code

kim-em:
This PR re-grids the L6-L8 tuning ladder after the native-Huffman-build (https://github.com/kim-em/lean-zip/pull/2769 "native Huffman build"), cost-based-lazy-accept (https://github.com/kim-em/lean-zip/pull/2776 "cost-based lazy accept rule"), half-depth-probe (https://github.com/kim-em/lean-zip/pull/2774 "half-depth lazy lookahead probe") and fused-split-pass (https://github.com/kim-em/lean-zip/pull/2762 "fuse split-heuristic passes") landings shifted the per-file cost balance, raising L6's `chainDepth` from 64 to 128 so that level 6 (the default) no longer compresses text worse than level 5. Every knob touched here is a search-or-emit heuristic re-verified at emission (`lz77ChainIter_resolves` holds for any depth/cap), so no proof changed and `lake exe test` stays green.

L6=64 relied on the observation-divergence split to compensate for a shallow chain, but the re-grid (new `bench/l68-sweep`, run over dickens + mozilla + Canterbury) shows the split only compensates on binary. On text, L6=64+split compressed *worse than L5*'s single block — dickens +0.19%, plrabn12 +0.10%, ptt5 +0.18% — a ladder-monotonicity defect on the default level. Chain 128 (sharing L5's depth, adding only the split) restores L6 ≤ L5 output on every file measured and gains ~0.15% ratio on the dickens+mozilla pair, for ~7% L6 throughput (25.08 → 23.31 MB/s, pinned idle core, 3 interleaved reps). L6=128 stays clearly faster than L7=256 (23.31 vs 21.05 MB/s), so ladder ordering holds. This is a default-level-behavior fix first, aggregate ratio second.

The same re-grid re-confirmed the other three L6-L8 knobs at their current values, recorded as null results (see `progress/l68-regrid-2785.md` for the tables):

- **niceLen L8 = 258** — a cutoff earns no speed (it does not fire on dickens/mozilla) and costs up to +0.5% ratio on image data (ptt5).
- **goodMatch L5+ = 259** — an L8 gate buys +15% speed but +0.40% ratio, a bad trade at a ratio tier.
- **split minBlockBytes/softMaxBlockBytes/checkTokens = 10000/300000/512** — the same grid optimum for L6/L7/L8 (no per-level split warranted); minBlockBytes 5000 gains ~0.12% binary ratio but costs ~2.5% speed on all of L6-L8.

Closes https://github.com/kim-em/lean-zip/issues/2785

🤖 Prepared with Claude Code

### DIFF
--- a/Zip/Native/DeflateDynamic.lean
+++ b/Zip/Native/DeflateDynamic.lean
@@ -566,12 +566,29 @@ attribute [irreducible] symbolBitCount fixedBlockBytes dynBlockBytes dynBlockByt
     (Silesia grid over chain × `goodMatch` × split, then interleaved pinned
     timing): L4 = 64 (with the lazy gate, the old L5 point), L5 = 128 (gate
     off — `goodMatch` buys more ratio per cycle than deepening the chain, so
-    L5 = (128, gate off) dominates the old L7 = (256, gate off)). **L6's depth
-    drops back to 64 on purpose**: the split tier starts there, and at equal
-    cycles the observation-divergence split + a shallow chain beats a deep
-    chain without the split — the block-split, not the chain, is L6's budget.
-    L7/L8 deepen the chain again within the split tier (256/512); past that
-    the chain saturates.
+    L5 = (128, gate off) dominates the old L7 = (256, gate off)). L6 shares
+    L5's depth 128 and adds the observation-divergence split on top: the split
+    is L6's whole budget increment over L5. L7/L8 deepen the chain again within
+    the split tier (256/512); past that the chain saturates.
+
+    **L6 previously dropped to 64** on the theory that a shallow chain plus the
+    split, at equal cycles, would beat a deep chain — but the #2785 re-grid
+    (`l68-sweep`, dickens + mozilla + Canterbury, after the native-Huffman #2769,
+    cost-based-lazy-accept #2776, half-depth-probe #2774 and fused-split #2762
+    landings shifted the balance) refuted that: chain 128 + split beats chain 64
+    + split on **every** file measured, and at chain 64 the split no longer
+    compensated on text — L6 (the default level) compressed *worse than L5* on
+    dickens (+0.19%), plrabn12 (+0.04%) and ptt5, a ladder-monotonicity defect.
+    Raising L6 to 128 restores L6 ≤ L5 output everywhere and gains ~0.15% ratio
+    on the dickens+mozilla pair, for ~7% L6 throughput (pinned, interleaved); L6
+    stays clearly faster than L7 (chain 256), so the ladder ordering holds. The
+    same re-grid re-confirmed `niceLen` L8 = 258 (a cutoff earns no speed — it
+    does not fire on these workloads — and only ever costs ratio), `goodMatch`
+    L5+ = 259 (an L8 gate buys +15% speed but +0.40% ratio, a bad trade at a
+    ratio tier), and the global block-split floors (`minBlockBytes` 5000 gains
+    ~0.12% binary ratio but costs ~2.5% speed on all of L6-L8, against the
+    speed-ward direction of the recent landings). See `progress/` for the full
+    tables.
 
     Level 1 is the `deflate_fast` corner (#2726): depth `4` is exactly zlib
     `-1`'s `max_chain`. A tokens-held-constant attribution on Silesia (see
@@ -591,8 +608,7 @@ def chainDepth (level : UInt8) : Nat :=
   else if level ≤ 2 then 16
   else if level ≤ 3 then 32
   else if level ≤ 4 then 64
-  else if level ≤ 5 then 128
-  else if level ≤ 6 then 64
+  else if level ≤ 6 then 128
   else if level ≤ 7 then 256
   else if level ≤ 8 then 512
   else 1024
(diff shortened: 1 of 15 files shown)


kim-em:
## Before/after `compress_mbps` — L6 chain 64→128

Machine `Linux chungus2 x86_64` (matches the dashboard). BEFORE (L6=64) and AFTER (L6=128) were both built and measured **in the same worktree** (temporarily reverting only the `chainDepth` line), so there is no cross-worktree `.lake` artifact. Native levels 1–10; Canterbury median-of-5, Silesia single-pass.

**The change is L6-only and it is a strict ratio win.** Every untouched level reads 0.99–1.00x (control), confirming a clean measurement; L6 moves left (better ratio on **every** corpus file, worst-compressing files gain most) for a ~6–8% L6 throughput cost. This fixes the pre-change defect where L6 (the default level) compressed text *worse than L5* (dickens/plrabn/ptt5). L6=128 stays clearly faster than L7, so ladder ordering holds.

### Silesia (geomean over 12 files)
silesia compress pareto

### Canterbury (geomean over 11 files)
canterbury compress pareto

### Per-level geomean (speedup = after/before)

| level | Canterbury spd | Silesia spd |
|------:|:--------------:|:-----------:|
| 1–5   | 0.99–1.00x     | 0.99–1.00x  |
| **6** | **0.94x**      | **0.92x**   |
| 7–10  | 0.99–1.00x     | 0.99–1.00x  |

### L6 ratio delta (all files improve; representative)

| file | Δratio | L6 speed |
|------|-------:|---------:|
| silesia/reymont | **−1.18%** | 0.79x |
| silesia/xml | −0.95% | 0.91x |
| silesia/nci | −0.75% | 0.90x |
| silesia/dickens | −0.24% | 0.91x |
| silesia/mozilla | −0.13% | 0.92x |
| canterbury/kennedy.xls | −0.61% | 0.86x |

Caveats: metric is `compress_mbps` (this is a compression change — the ratio moves, intentionally, only at L6; `max |Δratio| = 0.0033` at silesia/reymont L6). Silesia is single-pass so its per-file speed carries run-to-run noise, but the ratio deltas are deterministic and the L6 speed cost matches an independent pinned interleaved sweep (`l68-sweep`). The dashboard (`bench/results/latest.json` + `bench/graphs/*.svg`) is refreshed with the AFTER numbers in this same PR.

🤖 Prepared with Claude Code

kim-em:
Closing: this change **shrinks the Pareto frontier**, so it should not land.

I optimized per-level ratio and level-vs-level monotonicity, not the achievable (ratio, speed) envelope — which is the only thing that matters. The mixing-curve test (compress a fraction at the higher level, the rest at the lower; linear in ratio, linear in sec/MiB) shows the old L6=64 is already an **outside vertex** of the L5↔L7 curve, and deepening the chain to 128 pulled it *inside*:

| ratio | before | after |
|------:|-------:|------:|
| 0.325 | 33.7 MB/s | 32.7 (**−2.9%**) |
| 0.324 | 32.5 | 31.4 (**−3.6%**) |
| 0.323 | 31.5 | 30.1 (**−4.2%**) |

At every ratio L6 serves, the AFTER frontier is slower. The "ratio improved on every file" result was real but off the boundary — a blend of L5 and L7 already reached those ratios faster. Monotonicity between named levels is not a frontier property.

Applying the same test to the rest of the re-grid, niceLen cutoffs and the L8 gate are also inside/dominated. Net result of #2785: **null — no knob setting expands the frontier; the current ladder is already on it.** Recorded on the issue.

Issue #2787, opened 2026-07-07
perf(explore): optimal-parse 64 MiB gate collapses L9/L10 to L8 ratio on large streams (windowed DP)


kim-em:
## Symptom

On a single stream larger than 64 MiB (the 212 MB `silesia.tar`), native levels **9 and 10 produce byte-identical output** (out_size 67,987,859 for both) and both collapse to the L6–8 split ratio — losing the optimal-parse crown:

| native level | 1-shot tar | pooled per-file | Δ vs per-file |
|---|---|---|---|
| L8 | 0.3209 | 0.3209 | ~0% |
| L9 | 0.3208 | 0.3137 | **+2.3%** |
| L10 | 0.3208 | 0.3096 | **+3.6%** |

Per-file (every file ≤ 51 MB) L9 (fast-optimal) and L10 (exact-optimal) are distinct and reach the crown; on the tar they are the same encoder. This also surfaces on the compress-Pareto as two coincident markers (now labelled `L9=L10` after the plotter fix).

## Root cause

The optimal-parse candidate is gated to inputs ≤ 64 MiB by `optimalMaxSize := 67108864` (`Zip/Native/DeflateDynamic.lean:1504`). Above the gate, the `deflateRaw` dispatch falls through to the base-vs-shared-split branch (`:1805`–`:1831`), which is identical for L9 and L10 (their `lzMatchP` parameters — chainDepth 1024, insertCap 1e9, goodMatch 259, niceLen 258, lazyDepth 512 — are equal). The gate is deliberate: the backward-DP parser holds ~27 B of transient state per input byte (≈5.7 GB at 212 MB), which is why it is capped.

## Proposed direction

Regionalize / window the optimal DP so it runs in bounded memory and scales past 64 MiB (process the stream in overlapping windows, or cap live DP state independent of input size), letting L9/L10 keep the crown on large inputs. Alternatively, at minimum, raise/adapt the gate as memory allows. Larger proof surface than the L1–5 split fix — lower priority.

## Scope note

Deliberate memory-safety behaviour today, and invisible per-file (all files < gate). Improvement, not a regression. Sibling finding to the L1–5 single-block defect.

PR #2790, opened 2026-07-07
perf: windowed optimal parse keeps the L9/L10 crown past the 64 MiB gate

kim-em:
This PR removes the 64 MiB collapse of native levels 9 and 10 by running the optimal parser in bounded memory above the `optimalMaxSize` gate, so the crown survives on large streams instead of falling through to the L8 split ratio (issue https://github.com/kim-em/lean-zip/issues/2787, "perf(explore): optimal-parse 64 MiB gate collapses L9/L10 to L8 ratio on large streams (windowed DP)").

The gate exists because `computeChoices` allocates two `data.size`-sized `Array Nat` choice arrays — the exact-DP peak's dominant cost (measured ≈2/3 of the resident set at 64 MiB: 1534 MB total, of which ≈1 GB is those arrays). This adds a windowed twin of both the exact and L9-fast parsers that caps live choice storage to a single region: it fills each region's choices into region-local (relative-indexed) buffers, emits that region's tokens immediately via an interleaved emitter, and discards the buffers before the next region, threading the hash-chain finder state across regions exactly as `computeChoicesLoop` does. Because the regions and carried state match the global parse, the token stream is byte-identical — only the choice storage is bounded (`ZipTest.OptimalParse` asserts the equality on the region-crossing inputs). Measured: 64 MiB peak 1534 MB → 543 MB, same tokens, same speed.

`deflateRaw` dispatches to the windowed candidate above the gate at levels 9 and 10; below the gate the existing global parse is unchanged. End-to-end on a 99 MiB input (above the gate): L8 0.2474, L9 0.2389, L10 0.2338 — the crown is kept and L9/L10 are distinct again instead of both collapsing to L8.

The windowed emitter re-verifies every match exactly as `optimalEmit` (the emission guard plus a fresh `countMatch`), so `windowedEmit_valid`/`_encodable` are the same argument extended by one non-emitting refill branch; the region fills are heuristic and opaque to correctness, using panic-checked reads that never fire (`base + r ≤ data.size` holds by construction). Validity, encodability, resolves, and empty contracts are proved once against the shared driver and specialize to both parsers, and the full `deflateRaw` roundtrip quadruple carries through the restructured dispatch.

🤖 Prepared with Claude Code

### DIFF
--- a/Zip/Native/DeflateDynamic.lean
+++ b/Zip/Native/DeflateDynamic.lean
@@ -1529,6 +1529,28 @@ def deflateDynamicBlocksOptimalFast (data : ByteArray) (tokChunk : Nat) : ByteAr
   else
     (emitSharedBlocks data (lz77OptimalFastIter data) tokChunk 0 BitWriter.empty).flush
 
+/-- Windowed twin of `deflateDynamicBlocksOptimal` (#2787): identical block
+    stream, but the exact-DP tokens come from `lz77OptimalWindowedIter`, whose
+    live choice storage is capped to one region — so the exact crown runs in
+    bounded memory past the `optimalMaxSize` gate. -/
+def deflateDynamicBlocksOptimalWindowed (data : ByteArray) (tokChunk : Nat) : ByteArray :=
+  if data.size == 0 then
+    let f := tokenFreqs #[]
+    (emitDynBlock BitWriter.empty data #[] (dynamicCodeLengths f.1 f.2).1 (dynamicCodeLengths f.1 f.2).2
+      (dynamicCodeLengths_length f.1 f.2).1 (dynamicCodeLengths_length f.1 f.2).2 true).flush
+  else
+    (emitSharedBlocks data (lz77OptimalWindowedIter data) tokChunk 0 BitWriter.empty).flush
+
+/-- Windowed twin of `deflateDynamicBlocksOptimalFast` (#2787): the region-capped
+    L9-fast parse (`lz77OptimalWindowedFastIter`). -/
+def deflateDynamicBlocksOptimalWindowedFast (data : ByteArray) (tokChunk : Nat) : ByteArray :=
+  if data.size == 0 then
+    let f := tokenFreqs #[]
+    (emitDynBlock BitWriter.empty data #[] (dynamicCodeLengths f.1 f.2).1 (dynamicCodeLengths f.1 f.2).2
+      (dynamicCodeLengths_length f.1 f.2).1 (dynamicCodeLengths_length f.1 f.2).2 true).flush
+  else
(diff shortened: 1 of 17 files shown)


kim-em:
## Before/after — `compress_mbps` (machine `chungus2`)

**Caveat up front:** this is a *memory* optimisation, and its win is invisible to the per-file speed-vs-ratio dashboard — every dashboard file is < 64 MiB, so the below-gate global path runs unchanged and the output is byte-identical. The two overlay graphs below confirm that (no regression, `max |Δratio| = 0.000000`); the effect the dashboard **cannot** show is in the first graph.

### The actual effect (above the 64 MiB gate)

crown recovery + memory

Measured on the same 99 MiB compressible input, above the gate:

| level | before (master) | out (before) | after (this PR) | out (after) |
|---|---|---|---|---|
| L8  | 0.2474 | 25,770,333 | 0.2474 | 25,770,333 |
| L9  | 0.2473 | **25,750,868** | **0.2389** | 24,885,740 |
| L10 | 0.2473 | **25,750,868** | **0.2338** | 24,351,766 |

Before this PR, above the gate L9 and L10 fall through to the split path and produce **byte-identical** output (25,750,868 for both) at essentially the L8 ratio — exactly the collapse issue #2787 reports. With the windowed parse they are distinct again and reach the crown (L10 beats L8 by 5.5%, L9 by 3.4%). Peak resident set for the exact-DP parse at 64 MiB drops **1535 MB → 543 MB** (2.8×, same tokens, same speed) because the two `data.size`-sized choice arrays are now region-local.

### No per-file regression (the dashboard view)

`max |Δratio| = 0.000000` (output-neutral) and 0.99–1.01x speed on every level, both corpora — the native before (grey) sits exactly under after (red):

silesia
canterbury

```
silesia    native compress_mbps geomean       canterbury  native compress_mbps geomean
lvl  ratio   before  after  speedup           lvl  ratio   before  after  speedup
 1   0.361    84.8   84.7    1.00x              1   0.328    61.8   61.6    1.00x
 6   0.323    31.0   30.9    1.00x              6   0.297    28.5   28.4    1.00x
 8   0.320    24.6   24.5    1.00x              8   0.295    23.6   23.5    1.00x
 9   0.314     4.3    4.3    1.00x              9   0.288     4.6    4.5    0.99x
10   0.309     2.6    2.6    1.00x             10   0.285     2.6    2.6    0.99x
```

BEFORE = `origin/master` (4480a60a), AFTER = this branch, both built and measured in the same worktree (no cross-worktree noise). The dashboard commit in this PR refreshes the committed `latest.json`/SVGs (it was stale since #2778, an output-neutral speed change that landed after the last snapshot).

PR #2792, opened 2026-07-07
perf: reuse split-sizing frequencies for the base candidate (#2772)

### LINKED ISSUE #2772: perf: derive whole-file frequencies from per-block (split-pass follow-up to #2762)
Follow-up to #2762 (split-pass perf). Items 1–3 landed in https://github.com/kim-em/lean-zip/pull/2771; this issue tracks the remaining item 4.

## Item

Derive the whole-file frequencies used by the base candidate by summing the per-block frequencies the split-sizing pass already computes, instead of a second full walk. `tokenFreqsP` (4.7% L6 dickens, 3.0% mozilla) runs whole-file for the base candidate (`deflateRawBasePPrep`, `Zip/Native/DeflateDynamic.lean`) while `sharedPartitionSizedP` computes per-block `tokenFreqsP` over the same tokens. Whole-file frequencies are (up to an end-of-block correction) the sum of the per-block frequencies: one per-block pass plus a cheap ~316-entry summation replaces the second full walk.

## Why this is a separate PR (proof cost)

Unlike the proof-free split heuristic (items 1–3), the base candidate feeds the **verified** roundtrip/padding specs (`deflateRawBasePPrep_emit`, `inflate_deflateRaw`, `deflateRaw_pad`). So this needs a **proven** equality, not just a conformance pin:

- `tokenFreqsP` pre-counts the end-of-block symbol 256 with value 1 (`Zip/Native/DeflateFreqs.lean:146`), so summing per-block frequencies over k blocks over-counts symbol 256 by (k−1) — an EOB correction is required.
- The equality must hold over the **same clamped partition** the emitter uses (`emitSharedBlocksAtP` / `sharedPartitionSizedP` recursion), so it has to be threaded through that recursion.
- There is no existing `tokenFreqs` additivity infrastructure.

## Suggested scope

1. Frequency-array element-wise addition lemmas.
2. An `tokenFreqsP`-over-append lemma with the EOB correction (`tokenFreqsP (a ++ b) = merge (tokenFreqsP a) (tokenFreqsP b)`, `merge` subtracting 1 at index 256).
3. A clamped-partition additivity theorem (`tokenFreqsP toks = EOB-corrected fold of per-block `tokenFreqsP` over the clamped cuts).
4. A freq-taking `deflateRawBasePPrep` variant whose emit theorem stays definitionally clean, wired into the levels-6–8 dispatch so the per-block freqs computed during sizing are reused for the base candidate when the split has cuts.

Expected: recover most of the `tokenFreqsP` cost (~3–5% compress at L6) on top of #2771, byte-identical output.

🤖 Prepared with Claude Code

kim-em:
This PR derives the whole-stream frequencies the levels-6–8 base candidate needs by summing the per-block frequencies the split-sizing pass already computes, instead of a second whole-stream `tokenFreqsP` walk. `tokenFreqsP` costs 4.7% of compress at L6 dickens / 3.0% at L6 mozilla; on inputs with cuts the base candidate no longer pays for it a second time — one per-block pass plus a cheap ~316-entry summation replaces the second full walk. Output is byte-identical and the verified roundtrip/padding specs are preserved.

The base candidate feeds the verified specs (`inflate_deflateRaw`, `deflateRaw_pad`), so this needs a proven equality rather than a conformance pin. `Zip/Spec/DeflateFreqsAdditive.lean` proves `tokenFreqsP` additivity over concatenation:

    tokenFreqsP (a ++ b) = mergeEOBFreqsP (tokenFreqsP a) (tokenFreqsP b)

where `mergeEOBFreqsP` is element-wise addition with a `−1` correction at the end-of-block symbol (256), which every per-block histogram pre-counts once. The proof characterizes each histogram entry as its initial value plus the number of processed words that bump it (`litDeltaP`/`distDeltaP`), and those bump-counts are additive over `++`; the only shared over-count is EOB, which no word ever bumps.

`sharedPartitionSizedFreqsP` fuses the frequency summation into the existing clamped-partition sizing recursion: it reuses the same per-block `tokenFreqsP` array `sizedTrees` already consumes, folding it into the whole-stream histogram with `mergeEOBFreqsP`. `Zip/Spec/DeflateBaseFreqsReuse.lean` proves its first component is definitionally `sharedPartitionSizedP` (so every sizing/emit theorem transfers unchanged) and its second component is `tokenFreqsP toks` (via the additivity theorem over the clamped cuts). The levels-6–8 dispatch, when the split has cuts, feeds those frequencies to a freq-taking `deflateRawBasePPrepF` whose emit theorem stays definitionally clean (`deflateRawBasePPrepF_obsFreqs` recovers `deflateRawBasePPrep`); when the split has no cuts the base still walks the whole stream, unchanged.

Follow-up to items 1–3 (https://github.com/kim-em/lean-zip/pull/2771). Closes https://github.com/kim-em/lean-zip/issues/2772.

🤖 Prepared with Claude Code

### DIFF
--- a/Zip/Native/DeflateDynamic.lean
+++ b/Zip/Native/DeflateDynamic.lean
@@ -1341,6 +1341,32 @@ decreasing_by
   simp only [Nat.not_le] at h
   omega
 
+/-- Fused twin of `sharedPartitionSizedP` (#2772): one pass over the clamped
+    partition yields the same `(bits, per-block trees)` **and** the whole-stream
+    frequencies, accumulated as the EOB-corrected element-wise sum of the per-block
+    `tokenFreqsP` (`mergeEOBFreqsP`) — the *same* histogram `sizedTrees` already
+    consumes, so no extra frequency walk. Component 1 is proved equal to
+    `sharedPartitionSizedP` (`sharedPartitionSizedFreqsP_fst`) and component 2 to
+    `tokenFreqsP (toks.extract pos toks.size)` (`sharedPartitionSizedFreqsP_snd`),
+    so the base candidate can reuse these frequencies instead of re-walking the
+    whole stream (`tokenFreqsP` is 4.7% L6 dickens / 3.0% mozilla of compress). -/
+def sharedPartitionSizedFreqsP (toks : Array UInt32) (cuts : List Nat) (pos : Nat) :
+    (Nat × List SizedTrees) × (Array Nat × Array Nat) :=
+  let j := min (max (cuts.headD toks.size) (pos + 1)) toks.size
+  let f := tokenFreqsP (toks.extract pos j)
+  let t := sizedTrees f.1 f.2
+  let blockBits := 3 + (writeDynamicHeader BitWriter.empty t.val.1 t.val.2).bitLength
+    + symbolBitCount f.1 f.2 t.val.1.toArray t.val.2.toArray
+  if j ≥ toks.size then ((blockBits, [t]), f)
+  else
+    let rest := sharedPartitionSizedFreqsP toks cuts.tail j
+    ((blockBits + rest.1.1, t :: rest.1.2), mergeEOBFreqsP f rest.2)
+termination_by toks.size - pos
+decreasing_by
+  rename_i h
+  simp only [Nat.not_le] at h
+  omega
+
 /-- Tree-taking twin of `emitSharedBlocksAtP`: same clamped cut points, but each
     block's `(litLens, distLens)` come from the `trees` list (in lockstep with
     `cuts`) instead of being recomputed from the group via `tokenFreqsP` +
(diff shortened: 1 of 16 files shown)


kim-em:
Before/after native `compress_mbps` (machine `Linux chungus2 x86_64`, median-of-5 on Canterbury, single-pass on Silesia; other-language curves reused).

**Output is byte-identical:** `max |Δratio| = 0.000000` (the change is a proven output-neutral frequency reuse). The speedup lands exactly where predicted — the levels-6–8 split-with-cuts path on the large Silesia files — while every unaffected level (L1–5, L9–10) sits at 1.00x, a clean binary-health control.

**Silesia** (large real-world files — the affected path):

| 5 | 0.328 | 38.3 | 38.1 | 1.00x |
| **6** | 0.323 | 30.9 | 31.8 | **1.03x** |
| **7** | 0.321 | 26.4 | 27.0 | **1.02x** |
| **8** | 0.320 | 24.6 | 24.7 | 1.01x |
| 9 | 0.314 | 4.3 | 4.3 | 1.01x |
| 10 | 0.309 | 2.5 | 2.5 | 0.99x |

**Canterbury** (small synthetic files — mostly below the split-cut threshold, so little change):

| 6 | 0.297 | 28.3 | 28.5 | 1.01x |
| 7 | 0.296 | 25.0 | 25.2 | 1.01x |
| 8 | 0.295 | 23.5 | 23.6 | 1.00x |

compress pareto silesia

compress pareto canterbury

Caveats: single-digit-% delta measured cross-session (BEFORE is the base-branch `latest.json` snapshot, commit `1a899305`; AFTER is `dcf4b958`). The affected-vs-unaffected level pattern (L6–8 up, everything else 1.00x) and the byte-identical `Δratio = 0` are the signal; the magnitude matches the issue's ~3–5% `tokenFreqsP` cost estimate at L6. The dashboard (`bench/results/latest.json` + `bench/graphs`) is refreshed with the AFTER numbers in this PR.

PR #2804, opened 2026-07-08
perf(decode): allocation-free overlapping match copy

kim-em:
This PR replaces the allocation-heavy overlapping-match copy in the DEFLATE decoder with a single-pass, overlap-aware extern `ByteArray.extendWithin`, following the extern-with-reference-body pattern of `ByteArray.copyWithin`. The overlap path (`length > distance`, i.e. short-distance / RLE matches) previously computed `buf ++ (fillDouble (buf.extract start (start+distance)) length).extract 0 length` — an `extract` alloc, `~log2(length/distance)` doubling appends (each an alloc plus two memcpys), a second `extract`, and a final append, on the order of six allocations to copy ten bytes at `distance = 1`. `ByteArray.extendWithin` grows the buffer in place and smears the `distance`-byte window forward with memcpy doublings (a `memset` fast path for the `distance = 1` RLE case), with no intermediate allocation.

The extern's reference body is exactly the previous `fillDouble`-based expression, so `copyLoop`'s overlap branch and every decode correctness proof are unchanged up to unfolding the extern; `copyLoop_eq_ofFn` gains a single `rw [ByteArray.extendWithin]`. `fillDouble` moves to the `ByteArray` namespace alongside the new primitive. The C guards `oldsz + len` against `size_t` overflow (unlike `copyWithin`, `len` here is unbounded) and clamps the effective period to the window `extract` produces, so the body stays a faithful model. A new `ZipTest/ExtendWithin.lean` sweeps the extern against an independent per-byte reference across every in-bounds `(srcOff, distance, length)` — the RLE `memset` path, the doubling smear with non-power-of-2 lengths, the `length ≤ distance` short case, the empty-window edges, copy-on-write, and the in-place growth chain the decoders take.

🤖 Prepared with Claude Code

### DIFF
--- a/Zip/Native/CopyWithin.lean
+++ b/Zip/Native/CopyWithin.lean
@@ -16,8 +16,8 @@
   `extract` is total and clamps its upper bound to `a.size`, so this is the
   **non-overlapping** path only (`srcOff + len ≤ a.size`); the C clamps `len` to
   match, so source and destination never overlap and the body stays a faithful
-  model. The overlapping / RLE case (`len > distance`) stays in pure Lean
-  (`fillDouble`) and does not use this primitive.
+  model. The overlapping / RLE case (`len > distance`) is handled by the sibling
+  primitive `ByteArray.extendWithin` (Zip/Native/ExtendWithin.lean).
 
   This is a project-local stopgap for the upstream Lean-core primitive proposed
   in lean#14158 (`feat: add ByteArray.copyWithin`); the owner has approved it as
--- a/Zip/Native/ExtendWithin.lean
+++ b/Zip/Native/ExtendWithin.lean
@@ -1,0 +1,67 @@
+/-!
+  Single-pass, allocation-free overlapping-match copy for the DEFLATE
+  back-reference.
+
+  `ByteArray.extendWithin a srcOff distance len` appends `len` bytes to `a`'s
+  own tail, where appended byte `i` is `a[srcOff + (i % distance)]` — the
+  periodic extension of the `distance`-byte window `a[srcOff, srcOff + distance)`.
+  This is the LZ77 **overlapping** back-reference (`len > distance`, e.g. RLE),
+  where the source window and the bytes being written overlap.
+
+  The reference body — and the trusted specification of the `@[extern]` — is
+  literally
+
+      a ++ (fillDouble (a.extract srcOff (srcOff + distance)) len).extract 0 len
+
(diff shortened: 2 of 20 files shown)


kim-em:
## Native decode throughput — before/after

Metric: `decompress_mbps` (decode-density experiment: every decoder timed on **byte-identical libdeflate raw-DEFLATE streams**, silesia, levels 1/3/6/9/12). Machine `Linux chungus2 x86_64` (matches the committed dashboard). BEFORE = `HEAD~1` decoder, AFTER = this branch's decoder, built and measured **back-to-back in the same worktree**. The change is provably output-neutral: **`max |Δratio| = 0.000000`** — only decode speed can move.

The win is exactly where the issue predicted — the short-distance / RLE-heavy overlapping-match path — and it is large on the most repetitive text, tapering to noise elsewhere:

| file | L1 | L3 | L6 | L9 | L12 |
|------|----:|----:|----:|----:|----:|
| **dickens** | **+28.0%** | **+19.4%** | **+15.3%** | +6.8% | −0.2% |
| x-ray | +3.4% | +0.6% | +1.3% | +1.7% | +3.3% |
| mr | +2.8% | +2.9% | +1.8% | +2.2% | +3.8% |
| mozilla | +3.2% | +0.3% | −0.3% | −0.3% | +0.1% |
| *(nci, ooffice, osdb, reymont, samba, sao, webster, xml)* | — | — | — | — | within ±2% noise |

silesia-wide native decode geomean: **L1 1.03x, L3 1.02x, L6 1.01x, L9 1.01x, L12 1.00x** (the per-file dickens win is diluted across 12 files). The gain concentrates at fast levels, where greedy parsing emits many overlapping/RLE matches; at L9/L12 the optimal parse produces fewer, so the delta vanishes.

per-file decode speedup

Positioning among the other decoders (AFTER solid red vs BEFORE dashed grey; C/Rust/SIMD curves reused): the pure-Lean decoder stays in its ~150–200 MB/s band below zlib/miniz_oxide/libdeflate — this change lifts it slightly, it does not close that gap.

decode throughput vs ratio

Caveats: silesia decode-density is single-pass (±1–2% run-to-run noise on the flat files), but the dickens signal clears that floor by ~10x and is back-to-back same-worktree, so it is not a cross-session artifact. Dashboard (`latest.json`, `decode_density.json`, silesia decode SVGs) refreshed in-PR in commit 8ef216fa.

PR #2808, opened 2026-07-08
perf(decode): bind the packed Huffman table entry once per symbol

kim-em:
This PR binds the packed Huffman decode-table entry **once** per symbol and reads `len`/`sym` out of that single word, replacing the four separate `lenAt`/`symAt` reads — each a guarded `dite` with its own bounds check — the shipped decode loop performed per literal. A new `DecodeTable.entryAt` does one guarded `packed[idx]` load returning the raw `UInt32`; `unpackLen`/`unpackSym` then extract the fields in native `UInt8`/`UInt16`/`USize` widths, so the compiled literal path lowers to one `lean_array_get` plus register ops instead of four reads and four boxed bounds checks with tagged-`Nat` round-trips. The change lands in the production `goTreeFreeU`, its tree-based twin `goFusedPU`, and `decodeSymCanon`; the boxed reference loops `goTreeFree`/`goFusedP` are left unchanged as the correctness anchor, and equivalence transfers through the `lenAt`/`symAt` = `unpackLen`/`unpackSym` ∘ `entryAt` bridge lemmas plus the restated `litGuard_usize`. It also corrects the `Inflate.lean` docstring that falsely claimed the compiler CSEs the shared read into one load — the generated C shows it does not, which is exactly why binding the entry once in the source is what collapses the four reads.

Measured with deterministic `perf stat` (same worktree, dispatch target the only difference) on Silesia and Canterbury payloads: roughly **5% fewer retired instructions and 1-2% fewer cycles** on the decode across `x-ray`, `ooffice`, `sao`, `alice29`, and `plrabn12`, never a regression. This is a real but modest win, well below the +40-80% the issue estimated: the read-count reduction is genuine, but the reads were cheap (cache-resident, tagged-scalar bounds checks), so the wall-clock gain is small. A follow-up will thread the `packed.size = 2^fastBits` invariant so the remaining guarded read becomes a proven-bounds `uget`, dropping the last per-literal bounds check.

Full library builds with no `sorry`; all tests pass. A second opinion (Codex) found no correctness or soundness concern in the definitions or the equivalence bridging.

🤖 Prepared with Claude Code

### DIFF
--- a/Zip/Native/Inflate.lean
+++ b/Zip/Native/Inflate.lean
@@ -160,14 +160,16 @@ where
 
 /-- Pack a slot's `(symbol, codeLen)` into one `UInt32` word: the codeword length
     in the low byte (bits 0–7), the 16-bit symbol in bits 8–23. libdeflate-style
-    single-word entry (libdeflate also uses a `u32`). The source reads the slot
-    through `lenAt`/`symAt`, which name the same `packed[idx]!` element; the
-    compiler CSEs the shared read, so a per-symbol decode lowers to **one**
-    `lean_array_get` plus register shifts (verified in the generated C for the
-    fastloop literal path), rather than two reads into the parallel `syms`/`lens`
-    arrays of #2650 (which also touched two cache lines). Even without the CSE it
-    is two reads of one tagged-scalar array, never the old two-array / two-cache-
-    line shape.
+    single-word entry (libdeflate also uses a `u32`). The hot decode loops read the
+    slot **once** per symbol via `entryAt` (a single guarded `packed[idx]` load) and
+    extract `len`/`sym` from that one word with `unpackLen`/`unpackSym` (a low-byte
+    truncation and a shift, register-only), so a per-symbol decode lowers to one
+    `lean_array_get` plus register ops. The
+    compiler does **not** CSE separate `lenAt`/`symAt` calls — each `dite` read
+    lowers to its own `lean_array_get_size` + `lean_nat_dec_lt` — so binding the
+    entry once in the source is what collapses the four reads (see `entryAt`). This
+    is one read of one tagged-scalar array, never the two reads into the parallel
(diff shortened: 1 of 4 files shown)


kim-em:
## Decode perf: bind the packed table entry once

Metric: retired instructions (deterministic) and CPU cycles, **branch vs merge-base `f715089e`**, same machine (`chungus2`), decoding libdeflate-encoded raw-DEFLATE streams with `Inflate.inflate` ×20 (pre-sized output).

before/after

| file | instructions | cycles |
|------|-------------:|-------:|
| alice29 | −4.4% | −1.3% |
| plrabn12 | −4.5% | −0.2% |
| ooffice | −5.2% | −0.8% |
| sao | −6.3% | −2.1% |
| x-ray | −5.4% | −2.1% |

The instruction reduction is deterministic and rock-stable (byte-identical across runs); the generated C for `goTreeFreeU`'s literal path now does **one** `lean_array_get` + one bounds check instead of four. The cycle reduction is smaller (~1-2%) — the removed instructions were cheap, cache-resident, tagged-scalar bounds checks with high ILP, so IPC dips slightly and wall-clock moves only ~1-2%. This is well below the issue's +40-80% estimate; the read-count reduction is real but those reads were not the decode bottleneck.

**Caveats.** This uses the deterministic instruction count as the primary evidence because the routine cross-session before/after Pareto overlay cannot honestly resolve a sub-2% delta (the skill's own guidance), and this session lacks the Go/JS/OCaml/Zig comparator toolchains needed to refresh the committed `decode_density.json` dashboard without dropping their curves — so the dashboard refresh is deferred to the dev machine rather than committed incomplete here.

A follow-up PR will thread the `packed.size = 2^fastBits` invariant so the remaining guarded read becomes a proven-bounds `uget`, dropping the last per-literal bounds check.

PR #2809, opened 2026-07-08
perf: enable -flto so the wide-load stopgap externs inline into the hot loops

### LINKED ISSUE #2806: perf: -flto to inline the wide-load stopgap externs out of countMatch/hash3 (~+2–5% compress, byte-identical)
Enable link-time optimization (`-flto`) for the shipped library so the wide-load stopgap externs (`ByteArray.ugetUInt64LE`, `ByteArray.ugetUInt32LE`, `UInt64.ctz`) inline into the hot matcher loops instead of staying as cross-object function calls, a measured byte-identical **+2–5%** end-to-end compress win on this toolchain.

## Why there is a win

The wide-load helpers in `Zip/Native/Wide.lean` are `@[extern]` functions whose bodies live in `c/bytearray_wide_ffi.c`, compiled to a separate object. The Lean compiler emits a `call` to the external symbol, and because lake compiles/links the C **without** `-flto`, that call survives to the binary — GCC/clang `-O3` inlining is per-translation-unit, so it cannot fold a definition that lives in another object without LTO.

This bites in the single hottest loop of the compressor, `lz77Greedy.countMatch` (the match-extension byte comparison), which per iteration calls the externs. Disassembly of the shipped bench binary, inside `countMatch`:

```
2 <lean_zip_uget_u64le>   # the 8-byte wide load, compared per iteration
1 <lean_zip_ctz64>        # first-mismatch byte index
```

These are non-inlined FFI calls on the critical path. `hash3` has the same pattern with `ugetUInt32LE` (one wide load per input position).

## Measured effect (spike)

I hand-relinked a bench executable with only `Zip/Native/Deflate.o` and `bytearray_wide_ffi.o` recompiled at lake's exact flags (`-O3 -DNDEBUG -fvisibility=hidden`) plus `-flto -fno-semantic-interposition`. Verified the calls inlined (`countMatch`: 3 extern calls → 0) and that compressed output is **byte-identical** to the non-LTO build (same `out=` / `ratio=` on every file).

Steady-state `deflateRaw` throughput (`compress-pareto`, AMD EPYC 9455, `taskset`-pinned):

| file | level | non-LTO | LTO | delta |
|------|-------|---------|-----|-------|
| lcet10 (text, 427 KB) | 6 | 14.44 / 14.51 / 14.64 MB/s | 15.23 / 15.19 / 15.26 | **+4.3 … +5.5%** (3 reps) |
| mozilla (binary, 51 MB) | 6 | 27.13 | 28.18 | **+3.9%** |
| x-ray (binary, 8.5 MB) | 6 | 17.36 | 17.70 | +2.0% |
| x-ray (binary, 8.5 MB) | 8 | 17.33 | 17.51 | +1.0% |

Text gains most (more/longer matches → `countMatch` dominates more of the pass). The win is modest but consistent and repeatable, and it is free of any ratio change.

## Scope / caveats to resolve

- **Overlaps with lean#14053 (`wide fixed-width load/store`).** These externs exist *specifically* as a project-local stopgap until core lands `uget*` as compiler primitives, which will inline **without** LTO (see the `Zip/Native/Wide.lean` module docstring). So the headline win here has a shelf life: once the toolchain bumps to a Lean with lean#14053, this particular benefit folds into that primitive. LTO may still help other cross-module inlining, but that is unmeasured. Worth weighing against the maintenance cost of an LTO build config.
- **Build-time cost is the main open question.** LTO moves codegen to link time, so the final link gets slower (and the spike only relinked two objects; a full-library `-flto` build recompiles every object to bitcode). This needs measuring on a clean full build before adoption — a large link-time regression could outweigh a ~3% runtime win for a library.
- **Lakefile integration.** The spike was a manual relink. Productionizing means threading `-flto` through `moreLeancArgs`/link args for the shipped `Zip` lib (and the extern-lib C targets), and confirming it composes with the prebuilt toolchain runtime (which cannot be LTO'd, so runtime calls stay out-of-line — fine, the matcher-internal externs are what matter).
- **Coverage.** The spike LTO'd only the matcher path (`Deflate.o` + the wide-load C). A full-library LTO might pick up additional inlining in the emit path (BitWriter / Huffman encode also cross object boundaries) or hit diminishing returns; measure end-to-end across the whole corpus, not just these four files.
- **Correctness.** LTO is a codegen-only transform; output was byte-identical here, and the roundtrip + fuzz tests and the encode/decode proofs guard semantics regardless.

## Not a #2779 spinoff in substance

This surfaced while investigating #2779 (packed `ByteArray` chain tables, measured negative), but it is a distinct and orthogonal lever: #2779 was about the *chain-table storage representation*; this is about *inlining the wide-load externs in the byte-comparison loop*. Keeping it as its own work item per the repo's PR-scope rule. The #2779 verdict (packing does not pay on ≥512 KB-L2 hardware) stands unchanged.

🤖 Prepared with Claude Code

kim-em:
This PR enables link-time optimization for the shipped library on Linux so the wide-load stopgap externs (`ByteArray.ugetUInt64LE`, `ByteArray.ugetUInt32LE`, `UInt64.ctz`) inline into the hot matcher loops (`lz77Greedy.countMatch`, `hash3`) instead of remaining per-iteration cross-object calls, a measured +1–6% end-to-end compress win with byte-identical output.

Closes https://github.com/kim-em/lean-zip/issues/2806 ("perf: -flto to inline the wide-load stopgap externs out of countMatch/hash3").

The Lean-emitted C gains `-flto` via `moreLeancArgs`, links go through `-flto -O3`, and the three project-local FFI primitives (`bytearray_wide`, `extend_within`, `copy_within`) switch from the system `cc` to the toolchain's own C compiler (`buildLeanO`) so all bitcode comes from one LLVM — the system clang here is version 22, whose bitcode the toolchain's lld (LLVM 19) cannot read. Two consequences of `buildLeanO`'s `-nostdinc` header sandbox: the copy primitives use `__builtin_memcpy`/`__builtin_memset` instead of `<string.h>`, and `copy_within_ffi.o` gains `-O2 -DNDEBUG` like its siblings (a clang `-O0` object carries `optnone`, which blocks LTO inlining). LTO is Linux-only (macOS ld64's libLTO need not accept toolchain bitcode; Windows untested) with a `LEAN_ZIP_LTO=0` opt-out.

Verified on this branch (EPYC 9455, taskset-pinned, steady-state `deflateRaw` via `compress-pareto`): every `lean_zip_*` extern call site in the bench binary inlines away (`countMatch`: 3 → 0), compressed sizes are byte-identical across 7 corpus files × levels {1,6,9}, and the full test suite plus bench conformance pass.

| file | level | master | LTO | delta |
|------|-------|--------|-----|-------|
| lcet10 (text, 427 KB) | 6 | 30.51 MB/s | 32.31 MB/s | **+5.9%** |
| mozilla (binary, 51 MB) | 6 | 27.74 | 29.10 | **+4.9%** |
| x-ray (binary, 8.5 MB) | 6 | 17.92 | 18.08 | +0.9% |
| x-ray (binary, 8.5 MB) | 8 | 17.83 | 18.10 | +1.5% |

Clean-build wall time is unchanged (~5 min, elaboration-dominated); the executable link steps grow from ~0.2 s to ~10 s (`test`) / ~3 s (`bench`). A flag-free downstream package (`require lean-zip`, no LTO flags of its own) builds, runs, and gets the same inlining — the toolchain lld consumes bitcode archive members natively — so consumers need no changes. A Codex second opinion reviewed the design; its concerns (downstream link-arg propagation, `LEAN_CC`/`LEAN_AR` overrides, env-toggle staleness against Lake's cached `run_io` config) are addressed in the `ltoFlags` docstring and verified by the downstream-consumer test above.

As the issue notes, the headline win has a shelf life: once the toolchain bumps past lean#14053 (`uget*` as compiler primitives) that part folds into core, while the other now-inlined externs (`extend_within`, `push_u64le`, `copy_within`) keep benefiting.

🤖 Prepared with Claude Code

### DIFF
--- a/bench/graphs/canterbury_compress_heatmap.svg
+++ b/bench/graphs/canterbury_compress_heatmap.svg
@@ -37,20 +37,20 @@ L 95.67625 49.4355
 z
 " style="fill: #ffffff"/>
    </g>
-   <g clip-path="url(#paf08724384)">
+   <g clip-path="url(#p2cec6f8756)">
     <image xlink:href="data:image/png;base64,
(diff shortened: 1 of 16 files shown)


kim-em:
Before/after `compress_mbps` Pareto (machine `Linux chungus2 x86_64` — matches the dashboard; BEFORE = master's committed `bench/results/latest.json`, AFTER measured on this branch, pinned via `bench/pin_core.sh`; other-language curves reused, not re-measured). Output-neutral as expected for a codegen-only change: `max |Δratio| = 0.000000` across all 230 rows.

compress pareto silesia

compress pareto canterbury

Native `compress_mbps` geomean by level:

| lvl | ratio | silesia before | after | speedup | | ratio | canterbury before | after | speedup |
|-----|-------|--------|-------|---------|-|-------|--------|-------|---------|
| 1 | 0.361 | 84.2 | 90.7 | **1.08x** | | 0.328 | 61.6 | 64.8 | 1.05x |
| 2 | 0.349 | 62.9 | 68.9 | **1.10x** | | 0.313 | 48.7 | 52.2 | 1.07x |
| 3 | 0.344 | 57.3 | 62.3 | **1.09x** | | 0.310 | 45.0 | 47.9 | 1.07x |
| 4 | 0.331 | 45.9 | 49.1 | 1.07x | | 0.303 | 38.1 | 40.3 | 1.06x |
| 5 | 0.328 | 38.2 | 40.7 | 1.07x | | 0.300 | 32.3 | 34.1 | 1.06x |
| 6 | 0.323 | 31.9 | 33.7 | 1.06x | | 0.297 | 28.9 | 30.1 | 1.04x |
| 7 | 0.321 | 27.0 | 28.4 | 1.05x | | 0.296 | 25.4 | 26.6 | 1.04x |
| 8 | 0.320 | 25.1 | 26.5 | 1.05x | | 0.295 | 23.9 | 25.0 | 1.05x |
| 9 | 0.314 | 4.3 | 4.5 | 1.04x | | 0.288 | 4.6 | 4.7 | 1.03x |
| 10 | 0.309 | 2.6 | 2.7 | 1.04x | | 0.285 | 2.6 | 2.7 | 1.04x |

Every level gains on both corpora, no level regresses, and the gain is largest at the fast end (L1–L3, where the matcher externs dominate the pass). Visually the AFTER curve is a clean vertical lift of the whole frontier — at L1–L3 on Silesia it visibly closes on the zlib curve. Decode also gains ~**+3.5%** from LTO (same-worktree A/B via `decode-ab`: ref 269.8 → 279.3 MB/s, buf 215.8 → 223.6 MB/s on lcet10 L6).

One discovery from the dashboard refresh (committed in this PR): master's `latest.json` **decompress** rows were stale — they predate the #2804 extendWithin decode win (that PR's refresh appears to have captured pre-change decode numbers, e.g. lcet10 L6 recorded 160.8 MB/s vs ~272 MB/s actually measured on the identical merge-base tree). The refreshed rows correct this; the large decode delta in this PR's `latest.json` diff is therefore mostly #2804's already-landed win being recorded properly, **not** an LTO effect. `bench/results/decode_density.json` (a separate recording) may carry the same staleness and deserves a re-record in a follow-up.

kim-em:
Regenerated after rebasing onto current master (#2807, #2808). Same protocol as [the previous comment](https://github.com/kim-em/lean-zip/pull/2809#issuecomment-4916271495): `compress_mbps`, machine `Linux chungus2 x86_64`, BEFORE = master's committed `latest.json`, AFTER pinned on this branch; still output-neutral (`max |Δratio| = 0.000000` across all 230 rows).

compress pareto silesia

compress pareto canterbury

| lvl | silesia | canterbury |
|-----|---------|------------|
| 1 | 1.08x | 1.06x |
| 2 | 1.09x | 1.08x |
| 3 | 1.09x | 1.07x |
| 4 | 1.08x | 1.08x |
| 5 | 1.07x | 1.08x |
| 6 | 1.06x | 1.06x |
| 7 | 1.06x | 1.07x |
| 8 | 1.06x | 1.07x |
| 9 | 1.04x | 1.03x |
| 10 | 1.03x | 1.03x |

The in-PR dashboard refresh now also folds in #2808's decode win (master's `latest.json` was not refreshed by #2808, and its decode rows already predated #2804 — see #2810).

PR #2814, opened 2026-07-08
perf(decode): proven-bounds uget for the literal fast-path read

kim-em:
This PR replaces the guarded `entryAt` read in the production tree-free decode loop's literal fast path with a proven-bounds `uget`, following up #2808. A new `DecodeTable.entryAtU` reads the packed slot by `Array.uget` on the `USize` window index, so the compiled literal path drops both the per-literal bounds check (`lean_array_get_size` + `lean_nat_dec_lt`) and the `UInt64→Nat` index round-trip that `entryAt`'s `dite` kept — the generated C for `goTreeFreeU`'s literal read is now a single `lean_array_uget_borrowed` with a `size_t` index and no size/`nat_dec_lt` guard. The bound `idx.toUSize.toNat < packed.size` is discharged from a threaded invariant `litTable.packed.size = 2^fastBits`, established at the table builders (`buildTableCanonicalFastWithCount_size`, from `buildCanonicalLoop_size` + `Array.size_replicate`, which move into the Native layer so the decoder's caller can use them without importing the spec).

Equivalence to the boxed reference is preserved and reproven: `entryAtU_window_eq` shows the `uget` slot is the same word as the guarded `entryAt` read, `litGuardU_usize` restates the literal guard on it, and `goTreeFreeU_eq` is reproven threading the invariant — the only changed case is the literal fast path, which recovers the original `lenAt`/`symAt` guard.

Measured on the same payloads as #2808 (deterministic `perf stat`, branch vs #2808 as baseline): **~2-3% fewer retired instructions and ~1-3% fewer cycles** across `x-ray`, `ooffice`, `sao`, `alice29`, and `plrabn12` — on top of #2808's ~5% / ~1-2%. Full build green with no `sorry`; all tests pass. A Codex second opinion found no soundness concern in the bound proof, the `entryAtU`/`entryAt` bridge, the size-lemma move, or the reproven equivalence.

🤖 Prepared with Claude Code

### DIFF
--- a/Zip/Native/Inflate.lean
+++ b/Zip/Native/Inflate.lean
@@ -270,6 +270,51 @@ theorem DecodeTable.symAt_eq_unpackSym_entryAt (t : DecodeTable) (idx : Nat) :
   unfold DecodeTable.symAt DecodeTable.entryAt
   split <;> rfl
 
+/-- `entryAt` with the bounds proof supplied by the caller: an unchecked `uget` on
+    a `USize` index, so the hot loop pays **no** per-symbol bounds check and no
+    `USize→Nat` index round-trip. The `2^fastBits`-slot invariant on `packed`
+    (`buildTableCanonicalFastWithCount_size`) plus `bitBuf &&& 0x7FF < 2^fastBits`
+    discharge the proof once at the call site. Equal to `entryAt` in bounds
+    (`entryAtU_eq_entryAt`), so every decode proof still transfers. -/
+@[inline] def DecodeTable.entryAtU (t : DecodeTable) (i : USize)
+    (h : i.toNat < t.packed.size) : UInt32 :=
+  t.packed.uget i h
+
+theorem DecodeTable.entryAtU_eq_entryAt (t : DecodeTable) (i : USize)
+    (h : i.toNat < t.packed.size) : t.entryAtU i h = t.entryAt i.toNat := by
+  unfold DecodeTable.entryAtU DecodeTable.entryAt
+  rw [dif_pos h]; rfl
+
+/-- The `fastBits`-bit window index, taken as a `USize`, indexes the `2^fastBits`
+    packed slots. `bitBuf &&& 0x7FF ≤ 0x7FF = 2^fastBits − 1`, and the `USize`
+    round-trip can only shrink it (`mod`), so the bound survives. -/
+theorem and_0x7FF_toUSize_toNat_lt (bitBuf : UInt64) :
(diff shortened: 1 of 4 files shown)


kim-em:
## Decode perf: proven-bounds `uget` for the literal read

Metric: retired instructions (deterministic) and CPU cycles, **branch vs #2808** (the merged baseline), same machine (`chungus2`), `Inflate.inflate` ×20 on libdeflate-encoded raw-DEFLATE streams, pre-sized output.

before/after

| file | instructions | cycles |
|------|-------------:|-------:|
| alice29 | −1.8% | −0.8% |
| plrabn12 | −1.8% | −1.6% |
| ooffice | −2.3% | −1.7% |
| sao | −3.0% | −2.4% |
| x-ray | −2.5% | −2.9% |

This is on top of #2808's ~5% instructions / ~1-2% cycles. Removing the last per-literal bounds check (`lean_array_get_size` + `lean_nat_dec_lt`) and the `UInt64→Nat` index round-trip turned out to be worth more than the flat estimate — the generated C for the literal read is now a single `lean_array_uget_borrowed` on a `size_t` index. The instruction counts are deterministic (byte-identical across runs); cycles track at ~1-3%.

Same caveat as #2808: the deterministic instruction/cycle counts are the evidence for this small delta (a cross-session Pareto overlay can't resolve a sub-3% change), and the full cross-language decode-density dashboard refresh needs the Go/JS/OCaml/Zig comparator toolchains not present in this session, so it's deferred to the dev machine.

kim-em:
### Decode-density "Pareto" (positioning)

Since decode speed doesn't trade off against ratio, the decode Pareto is the throughput-vs-input-density view plus the ranking lollipop (native vs the in-process C/Rust references + the `memcpy` bandwidth ceiling), on identical libdeflate-L6 Silesia streams:

ranking
density

Native (lean-zip) sits at **289 MB/s** Silesia geomean — behind zlib (466), miniz_oxide (531), and libdeflate (904), well under the ~36 GB/s memcpy ceiling.

**Important read:** this chart shows native's *position*, not this PR's *delta*. A ~2-3% shift is far below the visual resolution of a log MB/s axis where native is ~3× behind libdeflate — the two red clusters (before/after) would overlap. The [deterministic instruction/cycle bar chart above](https://github.com/kim-em/lean-zip/pull/2814#issuecomment-4916605772) is the artifact that actually resolves the improvement.

Caveats: the Go/JS/OCaml/Zig comparator curves are absent (those toolchains aren't in this session), and this is not committed to the dashboard (committing the 5-decoder snapshot would drop the committed dashboard's other-language rows) — a full-field refresh belongs on the dev machine.

PR #2816, opened 2026-07-08
perf(decode): build copy_within_ffi.o with -O2 -DNDEBUG like its siblings

### LINKED ISSUE #2811: perf(decode): compile copy_within_ffi.o with -O2 -DNDEBUG like its siblings
The \`copy_within_ffi.o\` target in \`lakefile.lean\` passes no optimization flags — unlike its siblings \`extend_within_ffi.o\` and \`bytearray_wide_ffi.o\`, which both pass \`-O2 -DNDEBUG\` with a comment saying the lean.h helpers "only inline away under optimization" (the \`extend_within\` comment even says "for the same reason as bytearray_wide_ffi", suggesting \`copy_within\` was simply missed when the flags were added).

## Evidence

\`objdump -d\` on the \`inflate-profile\` binary: \`lean_zip_byte_array_copy_within\` contains out-of-line calls to \`lean_sarray_cptr\` (×3), \`lean_usize_of_nat\` (×2), \`lean_sarray_size\`, \`lean_sarray_set_size\`, \`lean_sarray_ensure_capacity\`, \`lean_sarray_capacity\`, \`lean_is_exclusive\`, \`lean_dec\`, \`lean_alloc_sarray\` — all \`static inline\` lean.h helpers that -O2 folds. \`lean_zip_byte_array_extend_within\` in the same binary has none of these (only \`memcpy\`/\`memset\`/alloc/panic). Because \`-DNDEBUG\` is also missing, the helpers run **with lean.h asserts enabled**: \`lean_to_sarray\`'s assert chain (\`lean_is_sarray\` → \`lean_ptr_tag\`) shows up as real symbols.

\`perf record\` of \`inflate-profile decode\` on silesia \`x-ray\` (level 6, the #2795 A/B baseline at 41b1b857):

| symbol | self% |
|---|---|
| goTreeFreeU (the decode loop itself) | 34.8% |
| lean_to_sarray | 9.1% |
| lean_zip_byte_array_copy_within | 7.5% |
| lean_is_sarray | 6.4% |
| lean_usize_of_nat | 4.0% |
| lean_is_exclusive | 3.9% |
| lean_sarray_set_size | 3.6% |
| lean_ptr_tag / lean_is_st / lean_sarray_cptr / lean_is_scalar | ~9% combined |

i.e. roughly a **third of whole-program decode self time** is the un-optimized copy_within TU — pure compilation overhead on the LZ77 match-copy path, on every decode of match-bearing data.

## Work

- Add \`-O2 -DNDEBUG\` (+ the existing \`-fPIC\`) to the \`copy_within_ffi.o\` target's \`hardArgs\`, mirroring \`bytearray_wide_ffi.o\` and its rationale comment.
- Re-measure decode throughput (this should be a large across-the-board decode win on match-heavy files) and note that it **re-baselines** decode percentage claims: recent A/B percentage deltas (e.g. #2793/#2808, the #2795 verdict) were measured against a denominator inflated by this overhead.

## Context

Found while attributing the #2795 batched-literal-emission regression (see the #2795 verdict in \`plans/track-d-state.md\`). The fix is one lakefile hunk plus the existing conformance tests (\`copyWithin\`'s reference body and the roundtrip suite already witness behavior).

🤖 Prepared with Claude Code

kim-em:
This PR builds the `copy_within_ffi.o` target with `-O2 -DNDEBUG` (keeping the existing `-fPIC`), so it matches its sibling primitives `extend_within_ffi.o` and `bytearray_wide_ffi.o`, which already carry these flags for the same documented reason. `ByteArray.copyWithin` backs the non-overlapping LZ77 back-reference path in `Zip.Native.Inflate.copyLoop`, so it sits on the decode hot path; leaving it at the default optimization level was an oversight when the flags were added to its siblings.

The flags strip the debug-only `lean.h` asserts from the release object and let the runtime static-inline helpers fold: the optimized `lean_zip_byte_array_copy_within` body drops from 191 to 104 instructions, with all nine out-of-line assert calls gone.

The end-to-end decode impact is compiler-dependent, and I want to be precise rather than overclaim. On gcc 15.2, which already inlines the `lean.h` helpers even at `-O0`, a same-worktree A/B (two saved `inflate-profile` binaries, `taskset`-pinned, canterbury/`plrabn12`) measured about 0.6% fewer retired instructions with wall time and cycles within noise. On toolchains where `-O0` leaves those helpers outlined (the configuration behind the original profile in the issue, where `copy_within` plus `lean_to_sarray`/`lean_is_sarray`/`lean_sarray_cptr` accounted for roughly a third of decode self-time), the win is larger. So this is best read as consistency with the two siblings plus removing a compiler-dependent `-O0` codegen cliff, not a guaranteed large speedup on every compiler. It also means recent A/B percentage deltas that were measured against a denominator inflated by this overhead only re-baseline on toolchains where the overhead was actually present.

Correctness is unchanged: `copyWithin`'s bounds are carried by its Lean-side reference body (`a ++ a.extract srcOff (srcOff + len)`) and the roundtrip conformance sweeps, exactly the model already used by the two siblings that ship these flags. The full test suite passes.

Closes #2811

🤖 Prepared with Claude Code

### DIFF
--- a/lakefile.lean
+++ b/lakefile.lean
@@ -99,7 +99,16 @@ target copy_within_ffi.o pkg : FilePath := do
   let srcJob ← copy_within_ffi.c.fetch
   let oFile := pkg.buildDir / "c" / "copy_within_ffi.o"
   let weakArgs := #["-I", (← getLeanIncludeDir).toString]
-  let hardArgs := if Platform.isWindows then #[] else #["-fPIC"]
+  -- -O2/-DNDEBUG for the same reason as bytearray_wide_ffi: this is a hot LZ77
+  -- match-copy primitive (the non-overlapping back-reference path in
+  -- Inflate.copyLoop). On some toolchains the lean.h static-inline helpers
+  -- (lean_sarray_cptr, lean_is_exclusive, ...) stay outlined at -O0; on others
+  -- (e.g. gcc 15.2) they inline anyway and only the debug-only lean.h asserts
+  -- remain. Either way -O2/-DNDEBUG matches the release runtime and this
+  -- primitive's two siblings. Its bounds are carried by copyWithin's Lean-side
+  -- reference body (+ the roundtrip conformance sweeps).
+  let hardArgs := #["-O2", "-DNDEBUG"] ++
+    if Platform.isWindows then #[] else #["-fPIC"]
   buildO oFile srcJob weakArgs hardArgs "cc"
 
 extern_lib libcopy_within_ffi pkg := do


kim-em:
Closing as superseded. While this branch was in flight, https://github.com/kim-em/lean-zip/pull/2809 ("perf: enable -flto so the wide-load stopgap externs inline into the hot loops") landed on master and, as part of its LTO work, already gave `copy_within_ffi.o` the `-O2 -DNDEBUG` flags this PR was adding (its commit message: "copy_within_ffi.o gains -O2 -DNDEBUG like its siblings"). Master's target now carries `-O2 -DNDEBUG` plus `-flto` and builds via `buildLeanO` (toolchain clang), so merging this PR would only regress it back to `buildO`/system `cc` without `-flto`. Nothing to add here.

kim-em:
Superseded by #2809, which already added -O2 -DNDEBUG to copy_within_ffi.o (plus -flto). See PR comment. (Session: `unknown`)

Issue #2801, opened 2026-07-08
perf(decode): subtables for 12-15-bit codes; retire the walkCanonical fallback


kim-em:
Resolve 12-15-bit Huffman codes through subtables instead of the boxed per-bit `walkCanonical` fallback. Low throughput priority (rare on typical streams) but it closes an adversarial-latency hole and deletes the last boxed decode path.

## Evidence

`walkCanonical` (`Zip/Native/InflateTreeFree.lean:116-134`) resolves codes longer than `fastBits = 11` by a per-length restart loop with boxed-`Nat` `code*2 + bit` arithmetic, `[]!`-checked reads of `firstCode`/`count`/`symbols` per bit, and 4 constructor allocations (a tuple + `Except`) per long code. A 15-bit code costs ~15 boxed iterations + 4 allocations versus libdeflate's single subtable load.

On typical streams the 11-bit root table covers the common codes, so the geomean impact is small — but a valid stream dense in 12-15-bit codes decodes at a crawl, which is DoS-adjacent given this decoder's security posture.

## Work

Adopt libdeflate's subtable scheme (`lib/deflate_decompress.c`, `build_decode_table` ~L858-971): subtables bump-allocated immediately after the root table in the same `Array` (`LITLEN_ENOUGH = 2342` entries worst case), each root entry for a long code carrying the subtable start index + the sub-index bit count, resolved with one extra masked load. Retire `walkCanonical`.

## Priority

Lowest of the decode-perf set for throughput; worth doing for the robustness story and to remove the final boxed-`Nat` decode path. Independent of the hot-loop changes.

🤖 Prepared with Claude Code

kim-em:
## Framing note before implementation

Landing context: the decode campaign has re-based. #2795 showed the pure-Lean per-symbol instruction-count lever is a ~1% floor (decode is latency-bound on the loop-carried recurrence), and #2811 (in flight) fixes an un-optimized match-copy FFI TU that is ~1/3 of decode self time — so **measure any A/B against the post-#2811 baseline**, not today's.

[...]

PR #2819, opened 2026-07-08
perf(decode): subtables for 12-15-bit codes; retire the walkCanonical fallback

kim-em:
This PR resolves 12-15-bit Huffman codes through libdeflate-style subtables instead of the boxed per-bit `walkCanonical` fallback, retiring the last boxed-`Nat` path from the production decode loop. Each long prefix's subtable-block offset is stored **inline in the root fast table's sentinel slot** (`packEntry (off+1) 0` — the length byte stays 0 so the fast path still misses and falls through; the biased block offset rides in the symbol field, 0 meaning "no long code with this prefix"), so there is no separate root-index array and no second `2^fastBits` allocation per block. `buildSubLoop` writes those offsets into the fast table in place and fills the flat `subs` array in one pass over the code lengths (threading `nextCode` exactly like `buildCanonicalLoop`); `subLookup` resolves a long codeword by reading the root entry `decodeSymCanon` already loaded (its symbol field is the block offset) plus one masked load into `subs` — no per-bit `Nat` accumulation. On the headline `silesia_decode_ranking` metric (native decode of libdeflate/zopfli streams) this is a +3.2% geomean win, concentrated in the long-code-dense files whose optimal Huffman coding actually exercises the >`fastBits` path (nci +37%, mr +8%) and flat elsewhere — the per-block subtable build costs the same as the fast-table build alone.

Correctness is a generational refinement holding the equivalence proof at every commit: `subLookup` is proven to return exactly what the boxed `walkCanonical` returns in the context it is called (the root table missed, so any codeword present is longer than `fastBits`), via the `buildSubLoop` fill invariant over the augmented root table (offsets in sentinel slots; non-sentinel slots preserved) and `subLookup_ok_iff_walkCanonical`. The top-level `decodeHuffmanFastBufTreeFree_ok_iff` bridges the augmented table to the tree-built table with a new `goFusedP_table_congr` congruence (the reference loop reads the table only through the fast path, where the two agree), so every downstream correctness theorem — up to `inflate_ok_iff_reference` — transfers unchanged, with no `sorry`. A differential conformance test (`ZipTest.InflateTable.subtableTests`) witnesses the equality at runtime across every code-length regime, including the near-EOF fallback.

🤖 Prepared with Claude Code


kim-em:
## ⚠️ Common-path decode regression — needs a design call before merge (marking draft)

The subtable decode is **fully verified and correct** (sorry-free, `max |Δratio| = 0.000000` — output-neutral, full test suite + 440K/1.54M-sample differential checks pass). But the perf measurement (chungus2, same-worktree before/after, noise floor +0.12%) shows a **uniform decode regression on the common path**, which #2801 explicitly forbids ("don't regress the common path"):

| corpus | metric | AFTER/BEFORE geomean |
|---|---|---|
| Canterbury (L1–8, median-of-5) | decode MB/s | **−3.87%** |
| Silesia (L6, single-pass) | decode MB/s | **−3.19%** |

Per-level Canterbury: −3.4% to −4.6%, uniform. Worst files: silesia/xml −7.2%, silesia/mr −7.1%, canterbury/kennedy.xls −13.6% (symbol-dense → more long codes). Output is byte-identical (correctness intact).

before/after decode

### Root cause
The regression is **not** in the decode hot loop (unchanged) or the cold `subLookup` fallback — it is the per-block subtable **build**. `buildLongDecodeWithCount` now allocates a parallel `rootSub` array of `2^fastBits = 2048` entries **per dynamic block** (to map an 11-bit prefix → subtable-block offset), *on top of* the existing `2^11` root fast-table build — roughly doubling the per-block table-build cost. Since block count ≈ file size, the hit is ~uniform per byte on both corpora (not a small-file artifact). Removing the now-dead `buildSymbols`/`firstIndex` (only the proof-side `walkCanonical` reads them) recovers only ~15% — the `2048`-entry `rootSub` replicate is the dominant cost.

### The fix, and a design call
libdeflate avoids the second array entirely: it stores each long prefix's subtable offset **inline in the root table's sentinel slots** (the slot's length stays 0 → still triggers the fallback; the offset rides in the symbol field). That eliminates the parallel `2^11` build → back to flat. Doing it here means augmenting the root table in place and reworking the ~945-line `buildSubLoop` fill invariant (currently phrased over `rootSub`) to read `table.symAt` instead — a substantial follow-up I did not want to rush at the end of this session.

**Two viable paths — your call:**
1. **Inline-pointer subtables** (this PR's approach, done right): flat common path, true O(1) long-code resolve, faithful to the issue. Larger proof rework.
2. **De-boxed `walkCanonical`**: keep the existing counting-sort structures (no new `2^11` array → zero build regression), just unbox the per-bit accumulation (UInt vs Nat, `Array UInt32`). Satisfies "boxing-free replacement" / "delete the boxed path" with a small, safe proof, but stays O(15) per long code rather than one masked load.

The verified subtable work here is a correct foundation for path 1. Marking this **draft** pending your steer — I did not refresh the committed dashboard (it would record the regression as the new baseline).

kim-em:
## ✅ Reworked to inline offsets (path 1) — regression gone, fully verified

Per your call, the subtable offsets now live **inline in the root fast table's sentinel slots** (`packEntry (off+1) 0` — length stays 0 so the fast path still misses; the biased block offset rides in the symbol field), so there is **no separate `rootSub` array** and no second `2^11` allocation per block. `subLookup` reads the offset from the root entry `decodeSymCanon` already loaded, then does one masked load into `subs`.

**Regression eliminated (output-neutral, `max |Δratio| = 0`):**

| corpus | metric | inline vs master | (was, rootSub design) |
|---|---|---|---|
| Canterbury (L1–8, median-of-5) | decode MB/s | **−0.82%** (noise floor +0.41%) | −3.87% |
| per-level | | **0.98–1.00x** | uniform 0.96x |

Micro-benchmark: the per-block build now costs the *same* as fast-table-only. The residual sub-percent is the irreducible `subs` fill (∝ long codes — the subtable itself has to exist). Before/after now overlap:

inline decode near-flat

**Fully verified — zero `sorry`.** The proof chain was reworked for the inline design: the `buildSubLoop` fill invariant now restates over the augmented root table (offsets in sentinels, non-sentinel slots preserved), and a new `goFusedP_table_congr` (sentinel-symbol-irrelevance) bridges `goFusedP(augmented) = goFusedP(tree table)` at the top level, so `inflate_ok_iff_reference` transfers unchanged. Full library + test suite (incl. the 3.08M-sample differential check) build green.

Refreshing the committed dashboard now. Marking ready for review; happy to squash the WIP history. Merge on your go-ahead.

kim-em:
## Correction: headline Silesia decode is **+3.2%**, not a regression

My earlier perf note measured the wrong things (Canterbury, and decode of lean-zip's *own* shallow-Huffman output — which barely uses the >11-bit path). On the metric we actually report — native decode geomean in `silesia_decode_ranking`, decoding real **libdeflate/zopfli** Silesia streams — this is a win:

**BEFORE (master) 304.0 MB/s → AFTER (inline) 313.7 MB/s = +3.2%** (geomean, 12 files × 6 libdeflate levels).

Concentrated exactly where the feature applies, flat elsewhere:

| file | Δ decode |
|---|---|
| nci | +37.3% |
| mr | +7.9% |
| mozilla / osdb / ooffice | +0.3–0.5% |
| reymont / samba / xml | ~0% |
| dickens / webster / x-ray / sao | −0.3 to −1.1% |

`nci`/`mr` are the real-world Silesia files whose optimal Huffman coding is dense in 12–15-bit codes — the exact path that used to hit the boxed `walkCanonical`. The inline subtable resolves them O(1); the `subs` build cost is negligible against real decode (flat elsewhere).

(Caveat: `nci` decode is high-variance run-to-run and before/after were separate runs, so the exact +37% is approximate — but the direction and mechanism are solid.)

So the net is a genuine common-path decode improvement on real streams, on top of the robustness/boxed-path-deletion goals. Supersedes the Canterbury framing in my earlier comment.

PR #2820, opened 2026-07-08
perf(decode): wide bit-refill via masked ugetUInt64LE in the tree-free loop

kim-em:
This PR adds a wide bit-refill to the production tree-free DEFLATE decoder: when at least eight bytes remain and the bit buffer holds `cnt ≤ 56` bits, the per-byte refill loop is replaced by a single little-endian eight-byte load (`ByteArray.ugetUInt64LE`) whose top `8 − k` bytes are cleared (a `<<< s >>> s` shift pair, `s = 64 − 8k`, `k = (56 − cnt)/8 + 1`) and ORed in at bit `cnt`; near the buffer end it falls back to the byte loop. The masked wide step is proven **state-equal** to the byte loop — same `(pos, bitBuf, cnt)` afterwards — so the whole inflate correctness chain transfers untouched.

The verification is contained in one localized lemma. `refill_eq_wide` (`Zip/Spec/RefillWide.lean`) shows the byte-refill `refill` function itself produces exactly the masked wide step, via `refill_chain` (unroll the `k` byte-loads), `chainOR_eq_recomb` (the `k`-byte OR chain equals the masked recombination, `bv_decide` per `k ∈ {1,…,8}`), and `ugetUInt64LE_recomb`. The production loop `goTreeFreeUWide` is then proven equal to the boxed reference `goTreeFree` (`goTreeFreeUWide_eq`) by reusing the existing `goTreeFree_absorb_refill` machinery: its decode cases are the `goTreeFreeU_eq` cases with an extra "not wide" guard threaded through, and only the wide case is new. The wide guard uses subtraction (`pos ≤ data.size.toUSize − 8`, guarded by `8 ≤ data.size.toUSize`) rather than `pos + 8 ≤ …` so it never wraps near the top of the address space.

The measured in-situ decode win is marginal and content-dependent. Pinned, drift-cancelled interleaved A/B (n=15) over canterbury payloads at level 6: lcet10 +1.5%, dickens +1.3%, ptt5 flat, kennedy.xls −1.0% — geomean ≈ +0.5%. Text-heavy streams gain a little; binary/structured data is flat-to-slightly-negative because the masked variant reads eight bytes even when the byte loop would have topped up with one or two. This matches the issue's own "single-digit / dilutes like the hash3 reader" prediction. The change carries no correctness risk (it is proven equal to the byte loop), and it is the masked foundation the giesen-trick follow-up (#2630's deferred variant) would build on — though the marginal in-situ slice here is a fair argument that the giesen bisimulation rework is not yet justified.

The reference/spec decoder `goFusedPU` (used only by `inflateRawReference`) deliberately stays byte-refill: it is not the production path, so widening it would add complexity to the spec anchor for no runtime win. Codex reviewed the diff and found no soundness issues (out-of-bounds, `k`/mask off-by-one, or `USize` wraparound); the `k = 8` / `cnt = 0` full-word case and the `≤ 64` post-refill invariant check out.

🤖 Prepared with Claude Code

### DIFF
--- a/Zip/Native/Inflate.lean
+++ b/Zip/Native/Inflate.lean
@@ -1020,6 +1020,50 @@ theorem refillGuard_usize (data : ByteArray) (pos cnt : USize) (hsz : data.size 
   rw [USize.le_iff_toNat_le, USize.lt_iff_toNat_lt, toUSize_toNat_of_lt hsz,
       USize.toNat_ofNat_of_lt (Nat.lt_of_lt_of_le (by decide) USize.le_size)]
 
+/-- The wide refill guard read in `USize` (`cnt ≤ 56` and eight bytes remain).
+    The bound is written as `pos ≤ data.size.toUSize - 8` (subtraction, guarded by
+    `8 ≤ data.size.toUSize`) rather than `pos + 8 ≤ …` so it never wraps near the
+    top of the address space, and matches the `Nat` guard `pos.toNat + 8 ≤
+    data.size` once the buffer is `USize`-addressable. -/
+theorem refillGuardWide_usize (data : ByteArray) (pos cnt : USize) (hsz : data.size < USize.size) :
+    (cnt ≤ (56 : USize) ∧ (8 : USize) ≤ data.size.toUSize ∧ pos ≤ data.size.toUSize - 8)
+      ↔ (cnt.toNat ≤ 56 ∧ pos.toNat + 8 ≤ data.size) := by
+  have h56 : (56 : USize).toNat = 56 :=
+    USize.toNat_ofNat_of_lt (Nat.lt_of_lt_of_le (by decide) USize.le_size)
+  have h8 : (8 : USize).toNat = 8 :=
+    USize.toNat_ofNat_of_lt (Nat.lt_of_lt_of_le (by decide) USize.le_size)
+  have hsize : data.size.toUSize.toNat = data.size := toUSize_toNat_of_lt hsz
+  rw [USize.le_iff_toNat_le, USize.le_iff_toNat_le, USize.le_iff_toNat_le, h56, h8, hsize]
+  constructor
+  · rintro ⟨hc, h8le, hp⟩
+    rw [USize.toNat_sub_of_le _ _ (USize.le_iff_toNat_le.mpr (by rw [h8, hsize]; omega)),
+        h8, hsize] at hp
+    exact ⟨hc, by omega⟩
+  · rintro ⟨hc, hp⟩
+    refine ⟨hc, by omega, ?_⟩
+    rw [USize.toNat_sub_of_le _ _ (USize.le_iff_toNat_le.mpr (by rw [h8, hsize]; omega)),
+        h8, hsize]
+    omega
+
+/-- The wide refill loads `k = (56 - cnt)/8 + 1` whole bytes; its `USize`
+    computation round-trips to the `Nat` value under `cnt ≤ 56`. -/
+theorem wideK_toNat (cnt : USize) (hcnt : cnt.toNat ≤ 56) :
+    ((56 - cnt) / 8 + 1 : USize).toNat = (56 - cnt.toNat) / 8 + 1 := by
+  have h56 : (56 : USize).toNat = 56 :=
+    USize.toNat_ofNat_of_lt (Nat.lt_of_lt_of_le (by decide) USize.le_size)
+  have h8 : (8 : USize).toNat = 8 :=
+    USize.toNat_ofNat_of_lt (Nat.lt_of_lt_of_le (by decide) USize.le_size)
+  have h1 : (1 : USize).toNat = 1 := USize.toNat_one
+  have hbig : (64 : Nat) < 2 ^ System.Platform.numBits :=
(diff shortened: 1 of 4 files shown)


kim-em:
## Decode measurement (byte-refill → wide-refill)

Single-decoder driver (`inflate-profile decode`, so only `Zip.Native.Inflate.inflate` runs), one core pinned via `taskset`, **drift-cancelled interleaved A/B** (byte binary from `origin/master`, wide binary from this branch, alternated back-to-back, n=15, geomean of per-pair ratios):

| payload | wide / byte | median | sd |
|---|---|---|---|
| lcet10 (text, 427 KB) | **1.015** (+1.5%) | 1.014 | 0.004 |
| dickens (text, 10 MB) | **1.013** (+1.3%) | 1.012 | 0.007 |
| ptt5 (image, 513 KB) | 1.000 (flat) | 0.999 | 0.007 |
| kennedy.xls (binary, 1 MB) | 0.990 (−1.0%) | 0.993 | 0.011 |
| **geomean** | **≈ 1.005 (+0.5%)** | | |

Honest read: a small, content-dependent win. Text-heavy streams gain ~1.5%; binary/structured data is flat-to-slightly-negative, because the masked variant reads a full eight bytes per refill even where the byte loop would have topped up with one or two. The change is proven state-equal to the byte loop, so there is no correctness risk either way — the question is purely whether the ~0.5% geomean (and the giesen-follow-up foundation) is worth the added loop + proof. I lean toward landing it as a marginal win + foundation, but the mixed in-situ slice is a legitimate argument for recording it as marginal instead.

Absolute throughput on this (shared 96-core) box for reference: lcet10 ~294→299, ptt5 ~438→442, kennedy ~247→247, dickens ~single-decode MB/s in the same band.

kim-em:
## Corrected: rigorous benchmark says decode-neutral (my earlier +2% was noise)

I re-ran this properly — full-corpus (Canterbury + Silesia), both binaries built in the **same worktree** (no stale-`.lake` cross-worktree artifact), a **self-controlled sandwich** (BEFORE→AFTER→BEFORE, 6 reps, pinned to one core), reading `decompress_mbps` with `compress_mbps` (code I did **not** touch) as the same-code control:

| corpus | decode wide/byte | compress control (should be ~0) |
|---|---|---|
| Canterbury | 0.968 (−3.2%) | 0.992 (−0.8%) |
| Silesia | 1.008 (+0.8%) | 1.020 (+2.0%) |

**Same-binary noise floor** (identical BEFORE binary run twice, M2/M1): **+3.2%** — the box is under load (loadavg ~20 during the run). Both decode deltas sit *inside* that floor, and the compress control (unchanged code) drifts ±2%, so the apparent decode movement is machine noise / warmup position bias, not the change.

**Honest verdict: the wide-refill is decode-neutral in situ — no measurable speedup.** My earlier microbench "+2% / +0.5% geomean" (single-decoder `inflate-profile`, cross-worktree) did not clear this noise floor and I should not have presented it as a win. The consistent pattern across every measurement is: text streams a hair faster, binary/structured data a hair slower (the masked variant reads a full 8 bytes even when the byte loop would top up with 1–2), netting to ~0.

This matches the issue's own prediction ("dilutes like the hash3 reader; single-digit at best"). The change remains **verified and correct** (proven state-equal to the byte loop), but it does not move decode throughput measurably, so the in-situ slice does **not** justify the giesen bisimulation follow-up — arguably it doesn't justify landing the masked variant either. I'd defer to Kim on whether to land it as a neutral verified refactor / giesen foundation, or record a neutral verdict and close.

(Not refreshing the committed decode dashboard, since there is no perf delta to record; happy to on a quiet machine if we keep it.)

kim-em:
## Clean decode-only measurement: slight net regression, closing

Re-ran on a decode-only harness (`inflate-profile decode`, no compress-warmup confound), both binaries built in the **same worktree**, self-controlled sandwich (BEFORE→AFTER→BEFORE, 8 reps, pinned to one core). This time the **noise floor is ±0.01%** (the identical BEFORE binary agrees with itself to 0.01%; per-payload sd 0.003–0.01), so the signal is real:

| payload | decode wide / byte |
|---|---|
| sao (incompressible float) | **−6.8%** |
| kennedy.xls (binary) | −1.0% |
| mozilla (binary exe) | −1.0% |
| alice29 (text) | −0.8% |
| dickens (text) | +0.8% |
| webster (text) | +0.8% |
| xml | +1.1% |
| nci | +1.2% |
| **geomean** | **−0.74%** |

The masked wide-refill is a **slight net regression**, worst on literal-heavy / incompressible streams (`sao` −6.8%) where it reads a full eight bytes every refill even though the byte loop would top up with one or two. Text streams are a wash (some +0.8–1.2%, alice −0.8%). There is no in-situ decode win — if anything a small loss.

Closing this PR: the implementation is verified and correct (proven state-equal to the byte loop), but it does not deliver the speedup the issue explored, so there is nothing to land. Recording the negative verdict on #2630.

kim-em:
Closing: the wide-refill is decode-neutral-to-slightly-negative in situ (clean decode-only sandwich, noise floor ±0.01%: geomean −0.74%, worst −6.8% on incompressible sao). Verified and correct, but no speedup to land. Negative verdict recorded on #2630.

PR #2822, opened 2026-07-08
perf: seed the lazy lookahead walk with the current match length

kim-em:
This PR passes the current match length as the initial best length of the lazy `pos+1` lookahead walk (zlib's `prev_length` seeding): below the walk's `niceLen` cutoff, the `scan_end` prefilter can then reject any candidate that cannot beat the pending match with a single byte compare, instead of paying a full `countMatch` extension; at or above the cutoff the walk stays unseeded (a seeded walk would early-stop and diverge). The output is proven byte-identical: `chainWalk_seed` (fuel induction on the reference walk) shows the seeded walk either reproduces the unseeded result or returns the seed itself, `lazyAcceptCost_of_le` shows a probe no longer than the current match is never accepted, and `seeded_probe_bridge` transports both through the packed USize walk into `mergedLoop_eq`, leaving every downstream theorem untouched. No `sorry`.

Measured (chungus2, pinned, interleaved A/B vs master): xml L6 +0.9%, dickens L6 +0.8%, wash at L4 (the `good_match` gate already skips most probes there) and on match-sparse binaries; byte-identical `out=` verified for {xml, dickens, mozilla, samba} × L4-L8. A modest but strictly free speed win; the seeding lemmas are also the proof machinery the planned hash3-singleton walk seeding reuses.

🤖 Prepared with Claude Code

### DIFF
--- a/Zip/Native/Deflate.lean
+++ b/Zip/Native/Deflate.lean
@@ -1920,8 +1920,18 @@ def lz77LazyMergedLoop (data : ByteArray)
             let head2 := headProbeGuarded c (prevSize + h2)
             let maxLen2 := min 258 (data.size - (pos + 1))
             have hmaxLen2P : (pos + 1) + maxLen2 ≤ data.size := by omega
+            -- Seed the lookahead walk with the current match length (zlib
+            -- `deflate_slow`'s `prev_length` trick): a candidate shorter than the
+            -- `pos` match cannot improve on it, so pre-loading `matchLen` as the
+            -- probe's best length lets `scan_end` skip those candidates' full
+            -- compares. The seed is only applied below the walk's own cutoff
+            -- (`min niceLen maxLen2`); at or above it a seeded walk would early-stop
+            -- immediately and diverge, so those (rare, long-match) probes stay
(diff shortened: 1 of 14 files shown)


kim-em:
Before/after `compress_mbps` (machine `Linux chungus2 x86_64`, pinned; BEFORE = master's `latest.json` @ post-#2823 refresh):

silesia compress pareto

canterbury compress pareto

`max |Δratio| = 0.000000` — output-neutral, as the seeding lemma proves. The overlay reads 0.99–1.00x at every level: this change's magnitude (+0.8–0.9% on match-dense text at the lazy tier, per the pinned *interleaved* A/B in the PR description) is **below the single-pass cross-session noise floor**, so per the perf-graphs protocol the overlay is not evidence either way — the interleaved A/B is. The change is strictly free by construction (the seeded walk provably does a subset of the unseeded walk's byte compares), and its `chainWalk_seed` machinery is load-bearing for the upcoming hash3-singleton PR, which seeds the same walk with verified length-3 candidates.

Dashboard refreshed with the AFTER numbers in this PR (40f06021).

PR #2821, opened 2026-07-08
perf: USize-native merged insertion loop (updateHashesMergedFastU)

kim-em:
This PR de-boxes the lazy matcher's interior-hash insertion loop: `updateHashesMergedFastU` runs the loop counter, insertion cap, data-bound test, bucket index, and chain mask on unboxed `USize` (with the addressability witness and `data.size - pos - 2` bound hoisted to a once-per-match guard), and `hash3U` hoists `hash3`'s per-call `USize` round-trip and position conversion to the caller. Writes and the head read go through proven-bounds `uset`/`uget`. This is the Wave 7 P1b treatment (`chainWalkPackedU`, #2755) applied to the other hot matcher loop, which profiles at 13-25% of L4-L8 compress time (Silesia dickens/xml, perf LBR). Output is byte-identical: `updateHashesMergedFastU_eq` proves the USize walk in lockstep against `updateHashesMerged`, and the rewired `updateHashesMergedGuarded` dispatches to it under a faithfulness guard whose fallback is the existing Nat walk, so `updateHashesMergedGuarded_eq` and every downstream proof are unchanged. No `sorry`.

Measured (chungus2, `bench/pin_core.sh`, `compress-pareto`, interleaved A/B vs master, spread < 1%): xml L4 +17.9% (89.1 to 105.1 MB/s), xml L6 +12.5%, mozilla L6 +4.5%, dickens L6 +4.2%, dickens L8 +3.2%; byte-identical `out=` verified across silesia xml/dickens/reymont/ooffice/mr at L4-L8.

🤖 Prepared with Claude Code

PR #2823, opened 2026-07-08
perf: array-native two-queue Huffman tree build

kim-em:
This PR replaces the per-block Huffman tree construction — still `List.mergeSort` + the O(n²) `insertByWeight` spec build after #2769 array-ified only the assignment step — with a van Leeuwen two-queue build over flat arrays (sorted leaf queue + merged-node queue, parent pointers upward, one reverse pass for depths): O(n) merging, no per-node allocation. On stat-shifting binaries the observation-divergence splitter produces hundreds of blocks, each paying two of the old builds; `insertByWeight` was the single largest profile entry on Silesia x-ray at L6 (16.5% self).

The new build is a heuristic replacement, not a byte-identical one: every property the roundtrip proof consumes (result size, lengths in [1,15], Kraft validity via `fixKraftList`, nonzero lengths for used symbols) depends only on the `symsByFreq.zip (expandBl bl ++ padding)` output shape, which is unchanged — those lemmas are re-proven for `limitedPairsN` on the shared infrastructure, the spec pipeline and all its theorems stay untouched, and tree optimality remains a ratio concern validated by the corpus: the corpus-weighted Silesia delta is at most +0.0026% (L6 +1797 bytes of 68.3 MB; L4/L5 byte-identical, dynamic path untriggered). No `sorry`.

Measured (chungus2, pinned, best-of-5 interleaved A/B vs master): x-ray L6 +21.6% (17.6 to 21.5 MB/s), x-ray L7 +20.7%, mozilla L6 +9.6%, osdb L6 +5.4%, sao L6 +4.4%, dickens L6 +4.3%, xml L6 +2.7%. `lake exe test` green including native/FFI roundtrip conformance.

🤖 Prepared with Claude Code

### DIFF
--- a/Zip/Native/HuffmanEncode.lean
+++ b/Zip/Native/HuffmanEncode.lean
@@ -12,15 +12,33 @@ members) no longer pays the `O(n²)` `List.set`/`List.length` churn that
 `assignLengths` incurs, nor the allocator/reference-count traffic of the
 `List Nat` result the emitters immediately `toArray` anyway.
 
-The correspondence theorems (`computeCodeLengthsN_toList`,
-`computeCodeLengthsN_toArray`) prove the array pipeline yields exactly the spec
-`List`, so every downstream property of `computeCodeLengths` (length,
-`ValidLengths`, completeness, nonzero) transfers verbatim and the compressed
-output is byte-identical.
+The array pipeline uses an array-native **two-queue (van Leeuwen) Huffman tree
+build** (`huffTreeDepthsN`) in place of the spec's `List.mergeSort` +
+`insertByWeight` construction. The spec built the tree with `O(n²)` boxed list
+insertion and a `BuildTree` allocation per merge; on stat-shifting corpora the
+observation-divergence splitter cuts each member into hundreds of blocks, each
+paying two tree builds (~286-symbol lit + 30-symbol dist), so `insertByWeight`
+was the top compress profile entry. The two-queue build is `O(n)` over a sorted
+leaf array with no tree allocation: merged nodes are produced in non-decreasing
+weight order, so a second array + two head indices suffice, and leaf depths are
+recovered by a single reverse pass over parent pointers.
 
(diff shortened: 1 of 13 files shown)


kim-em:
Before/after `compress_mbps` (machine `Linux chungus2 x86_64`, pinned single core; BEFORE = master's committed `latest.json` @ post-#2821 refresh):

silesia compress pareto

canterbury compress pareto

Silesia by level, geomean and (corpus-weighted = total bytes / total time — the metric the mixing frontier uses):

| lvl | geomean before → after | weighted before → after |
|---|---|---|
| 1-5 | 1.00x | −0.2…−0.6% (drift; untouched-path control) |
| 6 | 36.0 → 36.8 (+2%) | 35.95 → **36.99 (+2.9%)** |
| 7 | 30.1 → 30.7 (+2%) | 29.70 → **30.46 (+2.5%)** |
| 8 | 27.9 → 28.5 (+2%) | 27.00 → **27.60 (+2.2%)** |
| 9 | +1% | +0.9% |
| 10 | +2% | +2.2% |

The geomean dilutes what is a file-concentrated win: the per-file interleaved A/B (PR description) has x-ray L6 +21.6%, x-ray L7 +20.7%, mozilla L6 +9.6% — the stat-shifting members whose hundreds of split blocks each paid two O(n²) list tree builds. Controls (L1–L5, where the timing drift shows −0.2…−0.6%) suggest the true L6–L8 weighted gain is nearer +3%. `max |Δratio| = 0.0005`, a single small Canterbury file at L3; the corpus-weighted Silesia ratio delta is ≤ +0.0026% as measured in the PR description — the two-queue tree differs from the insertion tree only on weight ties.

Caveat: Silesia is single-pass, and a concurrent Lean build (another agent's proof work) ran during part of this measurement; the per-file A/B in the PR description was quieter and is the better magnitude estimate for x-ray/mozilla.

Dashboard refreshed with the AFTER numbers in this PR (b257dfbb).

PR #2824, opened 2026-07-08
perf: hash3-singleton length-3 coverage at the split tier (L6-L8)

kim-em:
This PR adds libdeflate's hash3-singleton to the split-tier lazy matcher (levels 6-8): a 32768-entry most-recent-position table keyed by the 15-bit 3-byte hash, maintained fused inside the USize insertion loop (one extra proven-bounds `uset` per inserted position, sharing the hash4 wide load), and consulted once per match-search position — a verified in-window length-3 candidate within `TOO_FAR = 4096` seeds the chain walk, restoring the length-3 matches a hash4-keyed chain cannot see. The lazy `pos+1` probe stays unseeded (a length-3 seed there is output-neutral: deferral acceptance requires strict improvement over a match that is already ≥ 3). Levels outside 6-8 are proven byte-identical (`useH3 = false` collapses the dispatcher to the existing loop); L9/L10 keep their own cache-build singleton.

This reopens #2742's plan item 1, which was measured and rejected on *unweighted geomean* — but the gain is concentrated exactly where the corpus bytes are: on weighted Silesia the L6 ratio moves 0.32236 → 0.31959 (−0.28 pp; x-ray −235 KB closing most of its gap to the hash3-capable codecs, mozilla −340 KB, sao −134 KB, ooffice −77 KB now *beating* miniz_oxide; match-dense text pays back tens of KB), and the same at L7/L8. The mixing-frontier test is decisive: the new L6 point (ratio 0.3196 at ~34 MB/s weighted) sits at a ratio today's frontier only reaches at ~14 MB/s via an L8↔L9 blend — far outside the hull — and it pulls the achievable frontier at miniz_oxide L6's ratio (0.3213) from ~28 to ~36 MB/s, putting that reference point inside our hull for the first time. Speed cost where gated on is file-dependent (mozilla −6%, dickens −11%, xml −14%, x-ray +34% — the length-3 references shrink its token stream), a price the frontier judgment already absorbs; gated-off levels are a measured wash.

The probe-seeded walk carries new obligations (`hash3Probe_spec`, `h3Seed_spec`, seed-general packed-walk decode lemmas) threaded through the boxed/packed/merged lockstep proofs; the emitted length-3 references are re-verified at emission as always, `lake exe test` (roundtrip + FFI conformance) is green, and there is no `sorry`. The production implementation reproduces the certified bench spike's out-bytes exactly on all 36 Silesia file×level rows.

🤖 Prepared with Claude Code

### DIFF
--- a/Zip/Native/Deflate.lean
+++ b/Zip/Native/Deflate.lean
@@ -886,6 +886,46 @@ walk is orders of magnitude deeper per position. -/
     else 0
   else 0
 
+/-- Reject length-3 singleton seeds farther than this (zlib `deflate.c`
+    `TOO_FAR`): a length-3 match at a large distance costs more bits than three
+    literals. The split-tier singleton probe caps its seed distance at
+    `min windowSize tooFar3`, exactly as the L9/L10 cache build already does;
+    measured optimal at 4096 (#2742 / the split-tier sweep). -/
+def tooFar3 : Nat := 4096
+
+/-- The hash3-singleton chain-walk seed at `pos`: when `useH3` is set, read the
+    most-recent length-3 candidate from `h3tab` and `hash3Probe`-verify it within
+    `min windowSize tooFar3`, returning the packed `cand3*512+3` on a hit (else
+    `0`); when `useH3` is clear, the empty seed `0`. The result is *always* a real
+    in-window backward match or empty (`h3Seed_spec`), so the seeded chain walk
+    may return it and the loop emit it as a verified length-3 reference — this is
+    how the split tier (L6–L8) sees the length-3 matches its hash4 chains miss.
+    The `h3tab` contents are pure heuristic state: the probe re-verifies the
+    bytes and the window, so they never enter a validity proof. -/
+@[inline] def h3Seed (useH3 : Bool) (data : ByteArray) (h3tab : Array Nat)
+    (windowSize pos : Nat) (hlt : pos + 2 < data.size) : Nat :=
+  if useH3 then
+    hash3Probe data (min windowSize tooFar3) pos (headProbeGuarded h3tab (hash3Single data pos hlt)) hlt
+  else 0
+
+/-- Interior-position hash3-singleton insertion, the length-3 twin of
+    `lz77Chain.updateHashes`' chain insertion: for each `j ∈ [j, matchLen)` (also
+    `≤ insertCap`) with three readable bytes, store `pos+j` into
+    `h3tab[hash3Single data (pos+j)]`. Runs at exactly the interior positions the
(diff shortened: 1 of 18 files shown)


kim-em:
Before/after `compress_mbps` (machine `Linux chungus2 x86_64`, pinned; BEFORE = master's `latest.json` @ post-#2822 refresh; this is a **parse change** — the ratio moves on purpose):

silesia compress pareto

canterbury compress pareto

Weighted Silesia (total bytes / total time — the frontier metric):

| lvl | before | after |
|---|---|---|
| 6 | 0.3224 @ 36.7 MB/s | **0.3196 @ 33.7 MB/s** |
| 7 | 0.3211 @ 30.6 | **0.3184 @ 27.9** |
| 8 | 0.3209 @ 28.2 | **0.3181 @ 25.4** |
| 1–5, 9, 10 | byte-identical, speed 0.99–1.01x (controls clean) | |

`max |Δratio| = 0.0278` at silesia/x-ray L6 (0.7401 → 0.7123) — the intended direction; sao 0.7544 → 0.7359, ooffice 0.5136 → 0.5011, mozilla −340 KB; match-dense text pays back tens of KB (dickens +20 KB), netting **−0.28 pp corpus-weighted at each of L6/L7/L8**.

**Mixing-frontier verdict** (the achievable-blend test, computed in sec/MB): the new L6 point sits at a ratio the old frontier reached only at ~14 MB/s (an L8↔L9 blend) — it is ~2.4× outside. Against the reference codecs, the frontier at each competitor's own (ratio, speed):

| reference point | frontier speed at its ratio | verdict |
|---|---|---|
| miniz_oxide L6 (0.3213 @ 35.5) | 35.5 | **on the line** (was −22%) |
| miniz_oxide L7 (0.3205 @ 27.8) | 34.6 | **inside, +24%** |
| miniz_oxide L8 (0.3201 @ 22.4) | 34.2 | **inside, +53%** |
| miniz_oxide L9 (0.3199 @ 20.3) | 34.0 | **inside, +67%** |
| zlib L7 (0.3205 @ 31.2) | 34.7 | **inside, +11%** |
| zlib L8 (0.3194 @ 20.0) | 32.4 | **inside, +62%** |

Four miniz_oxide points and two zlib points move inside our hull in one change; the miniz L6 headline point lands exactly on the boundary in this run (single-pass Silesia, ±1–2%) — a follow-up knob re-grid on the singleton's new cost structure should push it clearly inside. Canterbury pays more than Silesia (0.89–0.91x at L6–L8, small files with few barely-compressible members); per the perf protocol Silesia carries the merge decision.

Dashboard refreshed with the AFTER numbers in this PR (5d41a41c).

PR #2827, opened 2026-07-09
perf: merged-array USize greedy matcher loop (levels 1-3)

kim-em:
This PR port the merged-array structural stack from the lazy tier to the greedy matcher (levels 1-3): `lz77ChainIterPMerged` runs the greedy main loop on one combined `Array Nat` (prev ring at offset 0, hash4 table above), with interior insertions through the existing USize merged insertion walk — so no `(hashTable, prev)` pair is allocated anywhere on the greedy hot path (confirmed in the generated C: zero `lean_alloc_ctor` in the loop body). The greedy tier never received the #2767/#2821 treatment and its outer loop profiled at 27-36% self time at L1-L3 plus 3-4% RC churn. Output is byte-identical: `greedyMergedLoop_eq` proves lockstep equality with `lz77ChainIterP` (the `mergedLoop_eq` shape minus the lazy branch and h3 seed), and the boxed/two-array functions stay as spec anchors. No `sorry`.

Measured (chungus2, pinned, interleaved medians of 5, zero sign flips across 25 pairs): dickens L1 +5.6%, mozilla L1 +2.6%, xml L1 +4.8%, dickens L2 +9.6%, mozilla L3 +7.9%. L1 gains less than L2/L3 because `insertCap = 2` leaves little interior-insertion churn to delete there; L4+ dispatch is untouched.

🤖 Prepared with Claude Code

PR #2829, opened 2026-07-09
perf: pack and pre-reverse the dynamic-emit Huffman tables; scalar BitWriter flush kernel

kim-em:
This PR cut the dynamic-emit path's per-token cost (24-27% of L1 compress time) in three output-byte-identical steps. The Huffman code tables were `Array (UInt16 × UInt8)` — a boxed `Prod` chase per token — and `writeHuffCode` re-reversed the code's bits on every symbol even though the reversed form is a per-block constant: `packCodeTab` now packs each entry as one `UInt32` holding the pre-reversed code plus length (one scalar array read per token), written through the new `BitWriter.writeRevCode`. The `writeBits`/`writeHuffCode` flush bookkeeping also moves from tagged-`Nat` arithmetic (`nat_add`/`nat_mod`/recursive `dropBytes` per flush) to UInt32/UInt64 register ops, with the `_def` equations' statements unchanged so no downstream spec proof moves. Byte-identity is machine-checked end-to-end (`writeRevCode_eq`, `packCodeEntry_write`, `emitTokensWithCodesPT_eq`); no `sorry`.

Measured (chungus2, pinned, interleaved medians of 7, from-scratch baseline after ruling out a stale-cache layout phantom — see below): dickens L1 +6.1%, mozilla L1 +6.2%, xml L1 +5.7%, dickens L2 +4.2%, xml L6 +0.4%; the emit share of dickens L1 drops 26.2% → 20.1%. Two measurement hazards found on the way, worth their own note: inlining a branch into the hot FBIP flush arm made the ctor-reuse pass pick the wrong reuse site (−10-14% — verify `del_object` placement in the generated C after touching hot writers), and a baseline linked from stale `.lake` objects carried a luckier GCC inline layout for the untouched matcher (check `nm` for hot-symbol presence before trusting an A/B).

🤖 Prepared with Claude Code

### DIFF
--- a/Zip/Native/BitWriter.lean
+++ b/Zip/Native/BitWriter.lean
@@ -1,3 +1,4 @@
+import Std.Tactic.BVDecide
 import Zip.Native.Wide
 
 /-!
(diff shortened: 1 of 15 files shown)


kim-em:
Before/after `compress_mbps` (machine `Linux chungus2 x86_64`, pinned; BEFORE = master's `latest.json` @ post-#2827 refresh; `max |Δratio| = 0.000000` — output-neutral as proven):

silesia compress pareto

canterbury compress pareto

Silesia geomean (Canterbury in parentheses):

| lvl | before | after | speedup |
|---|---|---|---|
| 1 | 95.3 | 103.0 | **1.08x** (1.06x) |
| 2 | 76.1 | 81.7 | **1.07x** (1.05x) |
| 3 | 69.0 | 73.4 | **1.06x** (1.05x) |
| 4 | 51.8 | 54.3 | 1.05x (1.04x) |
| 5–8 | — | — | 1.02–1.04x |
| 9–10 | — | — | 1.00–1.02x |

The BitWriter kernel + packed tables help every level that emits dynamic blocks, so the whole curve lifts; part of the L4–L8 movement is also recovering the ~1% control depression the previous refresh carried (that AFTER was built in a fresh worktree; this one in a long-lived worktree — the two dashboards bracket the true baseline). The interleaved same-binary-protocol A/B in the PR description (L1 +6%, L2 +4%) is the conservative magnitude estimate. Output-neutral, so the L1 endpoint moves straight up — outside the extrapolated L1–L2 line by construction.

Weighted-Silesia L1 now reads 103 MB/s — native has passed the JS fflate curve across the board and is closing on Go/zlib in the fast band.

Dashboard refreshed with the AFTER numbers in this PR (96cad233).

PR #2830, opened 2026-07-10
perf: re-grid L2/L3 for the merged-greedy + packed-emit cost structure

kim-em:
This PR re-grid the greedy-tier L2/L3 knobs after #2827 (merged greedy loop) and #2829 (packed emit tables) shifted the tier's cost balance. The `l1-sweep2` grid (chain × insertCap × niceLen through the exact production fast-tier dispatch, all 12 Silesia files, pinned interleaved timing; certification rows byte-exact against production at L1/L2/L3) showed both production mid rows sitting ~10% below the greedy-band mixing frontier, with the libdeflate `niceLen` 10/14 cutoffs (#2744) as the culprit — they cost ~0.5pp weighted ratio for speed better bought by shallowing the chain. L2 becomes (chain 8, insertCap 8, cutoff off): equal weighted ratio (0.34543 vs 0.34545) at **+12.0% speed**. L3 becomes (chain 16, insertCap 32, cutoff off): **better ratio** (0.34028 vs 0.34171) at **+5.3% speed**. Both are strict dominations of the rows they replace, the ladder stays ratio- and speed-monotone, and L1/L4+ are byte-identical. Worst per-file regressions: reymont +2.2% (L2) / +1.3% (L3), against xml −5.7% / nci −4.3% wins. The same sweep's round 1 settled the L1 endpoint question: no weaker-matcher corner comes within 40% of the extrapolated L1–L2 line (the head-probe-only corner reaches 111 MB/s where the line demands ~340), so L1's knobs are confirmed and its movement comes from the output-neutral structural PRs. All knobs are emission-re-verified heuristics; no proof changes; 24/24 out-bytes verified against the sweep's measured rows; `lake exe test` green.

🤖 Prepared with Claude Code

### DIFF
--- a/Zip/Native/DeflateDynamic.lean
+++ b/Zip/Native/DeflateDynamic.lean
@@ -662,6 +662,13 @@ attribute [irreducible] symbolBitCount fixedBlockBytes dynBlockBytes dynBlockByt
     so the deep-chain point fell inside the new frontier and L7 adopts the old
     L6 config wholesale. L8 keeps 512; past that the chain saturates.
 
+    **L2/L3 dropped to 8/16 in the greedy re-grid** (`l1-sweep2`, run after the
+    merged-greedy-loop and packed-emit landings shifted the tier's cost
+    balance): with the `niceLen` cutoff disabled (see there), (chain 8, cap 8)
+    matches the old L2's weighted-Silesia ratio exactly at +12% speed, and
+    (chain 16, cap 32) beats the old L3 on both axes — the old rows sat ~10%
+    below the greedy-band mixing frontier.
+
     Level 1 is the `deflate_fast` corner (#2726): depth `4` is exactly zlib
     `-1`'s `max_chain`. A tokens-held-constant attribution on Silesia (see
     `ZipL1Attrib`/`ZipL1Sweep`) showed L1 is emit-bound — the token walk +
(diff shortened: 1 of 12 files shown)


kim-em:
Before/after `compress_mbps` (machine `Linux chungus2 x86_64`, pinned; BEFORE = master's `latest.json` @ post-#2829 refresh; the L2/L3 ratio moves by design — knob change; `max |Δratio| = 0.009` at silesia/xml L2, weighted totals per the table):

silesia compress pareto

canterbury compress pareto

Weighted Silesia:

| lvl | before | after |
|---|---|---|
| 2 | 0.3455 @ 81.7 MB/s | **0.3454 @ 93.4 MB/s (+11%)** |
| 3 | 0.3417 @ 73.4 | **0.3403 @ 81.6 (+7%)** |
| 1, 4–10 | byte-identical, 0.98–1.00x controls | |

**Reference-point standing after this PR** (mixing-frontier test): zlib L3 moves **inside** our hull (+2.2%) — the first zlib fast-band point to fall — and the entire JS fflate curve L1–L4 is now inside (+31/+23/+6/+1%); go L3 is within 6%, go/zlib L2 within 13–14%.

**Campaign scoreboard** (this PR + #2827 + #2829, weighted Silesia vs the campaign-start dashboard): L1 91.5 → 103.4 (+13%, byte-identical output, hence outside the extrapolated L1–L2 endpoint line by construction), L2 69.0 → 93.4 (+35% at equal ratio), L3 62.7 → 81.6 (+30% at better ratio), plus L4–L8 +2–5% from the shared emit-path work.

Dashboard refreshed with the AFTER numbers in this PR (e5f1f3c7).

PR #2833, opened 2026-07-11
perf: fuse token-frequency counting into the L1-L3 matcher pass

kim-em:
This PR fuse the whole-stream token-frequency counting into the greedy matcher pass: `lz77ChainIterPMergedF` produces the packed tokens AND the two `tokenFreqsP` histograms in one walk (bumping at each push site with the same helpers, EOB pre-count included), and levels 1-3 dispatch through `deflateRawBaseF`/`deflateRawBasePF`, which size and emit the base candidate from the fused frequencies — eliminating the separate whole-stream `tokenFreqsP` re-read that followed the matcher (8.5% of L1 self time on dickens, plus the cache traffic of streaming a token array that reaches ~120 MB on mozilla back through the hierarchy). The fused histograms are proven equal to `tokenFreqsP` of the produced tokens, so every downstream sizing/emission proof rewrites through the existing anchors; output is byte-identical (15/15 file×level csize checks, non-empty-verified, plus the equality theorem). Levels 4+ keep the split-tier per-block walks, which the whole-stream fusion cannot serve. No `sorry`.

Measured (chungus2, same-worktree binaries, pinned 5-rep sandwich A/B, same-binary noise floors 0.15-0.22%, all 15 effects positive): mozilla L1 **+6.3%**, dickens L1 **+3.8%**, dickens L2 +1.4%. The Stage-0 isolated-walk probe predicted only 0.6-2.4% — the production win is larger because the fusion also removes the array re-read's cache pollution, which the isolated probe could not see; both probe and production A/B are reproducible via the committed `bench/ZipFreqFusion.lean`.

🤖 Prepared with Claude Code

### DIFF
--- a/Zip.lean
+++ b/Zip.lean
@@ -43,6 +43,8 @@ import Zip.Native.BitWriter
 import Zip.Native.Deflate
 import Zip.Native.DeflateDynamic
 import Zip.Native.DeflateFreqs
+import Zip.Native.DeflateFreqsFused
+import Zip.Spec.DeflateFreqsFusedCorrect
 import Zip.Native.DeflateParse
 import Zip.Spec.LZ77NativeCorrect
 import Zip.Spec.LZ77OptimalCorrect
--- a/Zip/Native/DeflateDynamic.lean
+++ b/Zip/Native/DeflateDynamic.lean
@@ -1,5 +1,7 @@
 import Zip.Native.Deflate
 import Zip.Native.DeflateFreqs
+import Zip.Native.DeflateFreqsFused
+import Zip.Spec.DeflateFreqsFusedCorrect
 import Zip.Native.DeflateParse
 import Zip.Spec.DeflateEncodeDynamic
 import Zip.Spec.DeflateStoredCorrect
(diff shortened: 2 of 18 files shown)


kim-em:
Before/after `compress_mbps` (machine `Linux chungus2 x86_64`, pinned; BEFORE = master's `latest.json` @ post-#2830 refresh; `max |Δratio| = 0.000000` — output-neutral as proven):

silesia compress pareto

canterbury compress pareto

| lvl | Canterbury | Silesia |
|---|---|---|
| 1 | **1.04x** | 1.12x |
| 2 | **1.03x** | 1.09x |
| 3 | **1.03x** | 1.08x |
| 4–10 (untouched) | **1.00x flat** | 1.02–1.09x |

**Binary-health note, read the Silesia column with care:** Canterbury's median-of-5 controls are perfectly flat, but Silesia's untouched levels float +2–9% — the recurring cross-worktree layout bracket between dashboard refreshes (single-pass Silesia, AFTER built in this PR's worktree, BEFORE recorded from another). The authoritative magnitudes are the same-worktree 5-rep sandwich in the PR description (noise floors 0.15–0.22%, all 15 effects positive): **mozilla L1 +6.3%, dickens L1 +3.8%, dickens L2 +1.4%** — attribute roughly that much of Silesia's L1–L3 movement to this PR and the rest to the bracket.

With this landing, weighted-Silesia L1 records ~109 MB/s: native L1 now sits essentially on zlib L2's point (0.354 @ ~118) from the better-ratio side, and the remaining gap to zlib L1 itself is ~17%.

Dashboard refreshed with the AFTER numbers in this PR (74cdd346).

PR #2835, opened 2026-07-11
perf: re-grid L5 onto the post-singleton frontier (light split tier)

kim-em:
This PR re-position L5, which had become a dead vertex: the recent landings moved L4 and L6 but left L5 (single-block, chain 128) sitting 15.6% inside the L4↔L6 mixing line on weighted Silesia. The gate-sweep grid (extended with a single-block mode; certification row byte-identical to production L5 on all 12 files) shows the whole single-block family is dead in this band and the frontier peaks one step past the original grid edge at a light split config: L5 becomes (chain 24, goodMatch 64, niceLen 65, lazyDepth /4, split tier on, singleton off), moving the split-tier boundary in `deflateRaw` from `6 ≤ level` to `5 ≤ level` (`useH3Level` stays 6-8; the roundtrip proofs case-split symbolically on the dispatch, so the boundary move is proof-transparent). Weighted Silesia: 0.3302 @ 46.4 → **0.3249 @ 46.2** — −0.53pp ratio at equal corpus speed, **+4.05% above the L4↔L6 blend** at its own ratio (threshold ≥3%; confirmed under both dashboard-scaled and in-harness anchors). Per-file, speed redistributes: few-cut files get much faster (dickens L5 +29%), match-rich heterogeneous ones trade ~20% speed for 5-7% smaller output (mozilla −4.74% size); worst ratio regressions nci +3.6%, reymont +2.8% against samba −6.7%, mozilla −4.7% wins. Levels 1-4 and 6-10 are sha256-identical to master on all 12 files (108/108); `lake exe test` green.

🤖 Prepared with Claude Code

### DIFF
--- a/Zip/Native/DeflateDynamic.lean
+++ b/Zip/Native/DeflateDynamic.lean
@@ -651,10 +651,16 @@ attribute [irreducible] symbolBitCount fixedBlockBytes dynBlockBytes dynBlockByt
 
     The mid-band values are the `mid-sweep` optimum for the #2737 ladder
     (Silesia grid over chain × `goodMatch` × split, then interleaved pinned
-    timing): L4 = 64 (with the lazy gate, the old L5 point), L5 = 128 (gate
-    off — `goodMatch` buys more ratio per cycle than deepening the chain, so
-    L5 = (128, gate off) dominates the old L7 = (256, gate off)). **L6's depth
-    drops back to 64 on purpose**: the split tier starts there, and at equal
+    timing): L4 = 64 (with the lazy gate, the old L5 point). **L5 = 24 since
+    the L5 re-grid** (`gate-sweep`, run after the hash3 singleton #2824, gm/ld
+    re-grid #2825, and greedy re-grid #2830 landings): the old L5 = (128,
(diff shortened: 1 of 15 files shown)


kim-em:
Before/after `compress_mbps` (machine `Linux chungus2 x86_64`, pinned; BEFORE = master's `latest.json`; ratio moves at L5 only, by design — `max |Δratio| = 0.0187` at silesia/mozilla L5, the intended direction):

silesia compress pareto

canterbury compress pareto

Weighted Silesia: L5 goes **0.3302 @ 44.2 → 0.3249 @ 44.1** (−0.53pp at equal speed); every other level reads 0.98–1.01x (controls clean, byte-identical by sha256 on all 12 files × 9 levels).

**Frontier verdict:** L5 was 15.6% *inside* the L4↔L6 blend; the full-matrix check now lists it as a frontier **vertex** (+4.05% above the blend in the sweep's paired test). One knock-on to record: with L5 moved left, **L4 (0.3326 @ 56.8) now sits ~2.3% inside the new L3↔L5 blend** — the sag shifted one rung down, marginal enough to be near noise but worth its own look in a future re-grid pass.

Dashboard refreshed with the AFTER numbers in this PR (21deb0b5).

PR #2836, opened 2026-07-11
perf: fuse each Huffman-code + extra-bits pair into one BitWriter write

kim-em:
This PR fuse the length-code and distance-code writes with their extra-bits writes in the packed emit loop: `BitWriter.writeRevCodeExtra` merges both fields into the accumulator with one flush comparison and one constructor on the no-flush fast path, falling back to the exact two-call `writeRevCode`/`writeBits` composition when the pair crosses a flush boundary — the case #2829's analysis flagged as not value-equal under naive fusion. `writeRevCodeExtra_eq` proves the fused writer equal to the composition under `extraCount < 256` (genuinely required: only the byte-field bound keeps the `UInt32` fast-path guard faithful against Nat aliasing; every DEFLATE extra count is a byte field, discharged by `codeExtra_lt_256`), and `emitRefWithCodesPT_eq` unfuses through it, so output is byte-identical (15/15 non-empty csize checks) and no downstream proof moves. No `sorry`.

Measured (chungus2, same-worktree binaries, pinned 5-rep sandwich, same-binary noise floor ±0.3%): dickens L1 +2.3%, dickens L2 +2.3%, xml L1 +1.8%, mozilla L1 +1.8%. A companion USize emit-loop-index conversion was built, measured a wash (best +0.76%, three of four inside the noise floor — the accumulator merges dominate per-token cost, not index bookkeeping), and reverted; recorded here as the measured negative verdict.

🤖 Prepared with Claude Code


kim-em:
Second commit added: the USize emit-loop index, **reversing the wash verdict in the PR description** at Kim's request via a stronger instrument. Rationale: the per-file sandwich (±0.3% floor, best +0.76%) under-resolved it; a paired 6-rep A/B on a single 212 MB `silesia.tar` stream at L1 reads **+1.53% mean, 95% CI [+0.77%, +2.29%]**, every pair positive (117.1–118.2 vs 114.6–117.0 MB/s), byte-identical output on the tar. The whole-stream shape (a ~200M-token array walked once) is where the index bookkeeping actually shows. Equality proofs (`emitTokensWithCodesPTU_eq`/`PTG_eq`) included; tests green; no `sorry`.

kim-em:
Before/after `compress_mbps` (machine `Linux chungus2 x86_64`, pinned; BEFORE = master's `latest.json` @ post-#2835 refresh; `max |Δratio| = 0.000000` — output-neutral as proven, both commits):

silesia compress pareto

canterbury compress pareto

Silesia: L1 1.02x, L2 1.02x, L3 1.01x, L4 1.01x, L5–L10 1.00x (controls clean). The authoritative magnitudes are the same-worktree sandwiches: fused writer +1.8–2.3% (per-file, floors ±0.3%), USize index **+1.53% CI [+0.77%, +2.29%]** on the whole `silesia.tar` single stream — the matrix's per-file shape blends these to the smaller numbers shown. Headline single-stream number: `zip silesia.tar` at L1 runs ~117–118 MB/s with this PR.

Dashboard + the pareto-history animation (new `animation-sync` check) refreshed in-PR (c76a388b).

Issue #2640, opened 2026-06-17
perf(explore): drop the dead L9 candidates — optimal-parse wins outright, base/SC/shared cost ~24% for nothing


kim-em:
**Status: exploration — measured strong, but needs a full-corpus byte-identical gate before landing. Do not claim until triaged.**

## Context
At level 9, `deflateRaw` (`Zip/Native/DeflateDynamic.lean:994`) computes **four** candidate streams and emits the smallest via `pickSmaller`:
1. **base** — greedy/lazy single block (`deflateRawBaseP`; consumes the shared `lzMatchP` pass)
2. **SC** — self-contained per-chunk split, fresh 16 KiB window (`deflateDynamicBlocksSC`; its *own* independent match pass)
3. **shared** — cross-block shared-window split (`deflateDynamicBlocksSharedSized`)
4. **optimal** — the cost-model backward-DP optimal parse (`deflateDynamicBlocksOptimal`)

## Measurement (this session, HEAD `8765a9f5`, machine `chungus`, median-of-5, deterministic sizes)
The **optimal** candidate is the strictly smallest on every Canterbury file — including the binary `kennedy.xls`, the one file the in-code comment says the self-contained split wins (that holds at L7–L8, where optimal is not computed, *not* at L9):

| file | base | SC | shared | **optimal** | winner | opt − best |
|------|------|----|--------|-------------|--------|-----------|
| alice29.txt | 54374 | 62217 | 54127 | **52111** | OPTIMAL | 0 |
| plrabn12.txt | 194720 | 219544 | 194380 | **186472** | OPTIMAL | 0 |
| lcet10.txt | 145058 | 168537 | 144537 | **138661** | OPTIMAL | 0 |
| fields.c | 3128 | 3128 | 3128 | **3027** | OPTIMAL | 0 |
| kennedy.xls (binary) | 214321 | 198075 | 200453 | **178311** | OPTIMAL | 0 |

Wall-clock, full L9 vs the optimal candidate alone:

| file | full L9 (4 cands) | optimal cand | non-optimal overhead |
|------|-------------------|--------------|----------------------|
| alice29 | 116.5 ms | 88.5 ms | **24%** |
| plrabn12 | 377.6 ms | 281.6 ms | **25%** |
| lcet10 | 319.5 ms | 242.0 ms | **24%** |
| kennedy.xls | 1021.7 ms | 699.7 ms | **31%** |

[...]

PR #2839, opened 2026-07-17
perf: sized floor at the optimal tiers (byte-identical; L9 +1.7%, L10 +1.2%)

### LINKED ISSUE #2782: perf(explore): lower the optimal-parse floor to L8 — cheap chain-32 static-cost optimal as a lazy-512 replacement
depends-on: #2638

## Lower the optimal-parse floor to L8 — a cheap static-cost optimal as a lazy-512 replacement

The biggest *ratio* opening on the L6-8 tier. Optimal parse was proven to Pareto-dominate the split candidates at L9/L10, and the L9-fast tier (#2638) runs a single-round, bounds-free, shallow-cache DP at chain-64 for ~2x the crown's speed. The natural next Pareto push is a *cheaper-still* optimal cheap enough to replace L8's lazy chain-512:

- **chain-32** candidate cache (vs L9-fast's 64),
- **static cost tables** only ([`computeChoicesFast`](Zip/Native/DeflateParse.lean#L1005) already skips the round-2 refit),
- **no hash3**, **no boundary scan** (already dropped in the fast path).

The hypothesis: at L8's current speed, a shallow static optimal beats deep lazy on ratio. If so, L8 stops being "deep lazy" and becomes "shallow optimal."

## Proof cost: near-zero

The choice-array format is byte-identical, and the re-verifying emitter + arbitrary-choice-array validity proofs ([`LZ77OptimalCorrect`](Zip/Spec/LZ77OptimalCorrect.lean)) already generalize over *any* choice arrays. This is almost entirely a matcher-tuning experiment.

## Plan (measurement-first)

1. Add an `lz77OptimalFast`-style entry at chain-32 and wire it as an L8 candidate (`pickSmaller` against the lazy base, same as L9's dispatch).
2. Measure ratio + speed at L8 vs the current lazy-512 on both corpora.
3. If it wins, retire L8's lazy path there; if only at some file types, keep both via `pickSmaller`.

🤖 Prepared with Claude Code

kim-em:
This PR stop emitting the always-discarded floor candidate at the optimal tiers: L9/L10's `pickSmaller` fully emitted the base candidate even though #2640 measured the optimal parse strictly smaller on every corpus file — the L6-L8 tier retired exactly this via the #2753 sized-prep arbitration, and the optimal tiers now use the same `emitSmallerBy` (prep-sized floor, only the winner's thunk forced). Output is byte-identical everywhere — verified including the Canterbury floor-wins files (ptt5, kennedy.xls), where the floor genuinely beats the optimal parse and any floor weakening shows in real output. Pinned sandwich medians: dickens L9 **+1.7%**, xml L10 **+1.2%** (same-binary floors ~0.3%). The dispatch's three roundtrip proofs re-thread through `emitSmallerBy` with a level-generalized floor lemma.

The rest of the #2782 investigation closes with two rejected variants, both recorded: the depth-32 L9 remap (0.3146 @ ~5.0 weighted — 3.6% below the old L8↔L9 mixing line with the old point left uncovered; postscript in `fastChainDepth`'s docstring), and a cheap-knobs floor matcher (~+7% L9 on Silesia where the floor never wins, but it costs ptt5 L9 +5.5% output where it does — the wrong trade at the max-ratio tiers; note in the dispatch comment). The core #2782 hypothesis — a cheap optimal parse at L8's speed — is refuted: the optimal pipeline floors at ~7 MB/s even at chain 16, nowhere near L8's 25.5. Closes #2782.

🤖 Prepared with Claude Code


kim-em:
Before/after `compress_mbps` (machine `Linux chungus2 x86_64`, pinned; BEFORE = master's `latest.json` @ post-#2836 refresh; `max |Δratio| = 0.000000` — output-neutral as verified, including the ptt5/kennedy floor-wins cases):

silesia compress pareto

canterbury compress pareto

Matrix: L9 1.02x, L10 1.00–1.01x, untouched levels 0.98–1.00x (the recurring cross-run bracket; the same-worktree sandwich — **dickens L9 +1.7%, xml L10 +1.2%**, floors ~0.3% — is the authoritative magnitude). A modest, strictly-free win that also aligns the optimal tiers with the L6–L8 arbitration architecture.

Full disclosure of the measurement path: the first push of this PR claimed L9 +9% from a version that also weakened the floor's matcher knobs; the full matrix then caught that variant changing real output on ptt5 (where the floor genuinely wins), and a proper decomposition showed the knob change — not the sized prep — carried most of that speed. It was reverted and is recorded in the PR body as a rejected variant. Dashboard + animation refreshed in-PR (2c79aafb).

Issue #2799, opened 2026-07-08
perf(decode): fastloop/slowloop margin split with write-once cursor output


kim-em:
Split the decode loop into a fastloop (guaranteed input/output margin, all capacity checks hoisted to one per-iteration precondition) and the existing verified loop as the slow path for buffer edges. Inside the fastloop, output becomes write-once `uset` at a cursor into a pre-extended buffer and input becomes proven-bounds wide reads. This is the architecture step that puts a memory-safe, fully-proven decoder in miniz_oxide's performance class.

## Why

miniz_oxide reaches **524 MB/s with zero unsafe code** (`#![forbid(unsafe_code)]`). Its 2× over Go/Zig is not language magic — it is exactly this structure: a fastloop entered only while `out_remaining ≥ 259` (2 literals + max match) and `in_remaining ≥ 14`, re-checked once per token, so that inside it there are no per-literal capacity checks, no per-refill input checks, and (in Rust) enough visible invariants to elide bounds checks. libdeflate uses the same shape with margins 299 (2 literals + 258 match + copy overshoot slack) / 25.

For a verified Lean decoder the margins are exactly what the proven-bounds campaign wants: the fastloop guard *is* the bounds proof for `uset`/wide-read in the body.

## Work

Enter the fastloop only while `out_remaining ≥ 299` and `in_remaining ≥ ~25`, re-checked once per iteration. By construction of the margins, inside the fastloop:

- output writes are proven-bounds `uset` at a cursor into a pre-extended buffer — write-once, no `push`, no capacity checks (this supersedes the literal-batching issue on this path). The exact-`sizeHint` path (gzip ISIZE, ZIP sizes — the archive workloads) pre-extends to final size; the unknown-size path extends in chunks and runs the fastloop per chunk.
- input reads are proven-bounds wide reads.
- match copies may overshoot up to a word into the output margin, enabling the word-store copies (and libdeflate's 5-word unconditional prologue).

The existing verified loop stays as the byte-identical slow path for block tails and buffer edges. The equivalence obligation is "one fastloop iteration = k steps of the reference loop", with the margin invariants discharging the bounds.

## Sequencing

Multi-session structural change; this is the tracking issue for it, with the margin arithmetic recorded up front (miniz 259/14; libdeflate 299/25). Land and measure the cheap wave-1 per-symbol changes first so this is built on the lean core, not the current one.

[...]

PR #2849, opened 2026-07-18
perf(bench): route decode-density through the verified exact-size fastloop

### LINKED ISSUE #2845: bench: route decode-density through the verified exact-size fastloop (inflateSized exact:=true)
Route the decode-density benchmark through the verified exact-size fastloop so `silesia_decode_ranking` reflects the fast path that ZIP extraction already runs in production.

## Problem

The decode-density harness calls the **push decoder** — `Zip.Native.Inflate.inflate data (sizeHint := origSize)` at [bench/ZipBenchReport.lean:210](bench/ZipBenchReport.lean#L210) and again at [:369](bench/ZipBenchReport.lean#L369). There `sizeHint` only pre-reserves capacity; it does **not** engage the verified `uset` fastloop. So `silesia_decode_ranking` measures the old decoder and does **not** reflect the +6.2% fastloop shipped to ZIP extraction in #2840. Every other decode win (LTO #2809, entry-binding #2793, uget-read #2814, subtables #2819) already lives in the base `inflate` and *is* on the dashboard; the fastloop is the one banked win missing from it, purely because the benchmark invokes the wrong entry point.

## Work

- Switch the two native-decode call sites to `Zip.Native.Inflate.inflateSized data maxOut origSize (exact := true)` (the decode-density scenario already passes the true `origSize`). `inflateSized_agrees` (from #2840) proves `inflateSized = inflate` **unconditionally**, so this is byte-identical to what the dashboard measured before — only faster on the exact-size path.
- Regenerate and commit `bench/results/decode_density.json` and the `silesia_decode_ranking` / `decode_density` SVGs per the `perf-pr-graphs` skill. **Caveat (from the #2808 dashboard-refresh note):** the other-language comparator curves (go/js/ocaml/zig) must not be dropped — refresh on a box with those toolchains, or preserve their rows.
- Note in `bench/README.md` that the headline native decode number represents the **known-exact-size** production path (what ZIP extraction uses), which is the decode-density scenario (the harness feeds the true size).

## Expected

~+6.2% geomean on `silesia_decode_ranking` (dickens +8.3%, x-ray +5.9%, nci +4.4%), stacking on the already-reflected gains. This is the change that makes the metric move.

Context: finishes the visibility half of #2799 / #2840. Sibling of the gzip production-wiring follow-up.

🤖 Prepared with Claude Code

kim-em:
This PR routes the decode-density benchmark through the verified exact-size fastloop, so `silesia_decode_ranking` reflects the fast path that ZIP extraction already runs in production. The two native-decode call sites in the decode-density harness (`nativeDecompress` and the `decodeForAB` A/B diagnostic) switch from `Zip.Native.Inflate.inflate data (sizeHint := origSize)` — where `sizeHint` only pre-reserves capacity — to `Zip.Native.Inflate.inflateSized data (sizeHint := origSize) (exact := true)`, which engages the branch-free `uset` exact-size fastloop shipped to ZIP extraction in https://github.com/kim-em/lean-zip/pull/2840. `inflateSized_agrees` proves `inflateSized = inflate` unconditionally, so the dashboard measures byte-identical output, only faster on the exact-size path.

The refreshed `bench/results/decode_density.json` and the two decode SVGs (`silesia_decode_ranking`, `silesia_decode_density`) re-record the native rows through the fastloop. The committed baseline (da9d9c0e) predated https://github.com/kim-em/lean-zip/pull/2799, https://github.com/kim-em/lean-zip/pull/2819 and https://github.com/kim-em/lean-zip/pull/2840, so this also brings the stale decode dashboard current. The go/js/zig/ocaml comparator rows are preserved from the prior run — they decode byte-identical libdeflate/zopfli streams, unaffected by this change (the toolchains are not built on this box; per the #2808 dashboard-refresh caveat, preserving rather than dropping their curves).

Because the committed baseline is stale and cross-session single-pass Silesia carries large noise, the honest measure of what this routing buys is a same-build paired A/B (push no-hint vs exact-size fastloop, 20 pairs/file, pinned to one core): silesia **+5.95%** (95% CI [1.055, 1.064]), canterbury **+3.42%** — in line with the ~+6.2% the issue predicted. `bench/README.md` now notes that the headline native decode number is the known-exact-size production path (what ZIP extraction uses).

Closes https://github.com/kim-em/lean-zip/issues/2845.

🤖 Prepared with Claude Code

### DIFF
--- a/bench/README.md
+++ b/bench/README.md
@@ -216,6 +216,16 @@ memory-bandwidth ceiling on emitting the output bytes. This is the rigorous way 
 isolate a decoder: an own-encoder scatter (the lzbench / Squash convention)
 confounds decoder speed with each encoder's ratio.
 
+The native row here is the **known-exact-size production path** — the same one
+ZIP extraction runs (`Zip.Archive` decodes each entry at its declared
+`uncompressedSize`). The decode-density harness feeds the decoder the stream's
+true output size, so native decodes through
+`Zip.Native.Inflate.inflateSized … (exact := true)`: the verified branch-free
+`uset` exact-size fastloop that writes every byte once into the pre-extended
+buffer, dropping the per-literal capacity and output-size checks. It is proven
+byte-identical to the push decoder `inflate` (`inflateSized_agrees`), so the
+measured throughput is the production fast path, not a benchmark-only shortcut.
+
 Two views over the fixed-encoder streams — libdeflate at levels 1/3/6/9/12 plus a
 zopfli stream per file (the densest realistic raw DEFLATE):
 
--- a/bench/ZipBenchReport.lean
+++ b/bench/ZipBenchReport.lean
@@ -187,10 +187,11 @@ def runWorkloads
           if level > 9 then pure none
           else do
             let ref ← RawDeflate.compress data level.toUInt8
-            -- The decompressed size is known here; pass it so the native decoder
-            -- pre-sizes its output buffer, matching the production ZIP/gzip path
-            -- where the uncompressed size comes from the archive metadata. The FFI
-            -- reference decoders ignore it.
+            -- The decompressed size is known here; pass it as the exact size so
+            -- the native decoder runs its verified `uset` exact-size fastloop
+            -- (`inflateSized … (exact := true)`), matching the production
+            -- ZIP/gzip path where the uncompressed size comes from the archive
+            -- metadata. The FFI reference decoders ignore it.
             let dNs ← measureNs size theReps (dec ref size >>= sink)
             pure (mbps size dNs)
       rows := { compressor := name, pattern := pat, size := size, level := level,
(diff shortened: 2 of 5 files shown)


kim-em:
## Decode-density dashboard: routed through the exact-size fastloop

Metric: `decompress_mbps` on byte-identical libdeflate/zopfli streams. Machine `Linux chungus2 x86_64` (matches the committed dashboard). Native now decodes through `inflateSized … (exact := true)` — the same verified `uset` fastloop ZIP extraction runs.

**Honest fastloop delta (same-build paired A/B, push no-hint vs exact-size fastloop, 20 pairs/file, pinned to one core):**

| corpus | speedup | 95% CI | no-hint | fastloop |
|---|---|---|---|---|
| silesia | **+5.95%** | [1.055, 1.064] | 342.1 MB/s | 365.2 MB/s |
| canterbury | **+3.42%** | [1.030, 1.038] | 265.6 MB/s | 275.5 MB/s |

This paired same-build measurement is the trustworthy number: the committed `decode_density.json` baseline (da9d9c0e) predates #2799 / #2819 / #2840, and a cross-session single-pass Silesia overlay against it would conflate several wins with ±30% noise. The +5.95% Silesia gain matches the ~+6.2% the issue predicted.

Refreshed dashboard (native marked `◄ subject`; go/js/zig/ocaml comparator curves preserved):

silesia decode ranking

silesia decode density

The refreshed `bench/results/decode_density.json` + both SVGs are committed in this PR. Ready for review — holding on the merge for your go-ahead.

PR #2842, opened 2026-07-18
perf(tune): deepen L6's lazy probe to 18 for strict miniz_oxide-L6 dominance

kim-em:
This PR re-pair L6's lazy probe depth from 8 to 18 so the default level strictly Pareto-dominates miniz_oxide L6 on the dashboard's own metric. The post-singleton L6 dominated on the byte-weighted plane but sat 0.8bp above miniz on the per-file geomean the Pareto SVG plots (the gm64/ld8 knobs' nci/xml regressions are amplified under equal-file weighting), leaving miniz L6 un-dominated in the L6/L7 notch. The fine `gate-sweep` ld-grid (12-file geomeans, deterministic) shows depth 18 is the shallowest probe clearing miniz with real margin; the full matrix confirms: **native L6 (0.321531, 35.85 MB/s geomean) vs miniz_oxide L6 (0.321637, 35.57) — under on ratio (margin 1.1e-4, deterministic), over on speed (+0.8%)** — and the weighted-plane dominance keeps its wider margin (0.3199 vs 0.3213 at +2%). The cost is the authorized give-back from the ld8 corner: dickens L6 −6%, xml −4%, mozilla −3% (pinned interleaved sandwich, floors ~0.3%); L6 remains a frontier vertex on both planes and every other level is byte-identical. Speed-margin caveat stated plainly: +0.8% on single-pass Silesia geomean means an occasional noisy refresh could read the speed axis as a coin-flip; the ratio axis can never flip.

🤖 Prepared with Claude Code

PR #2850, opened 2026-07-18
perf(decode): wire the verified exact-size fastloop into gzip/tar.gz decode

### LINKED ISSUE #2846: perf(decode): wire the verified exact-size fastloop into gzip/tar.gz decode (follow-up to #2840)
Wire the verified exact-size `uset` fastloop into gzip / tar.gz decode, the follow-up #2840 deferred, so the +6.2% reaches the gzip decode surface too.

## Background

#2840 shipped the fastloop to ZIP extraction via `Inflate.inflateSized data maxOut sizeHint (exact := true)`, taking the exact size from the central-directory `uncompressedSize`, and proved `inflateSized_agrees` — that `inflateSized = inflate` **unconditionally, for every input**. gzip was deferred because its `ISIZE` lives in the trailer *after* the DEFLATE stream and members can concatenate, so the exact size is not available upfront the way ZIP's header size is.

## Why this is now low-risk

Because `inflateSized_agrees` is unconditional, passing `ISIZE` as the hint is **provably byte-identical to `inflate` even when `ISIZE` is wrong or malicious** — a bad hint just makes the fastloop reject and fall back to the push decoder. The gzip `ISIZE` / CRC32 trailer checks stay integrity checks, not soundness backstops. So no new heavy proof is required; this reuses the proven machinery.

## Work

- In the gzip / tar.gz decode path, when decoding a **single, non-concatenated** member whose `ISIZE` is readable from the trailer (clamped by the existing presize cap), decode via `inflateSized … (exact := true)` with `sizeHint = ISIZE`.
- Fall back to the push decoder (`exact := false`) for concatenated multi-member streams, or whenever the member boundary / trailer size is not known upfront. (`ISIZE` is size mod 2^32; the presize cap already bounds it, and outputs above the cap take the slow path.)
- Conformance tests: exact-size hit (fast path taken, bytes == push decoder == FFI), deliberately-wrong `ISIZE` (fastloop rejects → fallback → identical bytes), and a concatenated multi-member stream (fallback path).

## Expected

Ships the verified +6.2% to gzip / tar.gz decode — the largest real-world decode surface (tar.gz). Closes the container-format half of #2799's production wiring.

🤖 Prepared with Claude Code

kim-em:
This PR wires the verified exact-size `uset` fastloop into gzip / tar.gz decode, the follow-up https://github.com/kim-em/lean-zip/pull/2840 ("perf(decode): decode ZIP entries with the verified uset fastloop at exact size") deferred, so the +6.2% reaches the gzip decode surface too — the largest real-world one, tar.gz, whose decode flows through `GzipDecode.decompress`.

It adds `Inflate.inflateRawSized data startPos maxOut sizeHint exact`, the `inflateRaw` counterpart of `inflateSized`: it decodes from a byte offset and returns the byte-aligned end position (which a container decoder needs to locate the trailer / next member), dispatching to the verified `inflateRawFastU` when `exact = true` and `sizeHint > 0`, and falling back to the push `inflateRaw` when the fastloop's exact-size contract rejects. It proves `inflateRawFastU_sound` (the raw-offset reverse bisimulation, `inflateLoopCur_treeFree` instantiated at `pos := startPos`) and hence `inflateRawSized_agrees`: `inflateRawSized = inflateRaw` for every `USize`-representable input, so a wrong or malicious hint never changes the decoded bytes or the returned end position.

In `GzipDecode.decompress` it decodes the single-member case with the fastloop: for a stream with exactly one member the trailing 8 bytes are its trailer, so `ISIZE = readUInt32LE data (size - 4)` is the exact decompressed length. It attempts this only on the first member (a dedicated flag, since `result.size == 0` is not a first-member test once empty members exist) and only when the hint is bounded by `presizeCap`, the member budget, and `UInt32.size` (ISIZE is size mod 2^32, so an output below 4 GiB makes it genuinely exact). A concatenated multi-member stream, trailing padding, or a corrupt ISIZE makes the fastloop reject and fall back, byte-identically. The CRC32 / ISIZE trailer checks stay integrity checks, not soundness backstops.

Conformance tests exercise the exact-size hit (bytes == push == FFI, plus a direct `inflateRawFastU` acceptance witness that the fast path actually fires), a deliberately-wrong ISIZE (fastloop rejects, fallback decodes, the ISIZE check catches the corruption), a concatenated multi-member stream and an empty-first-member stream (fallback path), and `inflateRawSized` at the unit level (exact and wrong hints both equal `inflateRaw`, bytes and end position). Still `sorry`-free; `lake build` + `lake exe test` green.

Closes https://github.com/kim-em/lean-zip/issues/2846.

🤖 Prepared with Claude Code

### DIFF
--- a/Zip/Native/Gzip.lean
+++ b/Zip/Native/Gzip.lean
@@ -1,5 +1,6 @@
 import Zip.Native.Inflate
 import Zip.Native.InflateTreeFree
+import Zip.Native.InflateFast
 import Zip.Native.DeflateDynamic
 import Zip.Native.Crc32
 import Zip.Native.Adler32
@@ -16,6 +17,12 @@ import ZipCommon.Binary
 namespace Zip.Native
 
 namespace GzipDecode
+
+/-- Absolute ceiling on the exact-size fastloop's speculative presize allocation,
+    mirroring `Zip.Archive.nativePresizeCap`. A member whose declared decompressed
+    size (gzip trailer `ISIZE`) exceeds this keeps the push decoder rather than
+    pre-extending a large buffer up front. -/
+def presizeCap : Nat := 64 * 1024 * 1024
 
 /-- Scan forward from `pos` in `data` for the next zero byte (NUL).
     Returns the index of the zero byte, or `data.size` if none is found. -/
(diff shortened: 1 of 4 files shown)


kim-em:
## Before/after: gzip/tar.gz single-member decode

This PR wires the verified `uset` exact-size fastloop into `GzipDecode.decompress`, so a single-member gzip/tar.gz now decodes through `inflateRawSized … (exact := true)` (the fastloop `inflateFastU`) instead of the push `inflate`. The honest before/after is therefore **push `inflate` vs the `uset` fastloop on the same binary** — measured with `inflate-profile decode` vs `decode-fast-u` (the fastloop is exactly what the new gzip path dispatches to).

Machine `chungus2` (x86_64), pinned to the idlest core, level-6 raw deflate, 3 rounds (mean±sd):

gzip decode fastloop before/after

| Silesia file | before (push) | after (fastloop) | speedup |
|---|--:|--:|--:|
| dickens (10 MB) | 257.8 MB/s | 278.2 MB/s | **+7.9%** |
| webster (41 MB) | 316.6 MB/s | 337.5 MB/s | **+6.6%** |
| samba (22 MB) | 417.2 MB/s | 442.2 MB/s | **+6.0%** |
| **geomean** | | | **+6.8%** |

This matches the +6.2% #2840 measured when it wired the same fastloop into ZIP extraction — this PR extends it to the gzip container surface (the largest real-world decode, tar.gz).

### On the standard dashboard graphs

The dashboard decode benchmark (`bench-report` `nativeDecompress`) measures `Inflate.inflate` with an *inert* capacity hint — a code path this PR does **not** touch — so the committed `decompress_mbps` / decode-density numbers are byte- and perf-identical by construction (the change is proven byte-identical via `inflateRawSized_agrees`). A standard before/after corpus overlay would be flat, and no dashboard refresh is warranted from this PR: the fastloop speedup only shows on the container decode path the dashboard bench doesn't route through, which is why this A/B is the right instrument.

(Note: `bench/results/latest.json` is separately stale on the **compress** side since #2842/#2844/#2848 skipped the in-PR dashboard refresh — a pre-existing condition unrelated to this decode change.)

Issue #2837, opened 2026-07-11
perf: land the rolling lazy2 deferral at L7 (certified +0.053pp at ~+2.4% matcher; multi-session proof threading)


kim-em:
Stage 1 of #2783 is now measured and certified (spike on branch `perf/lazy2-deferral`, `bench/ZipLazyRollSweep.lean`, commit 0a42c3d5): a true rolling pending-match deferral (cap 4, /4 probe ladder), composed with the hash3 singleton and seeded probe, gains **+0.053pp corpus-weighted Silesia at L7 for ~+2.4% matcher cost** — about 5x more ratio-efficient than the L7→L8 ladder direction, a genuine frontier gain. The mirror config (cap=1) is byte-identical to production on all 24 file×level pairs, so the harness is certified. L8 gains more (+0.068pp) but overruns its speed budget (~+7.3% matcher — each extra probe is ~8x costlier at chain 512).

What blocks landing: routing L7 through the rolling arm forces a re-thread of the entire lazy correctness tower (`lz77ChainLazy.mainLoop` + `_valid`/`_encodable` → `mainLoop_eq_chainLazy` → `lz77ChainLazyIterP_eq` → `mergedLoop_eq`), and the rolling arm emits a **data-dependent number of literals**, breaking the fixed-pushes-per-branch alignment every `strongRecOn` lockstep rung relies on — each rung needs a new nested induction over the rolling sub-recursion, with `lazy2Steps` universally quantified. This is a proper multi-session proof campaign per the sorry-policy, not a one-session force.

Plan: thread a `lazy2Steps : Nat := 1` knob through the tower one rung per session (intermediate commits with clearly-marked `sorry`s per the repo policy), starting from the boxed base; the spike stays the behavioral reference (its per-file table is the acceptance test). Ladder value: L7 → cap 4; L8 stays 1.

🤖 Prepared with Claude Code

kim-em:
## Rung 1 landed (branch `feat/lazy2-rung1`, not pushed)

**Commit:** `feat: rolling lazy2 deferral, rung 1 — generalized boxed base + contracts (#2837)`

### What landed
[...]

PR #2858, opened 2026-07-19
perf: rolling lazy2 deferral at L7, fused into the merged loop hot path (#2837)

kim-em:
This PR turns on the rolling lazy2 deferral (#2837) at L7 with cap 4, completing the campaign the five proof rungs prepared. The knob flip alone was frontier-rejected: routing every accepted deferral through the mutual \`mainLoop → rollDefer\` boundary cost 17% corpus-weighted at L7 (real outcalls with ~17 args and refcount traffic on four arrays, paid on the common accept path of every file). The fix fuses \`rollDefer\` into \`lz77LazyMergedLoop\` as a single well-founded function carrying flat \`Nat\` rolling state (\`pLen pMatchPos step\`, with \`pLen = 0\` as the fresh-position sentinel), so an accepted deferral rolls by self-recursion and the hot loop never leaves itself. The seeded re-probe (depth \`maxChain/4\`, pending-length seed) matches the certified spike exactly; outputs are byte-for-byte identical to the spike's c4m0 row on all 12 Silesia files, and \`lazy2Steps = 1\` remains byte-identical to the previous production matcher at every other level. \`mergedLoop_eq\` is re-proved against the unchanged packed twins via \`seeded_probe_bridge\`; no theorem statements were weakened and no new sorries.

Measured on the full Silesia matrix (weighted plane): L7 moves 0.31960 → 0.31907 (+0.053pp) at −4.7% raw speed (~−3.3% net of the −1.2% ambient on untouched controls), which sits outside the L6↔L8 mixing curve by +10.4% and outside the old-L7↔L8 curve by +7.6% (geomean plane: +6.4% / +4.4%) — a genuine frontier expansion. Before fusion the same ratio point cost −17.1% and sat inside both curves. The dashboard and Pareto-history animation are refreshed in-PR.

🤖 Prepared with Claude Code


kim-em:
Before/after `compress_mbps` (machine `Linux chungus2 x86_64`, same as the committed dashboard; BEFORE = master's committed `latest.json`, AFTER = full pinned `bench-report --native-only` on this branch):

silesia compress pareto

canterbury compress pareto

Silesia geomean by level:

|---|---|---|---|---|
| 1 | 0.361 | 111.6 | 112.5 | 1.01x |
| 2 | 0.348 | 96.3 | 97.3 | 1.01x |
| 3 | 0.342 | 83.0 | 83.6 | 1.01x |
| 4 | 0.331 | 53.3 | 53.7 | 1.01x |
| 5 | 0.326 | 43.1 | 44.1 | 1.02x |
| 6 | 0.322 | 35.9 | 35.3 | 0.98x |
| **7** | **0.320→0.31907 (w)** | **34.4** | **32.8** | **0.95x** |
| 8 | 0.319 | 26.8 | 26.4 | 0.99x |
| 9 | 0.314 | 4.7 | 4.6 | 1.00x |
| 10 | 0.309 | 2.7 | 2.8 | 1.00x |

Frontier verdict (mixing-curve test, seconds-per-MiB): the new L7 point (0.31907 weighted @ 33.1 MB/s) is **outside** the L6↔L8 mixing curve by +10.4% and outside the old-L7↔L8 curve by +7.6% (geomean plane: +6.4% / +4.4%) — a genuine Pareto expansion. Only L7's ratio changes (max |Δratio| 0.0011 at reymont, improving); every other level is byte-identical, controls at 0.99–1.03x.

Caveats: the −4.7% raw L7 speed reads ~−3.3% net of the −1.2% ambient drift visible on untouched controls. Before the fusion commit this same ratio point cost −17.1% and was frontier-rejected; the cap-2 discriminator (−16.5% with rolls halved) pinned the cost on the mutual-boundary crossings rather than the extra probes, which is what the fusion removes. Byte-identity vs the certified spike (c4m0, all 12 Silesia files) re-verified on the fused loop.

PR #2859, opened 2026-07-21
perf(tune): shallow L6's lazy probe behind the rolling deferral for stronger miniz dominance (#2837)

kim-em:
This PR retunes L6 to exploit the fused rolling lazy2 deferral from https://github.com/kim-em/lean-zip/pull/2858 ("perf: rolling lazy2 deferral at L7, fused into the merged loop hot path"): the lazy probe shallows from depth 18 (set by https://github.com/kim-em/lean-zip/pull/2842, "perf(tune): deepen L6's lazy probe to 18 for strict miniz_oxide-L6 dominance") to depth 10, paired with rolling cap 2, so the roll buys back — and widens — the ratio margin the shallow probe gives up. L6's `goodMatch = 64` gate keeps the roll cheap (only sub-64 pending matches roll).

Ratio grid over the exact production dispatch (12 Silesia files; weighted / geomean; miniz_oxide L6 = 0.32131 / 0.32164) picked ld10 x cap2 as the knee. It strictly beats the shipped L6 on all three axes: weighted ratio 0.31985 (was 0.31994), geomean ratio 0.32143 (was 0.32153, doubling the margin under miniz to -0.020pp), and +1.2% median compress throughput on a self-controlled interleaved A/B (webster +1.2%, dickens +2.4%, mozilla +1.1%, nci +0.2%); whole-tar wall neutral with 18KB smaller output. ld8 cannot hold the geomean even at cap 4, so https://github.com/kim-em/lean-zip/pull/2842's depth is not simply revertible; the rolling lever is what lets the probe shallow at all.

Against miniz_oxide L6 the picture is honest and plane-dependent. On the weighted (bytes/time) plane native L6 strictly dominates — better ratio and faster — with a margin this PR widens. On the geomean (per-file) plane native wins the ratio (now 2x the margin) but sits within noise of a speed tie (~-1% warm, structural: native trails on the mid-size binaries nci/osdb/samba where the h3 singleton and split arbitration cost per-file overhead miniz does not pay), so it is a ratio-dominant Pareto point there, not strict dominance. Closing that plane needs a pipeline-speed lever, not a parse-knob turn — tracked as follow-up.

The change is a within-native frontier gain on both planes (outside the L5-L7 mixing curve by +3.9% weighted / +2.1% geomean). It redistributes per-file ratio (a little given up on the less-compressible binaries, more won back on compressible text) but nets better on both aggregate planes, and the office/bitmap files (kennedy.xls, ptt5) both improve. Untouched levels move <=0.5% with byte-identical output; the only ratio change is at L6. Pure parse heuristic: the proof tower (`mergedLoop_eq` and the encoder contracts) holds for every `lazy2Steps`, so no obligation moves.

🤖 Prepared with Claude Code


kim-em:
Before/after `compress_mbps` (machine `Linux chungus2 x86_64`, same as the committed dashboard; BEFORE = master's `latest.json` with the old ld18 L6, AFTER = full pinned native matrix on this branch):

silesia

canterbury

**The overlay under-resolves this change** — it is a sub-1% retune and single-pass Silesia carries ±30% per-file noise, so the L6 dot reads flat (1.00x) here. The real signal is in the steady-state self-controlled A/B (in the commit body): **+1.2% median compress throughput** vs the shipped L6, whole-tar wall neutral with 18KB smaller output.

Ratio (12 Silesia files; miniz_oxide L6 = weighted 0.32131 / geomean 0.32164):

| plane | native L6 before | native L6 after | vs miniz (after) |
|---|---|---|---|
| weighted | 0.31994 | **0.31985** | −0.147pp, **+0.6% faster → strict dominance** |
| geomean | 0.32153 | **0.32143** | −0.020pp ratio (2× the margin), speed ~tie (−1% warm) → ratio-dominant Pareto point |

Frontier: the new L6 is outside the L5↔L7 mixing curve by **+3.9% weighted / +2.1% geomean** (a genuine within-native gain on both planes). Visually, native L6 sits right on miniz_oxide's L6 triangle — the ratio edge with the per-file speed tie, made visible.

Caveats: (1) geomean strict dominance over miniz is **not** achieved — native trails ~1% on the mid-size binaries (nci/osdb/samba) where the h3 probe + split arbitration cost per-file overhead; that is a pipeline-speed gap, tracked as follow-up, not closable by this parse knob. (2) The change redistributes per-file ratio (a little given up on less-compressible binaries, more won on compressible text) but nets better on both aggregate planes; kennedy.xls and ptt5 both improve, so no regression on the office/bitmap class. Untouched levels ≤0.5%, byte-identical.

PR #2860, opened 2026-07-21
perf: content-gate the L6-8 hash3 length-3 probe, completing strict miniz-L6 dominance (#2837)

kim-em:
This PR gates the hash3 length-3 singleton probe (levels 6-8) on a cheap content prescan, turning it on only for low-compressibility inputs where it earns ratio and off everywhere else, where profiling showed it was strictly dominated — worse ratio AND 12-21% slower. The result completes the L6 campaign: native L6 now strictly dominates miniz_oxide L6 on BOTH the weighted and the geomean plane — better ratio and faster on each — where before it was a ratio win with a per-file speed tie.

The static `useH3Level level` becomes `useH3For data level`, which keeps the level-6-8 gate but additionally requires `h3IncompressibleScan data` (or `data.size < 1MiB`, preserving small-input bytes). The scan reuses the existing `incompressiblePrescan` region layout (four 32KiB windows spread end to end) and its single-hash 4-gram filter, firing when any region's 4-gram collision fraction falls below 61% — a threshold that cleanly separates the four ratio winners (x-ray 9%, sao 20%, ooffice 46%, mozilla 58%) from every loser (all regions >=64%). osdb (53%) is unavoidably kept on and reproduces the master bytes there.

On the 12-file Silesia matrix the gate improves BOTH planes at L6/L7/L8: L6 weighted 0.31985 -> 0.31905, geomean 0.32143 -> 0.32043; L7 and L8 track the same -0.08pp / -0.09pp. Against miniz_oxide L6 that is -0.227pp weighted / -0.121pp geomean ratio, at +6.3% / +5.3% speed — strict dominance on both. L6 sits outside the L5-L7 mixing curve by +5.1% weighted / +4.4% geomean. The speedup comes from the losers no longer running the probe (nci -11.5%, webster -11.8%, samba -7.9% per-file); stay-on files pay a one-time ~8ms scan and keep byte-identical output. The mechanism is a genuine both-axes win because h3 was actively costing ratio on structured files, not just speed.

Untouched levels (1-5, 9, 10) are byte-identical with 0.000 ratio delta; only L6-8 move, all in the improving direction. The matcher contracts are already universally quantified over `useH3 : Bool`, so no theorem statement changed — the dispatch proofs (`lzMatch_{encodable,empty,resolves}`, `lzMatchP_{eq,map}`) simply pass the data-dependent bool to the same parametric lemmas. Zero new sorries.

🤖 Prepared with Claude Code


kim-em:
Before/after `compress_mbps` (machine `Linux chungus2 x86_64`, same as the committed dashboard; BEFORE = master's `latest.json` (post-#2859 L6), AFTER = full pinned native matrix on this branch):

silesia

canterbury

The Silesia overlay resolves the change clearly (L6/L7/L8 at **1.06x / 1.05x / 1.04x**): the AFTER curve moves up-and-left of miniz_oxide's L6-8 triangles — faster **and** better ratio. Canterbury is unchanged (its files are <1 MiB, so the gate keeps the static decision; the win is entirely on Silesia's large real-world files).

**Native L6 now strictly dominates miniz_oxide L6 on both planes:**

| plane | native L6 (after) | vs miniz L6 ratio | vs miniz L6 speed | strict dominance |
|---|---|---|---|---|
| weighted | 0.31905 @ 37.76 | **−0.227pp** | **+6.3% faster** | ✅ |
| geomean | 0.32043 @ 37.45 | **−0.121pp** | **+5.3% faster** | ✅ |

The geomean plane was a ~1% speed *deficit* one PR ago; it is now a +5.3% win with a 6× wider ratio margin. L6 is outside the L5↔L7 mixing curve by +5.1% weighted / +4.4% geomean. L7/L8 improve too (the gate spans 6-8): +5.3%/+4.1% speed, −0.08/−0.09pp ratio. Untouched levels 1-5, 9, 10 are byte-identical (ratio Δ 0.000, speed within ±1% noise); max |Δratio| 0.0053 at mr L7, in the improving direction.

Why it is a both-axes win: the h3 probe was *actively costing ratio* on structured/compressible files (nci −1.1%, mr −1.4%, webster −0.55% when removed) while also costing 12-21% speed there. It only earns ratio on incompressible binaries (x-ray +3.9%, sao +2.4%). Gating keeps every win and deletes the waste. Zero new sorries; no theorem statement changed (the contracts are already ∀ useH3).

PR #2861, opened 2026-07-21
perf(tune): anchor L6 at ld8/no-roll for strict whole-tar miniz domination (#2837)

kim-em:
This PR reverts L6's lazy probe to depth 8 (chainDepth/8) and turns off its rolling lazy2 deferral, restoring the pre-#2842 configuration that strictly dominates miniz_oxide L6 on the single-stream `zip silesia.tar` workload — the priority use case — in both size and wall time. It is step 1 of a revert / pin / recover plan: anchor the whole-tar win, gate it in CI, then recover the per-file geomean margin without breaking it.

The intervening L6 tuning (ld18 in #2842, then ld10/cap2 in #2858/#2859) each bought per-file geomean ratio margin but spent whole-tar wall time. The cause is structural, not compute: lean materializes the full 45.9M-token, ~367 MB token array (larger than the 211 MB input; miniz streams through a bounded window and never allocates it) and traverses it about four times, and the resulting ~0.24s page-fault system-time tax is level-independent. So added matcher work pushes the single-shot wall past miniz even when steady-state per-file throughput reads faster. Whole-tar L6 (silesia.tar, best-of-5):

| config | size | wall | vs miniz |
|---|---|---|---|
| miniz_oxide L6 | 68,112,444 | 5.805s | — |
| ld10/cap2 (before) | 67,798,134 | 5.970s | smaller, +2.8% slower |
| ld8/no-roll (this) | 67,944,422 | 5.721s | smaller AND 1.4% faster |

ld8 still clears miniz on both per-file planes (0.31969 weighted / 0.32132 geomean vs miniz 0.32131 / 0.32164), by a slimmer geomean margin than the superseded ld10/cap2 point (0.31905 / 0.32043) — that margin is the deliberate give-back, recovered later once the token-footprint work (fault-reserve + `ByteArray` unboxing) widens the whole-tar time headroom. The h3 content gate (#2860) and L7's rolling are untouched (both whole-tar-safe). Pure parse heuristic: no theorem statement changes, zero new sorries.

🤖 Prepared with Claude Code


kim-em:
**The justification for this PR is the whole-tar single-shot, which the per-file overlay under-resolves.** `zip silesia.tar` at L6 (211,957,760 B, best-of-5 pinned, chungus2):

| config | size | wall | vs miniz |
|---|---|---|---|
| miniz_oxide L6 | 68,112,444 | 5.805s | — |
| ld10/cap2 (before) | 67,798,134 | 5.970s | smaller, **+2.8% slower** |
| **ld8/no-roll (this PR)** | **67,944,422** | **5.721s** | **−168 KB AND −1.4% faster** ✅ |

Strict domination on both axes restored. Confirmed on user-CPU too (task-clock 5,660ms vs miniz 5,818ms, −2.7%).

Before/after `compress_mbps` per-file overlay (BEFORE = master post-#2860; AFTER = this branch):

silesia

Per-file L6 reads +4% faster (37.4→39.0 MB/s geomean); other levels flat; max |Δratio| 0.0013 (fields.c L6). The per-file **geomean ratio** is the deliberate give-back: 0.32043 → 0.32132 (still under miniz's 0.32164, but the −0.121pp margin shrinks to −0.032pp). Weighted 0.31905 → 0.31969 (miniz 0.32131 — dominance preserved).

Why the whole-tar margin is only ~2%: lean's ~0.24s system-time fault tax from the 367 MB materialized token array (vs miniz's bounded streaming window). That tax is the target of the follow-up footprint work (fault-reserve + ByteArray unboxing), which widens the margin and funds recovering the geomean give-back — step 3 of the plan.

Not auto-merged — this is a deliberate ratio trade for review.

PR #2870, opened 2026-07-22
perf(mem): unbox the LZ77 token stream (Array UInt32 → ByteArray), −25% peak RSS, wall-neutral

### LINKED ISSUE #2868: perf(mem): retry token-stream unboxing (Array UInt32 → ByteArray) with capacity pre-sizing + sandwich verification (#2837)
## Goal

Unbox the LZ77 token stream: store tokens as a 4-byte-per-token `ByteArray` (`TokenArray`) instead of Lean's 8-byte-per-boxed-slot `Array UInt32`, threaded through `deflateRaw` end-to-end, to **reduce peak RSS**. Motivation is memory footprint (defends a "Lean is faster than Rust" writeup against the "but Lean uses ~2.4× the memory" counter). It is NOT a wall-clock play — measurement showed lean already ties/wins wall within noise; wall is user/CPU-bound where lean's codec is faster.

The token buffer is the single largest allocation: ~45.9M tokens × 8 B ≈ 367 MB on silesia.tar, larger than the 211 MB input. miniz_oxide streams through a bounded window and never materializes it. Halving to 4 B/token is the target.

## History: attempt #1 merged then REVERTED

PR #2866 (branch `perf/token-unboxing`) merged then was reverted by #2867. It was **byte-identical** (`max |Δratio| = 0.000000`, all levels) and the proofs were sound (see "recoverable assets"), **but it REGRESSED both axes it claimed to improve**:
- Compress **~8–14% slower** (L1–L8), monotonic by level, converging to ~0.99× at the DP-bound L9/L10.
- Peak RSS **~6–8% HIGHER** (true merge-base 661 MB → branch 711 MB).

The "−33% RSS win / wall-neutral" originally reported was WRONG: it compared against a bogus 1067 MB baseline (a mis-measurement), and did not use a same-worktree sandwich. A rigorous re-measurement (below) found the regression.

## Root-cause hypothesis (START HERE)

The refactor **dropped the `Array.emptyWithCapacity data.size` pre-sizing** — the matcher seeds `TokenArray.empty`, which grows by **doubling** as tokens are pushed. That single change plausibly explains BOTH symptoms:
- Doubling reallocation churn → the per-token CPU cost (compress slowdown).
- The doubling growth transient (up to ~2× the live bytes momentarily) → dominates *peak* RSS even though steady-state token bytes shrink.

**First thing to try:** add `TokenArray.emptyWithCapacity (n : Nat)` (pre-allocate `ByteArray` of `4*n` bytes) and seed the token accumulators with it (mirroring the old `Array.emptyWithCapacity data.size`), then RE-MEASURE. The 4 B-vs-8 B steady-state win should then materialize without the realloc cost.

## MANDATORY measurement protocol (this is what caught the regression; do NOT skip)

Do NOT trust any headline RSS/wall number. For every claim:
1. **Same-worktree sandwich**: build BEFORE (merge-base) and AFTER (branch) binaries in the SAME worktree/`.lake` (a fresh-worktree build can read ~9% off due to stale link state). Interleave BEFORE→AFTER→BEFORE per rep, one PINNED core (`bench/pin_core.sh`), ≥4 reps.
2. **Noise floor**: run one binary against itself; require the self-ratio ≈ 1.00 (±~0.2%) or the run is contaminated (contention). The real run had a 0.1% floor.
3. **Flat-decode discriminator**: measure the UNTOUCHED byte-identical DECODER (`decompress_mbps`) in the same session. This refactor doesn't touch decode, so decode MUST stay flat (~1.00×). If a compress "regression" also depresses decode → it's a whole-binary artifact (stale `.lake`/contention). If decode is flat but compress drops → the compress cost is REAL. (In attempt #1, decode was flat and compress dropped 8–14% — real.)
4. **True baseline**: the pre-PR merge-base peak RSS is ~661 MB (202 input + 367 tokens + 68 output + overhead), NOT 1067 MB. Measure it yourself; don't inherit a number.
5. **Peak RSS**: GNU `time -v` MaxRSS, ≥8 interleaved reps, same worktree.
6. **Byte-identity**: `max |Δratio| = 0.000000` across L1–L10 both corpora, AND `cmp`/checksum of compressed streams on Silesia + Canterbury floor files (ptt5, kennedy.xls).

## ACCEPTANCE / go-no-go

Merge ONLY if the sandwich shows: peak RSS materially LOWER than the 661 MB baseline AND compress throughput within noise (no per-level regression, decode flat) AND byte-identical. If it does not deliver the RSS win with no compress regression, do NOT merge — report the honest trade and stop. (The whole point of this retry is to hold the RESULT to the same measure-first bar the premise was held to.)

## Recoverable assets (the reverted commits — the refactor + proofs were correct, only the perf was bad)

On the reverted `perf/token-unboxing` history (before the #2867 revert), these commits are cherry-pickable as the staged blueprint — each was byte-identical + built green + zero new sorries, and the `map_push → push_toArray` proof substitution was validated MECHANICAL throughout:
- `a6d7a405` stage 1: `Zip/Native/TokenArray.lean` — the container (`ByteArray` newtype + erased `aligned : bytes.size % 4 = 0` proof field) + bridge lemmas `empty_toArray`/`size_toArray`/`get_toArray`/`push_toArray`/`extract_toArray` + `ByteArray.pushUInt32LE`. **`push_toArray`/`extract_toArray` are genuinely load-bearing proofs** (reach the actual bytes), per a Codex review.
- `8ec8ae4d` stage 2: greedy twin (L1–3).
- `682dde30` stage 3: lazy/merged production matcher (L4–8); the 2M-heartbeat `mergedLoop_eq` needed only its accumulator type changed, body unchanged.
- `dc846e94` stage 4: fused matcher+freqs.
- `9fca1217` stage 5+6: end-to-end integration (retype consumers/emit, thread `TokenArray` through `deflateRaw`, remove the last runtime `.toArray`). NOTE: emit-consumer retype has a hard ordering dependency on the source flip — do source side (lzMatchP/deflateRawBaseP*/DeflateRoundtrip) FIRST, then consumers.
- `d4150621` refinement proofs (from a Codex second opinion — REQUIRED for a verified library): the split heuristic and sizing walkers were retyped in-place with byte-identity by TEST not PROOF. `d4150621` adds real refinement theorems — `chooseSplitsHeuristicP_toArray` (full scalar-state cut-list equality via a `funext`+IH trick sidestepping the 44-leaf match) and `sharedPartitionSizedP_toArray`/`deflateRawBasePPrep_fst_toArray` (pins the size-selection `.1`). KEEP THESE — control-flow functions must be proven byte-identical, since `inflate_deflateRaw` proves any VALID output decodes, not that output is UNCHANGED.
- `4eab40e3` fixed a real build break: `ZipBench.lean`'s `match-hist` loop failed to type-check because `lzMatchP` returned `TokenArray` (no `ForIn`) — CI misses it (doesn't build `bench`). Any retry must fix this too and should build `bench` in verification.

## Known gotchas
- Dropping `emptyWithCapacity` is the suspected regression — add `TokenArray.emptyWithCapacity` and seed with it.
- `bench` exe isn't built by CI; build it (`lake -d bench build bench`) in verification or the `match-hist` break slips through.
- The staged approach kept a `.toArray` frontier moving outward; RSS only changes at the final integration (stage 6) — so measure at the END, not intermediate stages.
- Codex LOW note: `TokenArray`'s public constructor allows any 4-aligned `ByteArray`, so "built via push" provenance is not enforced (docstring only) — not a soundness hole (all bridge lemmas are universal over aligned inputs).

## Suggested plan
1. Recreate `perf/token-unboxing` from the reverted commits (cherry-pick `a6d7a405..d4150621`, or re-derive).
2. Add `TokenArray.emptyWithCapacity` and seed the matcher token accumulators with `data.size` capacity.
3. Run the MANDATORY sandwich protocol. Only proceed if RSS drops with no compress regression.
4. If it delivers: re-open the PR with sandwich-verified numbers (peak RSS before/after on the true baseline, flat decode, byte-identity) — and this time hold the graphs/RSS result to the same bar, don't merge on a headline.

Related: the peak-RSS *tracking* infra (`whole_tar_l6.json` gains `maxrss_kb`/`rss_ratio`) is being salvaged separately (it's bench-only and correct).

kim-em:
This PR unboxes the LZ77 token stream: it stores the packed tokens as a 4-byte-per-token `ByteArray` (`TokenArray`) threaded through `deflateRaw` end-to-end, instead of Lean's 8-byte-per-slot `Array UInt32`, cutting peak RSS on a whole-tar compress by 25–30% with compress throughput neutral-to-slightly-faster and byte-identical output. The token buffer is the single largest allocation (~45.9M tokens × 8 B ≈ 367 MB on silesia.tar, larger than the 211 MB input); halving it to 4 B/token is the win.

This is the measure-first retry of the reverted attempt (https://github.com/kim-em/lean-zip/pull/2866 "unbox the LZ77 token stream", reverted by https://github.com/kim-em/lean-zip/pull/2867). That attempt was byte-identical but regressed both axes it claimed to improve — peak RSS was actually higher and compress was 8–14% slower — because it dropped the accumulator pre-sizing and paid four FFI calls per token for the pack/unpack. Two changes fix both:

- Seed the matcher accumulators with `TokenArray.emptyWithCapacity data.size`, pre-allocating the backing `ByteArray` so it never grows by doubling. Each LZ77 token consumes at least one input byte, so `data.size` token capacity is never exceeded and the buffer never reallocates — the doubling-growth transient (up to ~2× the live token bytes) that dominated peak RSS in the reverted attempt is gone.
- Compile `TokenArray.push` and `TokenArray.get`, via `@[implemented_by]`, to the project's existing single-call wide `ByteArray` primitives (`pushUInt64LE` / `ugetUInt32LE` from `Zip/Native/Wide.lean`, the same wide loads `lz77Greedy.hash3` already uses) instead of four `ByteArray.push` / `getElem` calls each. The reference bodies stay the four-op forms, so every bridge and refinement proof is unchanged — only the compiled runtime swaps 4 calls for 1, matching the boxed `Array UInt32`'s one-call-per-token cost.

Measurement followed the issue's mandatory protocol: same-worktree sandwich (BEFORE at the merge-base and AFTER built in the same `.lake`), one pinned core, GNU `time -v` MaxRSS, interleaved reps, a before-vs-before noise floor of ±0.06%, and byte-identity confirmed by matching `deflateRaw` output size and a rolling checksum across L1–L10 on silesia.tar plus the Canterbury floor files (ptt5, kennedy.xls).

| level | peak RSS (before → after) | compress wall |
| --- | --- | --- |
| L1 (greedy) | 704 → 495 MB (−29.6%) | −2.3% |
| L6 (lazy) | 649 → 486 MB (−25.1%) | −3.5% |

The control-flow functions (split heuristic, sizing walkers) carry real byte-identity refinement proofs, not just tests, since `inflate_deflateRaw` proves any valid output decodes rather than that the output is unchanged. The full library builds with zero new sorries, `lake exe test` passes, and `bench` (not built by CI) builds.

Closes https://github.com/kim-em/lean-zip/issues/2868

🤖 Prepared with Claude Code


kim-em:
## Performance verification — RSS win, wall-neutral, byte-identical

This is a memory-footprint change, so the headline is **peak RSS**, not the speed-vs-ratio Pareto (the ratio does not move — output is byte-identical). Both are below.

### Peak RSS — whole-tar L6 (`zip silesia.tar`, GNU `time -v`, cold single-shot CLIs)

| | lean maxrss | rust maxrss | rss_ratio (lean/rust) |
| --- | --- | --- | --- |
| **before** | 664 MB | 276 MB | **2.405** |
| **after** | 482 MB | 276 MB | **1.745** |

−27% lean peak RSS; the "Lean uses ~2.4× the memory" gap closes to 1.74×. Output byte-identical, codec still smaller and codec-faster (`size_margin_pct` +0.25%, `time_ratio_median` 0.957), end-to-end wall neutral (`wall_ratio_median` 0.973).

Per-level same-worktree sandwich (pinned core, GNU `time -v`, ≥6 interleaved reps, before-vs-before noise floor ±0.06%, byte-identical L1–L10):

| level | peak RSS before → after | compress wall |
| --- | --- | --- |
| L1 (greedy) | 704 → 495 MB (−29.6%) | −2.3% |
| L6 (lazy) | 649 → 486 MB (−25.1%) | −3.5% |

### Compress speed vs ratio (native before/after; machine `Linux chungus2 x86_64`)

Included for the mandatory no-regression check. `max |Δratio| = 0.000000` (output-neutral), before/after curves coincide — the change trades no speed and no ratio for the memory win.

silesia compress pareto

canterbury compress pareto

silesia native compress_mbps geomean by level (before → after):

| --- | --- | --- | --- | --- |
| 1 | 0.361 | 111.4 | 111.7 | 1.00x |
| 3 | 0.342 | 82.7 | 83.3 | 1.01x |
| 5 | 0.326 | 43.9 | 45.0 | 1.02x |
| 6 | 0.321 | 38.8 | 39.4 | 1.02x |
| 8 | 0.318 | 27.5 | 27.9 | 1.01x |
| 10 | 0.309 | 2.7 | 2.7 | 0.99x |

Caveat: the overlay's BEFORE is the committed `latest.json` snapshot (a prior session), so its few-percent per-level scatter is cross-session noise, not this PR — the authoritative speed verdict is the same-worktree sandwich above (noise floor ±0.06%), which shows wall neutral-to-faster at every level.

PR #2876, opened 2026-07-25
chore: bump toolchain to v4.33.0-rc1

kim-em:
This PR moves the toolchain from v4.30.0-rc2 to v4.33.0-rc1, and repins lean-zip-common to the bump that landed there (https://github.com/kim-em/lean-zip-common/pull/4 "chore: bump toolchain to v4.33.0-rc1", merged).

`lake build` and `lake exe test` are clean in the root package and in `bench/`, and the sorry count is unchanged. No theorem statement changed: every fix is tactic drift.

## What broke, and why

Nearly all of it is one theme. A step that used to reduce something — the option/except monad, an `if`, a `match` on a literal scrutinee, a numeral, a constant — no longer does, so the tactic after it either has nothing left to do (`simp made no progress`, `No goals to be solved`) or has too much (`unsolved goals`).

- **Now-redundant tactics**: five trailing `rfl`s in `EmitPackedCorrect`, and the `pure`/`Except.pure` unfoldings in `BitReaderInvariant`, `DecodeCorrect`, `DynamicTreesCorrect` (six sites) and `InflateRawSuffix`.
- **Reductions that stopped happening**: twelve sites in `InflateBufCorrect`, where paired rewrites in nested `by_cases` blocks left a `match throw …, throw …` that used to close by the `rw`'s implicit `rfl`.
- **`simp only [bind, Option.bind]`** no longer reduces the option monad in `BitstreamCorrect`; the explicit monad lemmas do, and are what the neighbouring theorem already used.
- **`List.headD_cons` cannot fire** under `emitDynBlock`'s dependent proof arguments (`DeflateBlockSplit`, `LZ77PackedCorrect`). The reduction is definitional, so `rfl` does it.
- **Literals and constants that no longer normalize** inside a `simpa`: `UInt16.toNat 16`, `(Array.replicate 19 0).size`, `HuffTree.fastBits`, and `(k : UInt64) <<< 3` for each of the eight byte offsets in `ugetUInt64LE_byte`.
- **`Array.getElem_toList`'s bounds side goal** is now stated over `toList.length`, so `omega` needs `Array.length_toList` first.
- **`ByteArray.beq`** is in core, so the `ZipTest` copy goes. Same story for the twelve `ZipForStd` lemmas in the companion PR.

Two sites needed more than a mechanical fix:

**`decodeHuffmanCurTablesU_eq` no longer rewrites** into the unreduced `match btype with`. The blocker is the proof argument's *type*, not the proof: folded through `Inflate.fixedLitTF` where the lemma is applied, unfolded in the `_proof_1` the goal carries, and the two no longer unify at the transparency `rw` matches with. Quantifying over the proof with its type spelled the goal's way fixes it.

**`generalize` does not terminate** on the packed-token goal in `Zip/Native/Deflate.lean` — a `whnf` timeout still running at 2,000,000 heartbeats, in a module that used to build in under a minute. That one is an upstream regression, minimised to seventeen lines with no imports; it is fast on v4.30.0-rc2, v4.31.0-rc1 and v4.32.0:

```lean
inductive Tok where
  | literal : UInt8 → Tok
  | reference : Nat → Nat → Tok

def packTok : Tok → UInt32
  | .literal b => b.toUInt32
  | .reference len dist => ((1 : UInt32) <<< 31) ||| (len.toUInt32 <<< 16) ||| dist.toUInt32

def unpackTok (w : UInt32) : Tok :=
  if w &&& ((1 : UInt32) <<< 31) = 0 then .literal w.toUInt8
  else .reference ((w >>> 16) &&& 0x7FFF).toNat (w &&& 0xFFFF).toNat

set_option maxHeartbeats 400000 in
example (len dist : Nat) : unpackTok (packTok (.reference len dist)) = .reference len dist := by
  simp only [packTok, unpackTok]
  generalize len.toUInt32 = l
  sorry
```

Writing that argument inline in the goal statement instead of reaching it by unfolding `packTok` is fast, so the trigger is the term the `simp only` produces rather than the displayed goal. Worked around here by stating the reference case as its own lemma over `UInt32` variables, which is what the `generalize` was producing anyway.

## Note

The `ZLIB_LDFLAGS` workaround downstream consumers needed becomes unnecessary on this toolchain: the failed-spawn buffer duplication behind https://github.com/kim-em/lean-zip/pull/2875 "fix: don't spawn probe tools that are not installed" is fixed from v4.31.0-rc1 on. The `PATH` lookup added there is still correct and stays.

🤖 Prepared with Claude Code


kim-em:
## Performance verification — free speedup, byte-identical output

The bump changes no executable code (the diff vs merge-base is 19 files, all proofs), so this measures **one thing: the compiler**. `v4.30.0-rc2` → `v4.33.0-rc1` on the same source is worth roughly **+3.5% compress** and **+6.7% decompress**.

Machine `Linux chungus2 x86_64` (matches the dashboard). Both binaries built in the *same* worktree — BEFORE from the merge-base `74925812`, AFTER from this branch — then run interleaved (B, A, B, A) pinned to core 22, whose SMT sibling was also idle. The committed `latest.json` could **not** serve as BEFORE here: the freshness guard flags it stale, since #2873 and #2874 landed after the recorded snapshot.

`max |Δratio| = 0.000000`, `max |Δout_size| = 0` — output is byte-identical at every level, as it must be for a compiler change.

silesia compress pareto

canterbury compress pareto

The AFTER curve sits uniformly just above BEFORE at identical x positions on both corpora — a clean vertical shift, no ratio movement. Position in the field is unchanged: still above JS/OCaml, below the Go/Zig/zlib cluster in raw speed but reaching further left in ratio, with libdeflate's SIMD curve well above.

### compress_mbps geomean by level (all four runs)

| lvl | canterbury before → after | | silesia before → after | |
| --- | --- | --- | --- | --- |
| 1 | 77.5 → 80.2 | 1.034x | 109.2 → 113.0 | 1.021x |
| 2 | 68.2 → 70.7 | 1.036x | 93.9 → 96.9 | 1.019x |
| 3 | 61.0 → 63.1 | 1.034x | 80.2 → 83.1 | 1.022x |
| 4 | 43.1 → 44.4 | 1.032x | 52.0 → 54.3 | 1.031x |
| 5 | 36.5 → 37.8 | 1.037x | 44.1 → 45.7 | 1.027x |
| 6 | 30.9 → 32.3 | 1.043x | 39.0 → 39.9 | 1.017x |
| 7 | 28.1 → 29.5 | 1.049x | 34.9 → 35.5 | 1.016x |
| 8 | 24.0 → 24.9 | 1.036x | 27.9 → 28.2 | 1.009x |
| 9 | 4.8 → 4.9 | 1.026x | 4.6 → 4.7 | 1.020x |
| 10 (crown) | 2.7 → 2.8 | 1.025x | 2.7 → 2.8 | 1.036x |
| **geomean** | | **1.035x** | | **1.022x** |

### decompress_mbps geomean by level

| lvl | canterbury before → after | | silesia before → after | |
| --- | --- | --- | --- | --- |
| 1 | 227.5 → 244.5 | 1.074x | 275.9 → 298.4 | 1.072x |
| 3 | 243.6 → 261.1 | 1.071x | 307.3 → 334.0 | 1.078x |
| 6 | 252.2 → 268.2 | 1.063x | 320.1 → 341.2 | 1.054x |
| 9 | 256.8 → 273.3 | 1.063x | 326.0 → 340.4 | 1.040x |
| **geomean** | | **1.067x** | | **1.057x** |

### Noise floor (same binary, same protocol)

The two BEFORE runs against each other give the floor for free:

| | floor (b2/b1) | effect (a/b) | verdict |
| --- | --- | --- | --- |
| canterbury compress | 1.000x | **1.035x** | real — clears the floor at every one of the 10 levels |
| canterbury decompress | 1.003x | **1.067x** | real |
| silesia compress | 1.034x | 1.022x | **inside the floor** — consistent in sign, not resolvable |
| silesia decompress | 1.022x | 1.057x | real |

Canterbury is median-of-5 and lands a ±0.04% floor; Silesia is single-pass and its floor came out at ±3.4%, wide enough to swallow the compress delta on its own. Read the compress verdict off Canterbury and treat Silesia's +2.2% as corroborating sign only. The box was busy throughout (42 live agent sessions, ~16 benchmarking lean-zip in other worktrees), which is the likely source of that wide single-pass floor.

Dashboard refreshed in-PR (`bench/results/latest.json` + `bench/graphs/*.svg`) so master's recorded BEFORE reflects the new toolchain.

🤖 Prepared with Claude Code

kim-em:
## Performance verification, re-measured on the current merge-base

Superseding the earlier comment: that one measured against `74925812`, before #2877 and #2878 landed. This is BEFORE = `486a77db` (the current merge-base) vs AFTER = this branch, so it isolates the compiler on today's code.

Both binaries built in the same worktree, run interleaved (B, A, B) pinned to core 16 with an idle SMT sibling. `max |Δratio| = 0.000000`, `max |Δout_size| = 0` — byte-identical output at every level.

silesia compress pareto

canterbury compress pareto

Before/after all but coincide (the grey dashed BEFORE hides under the red AFTER) at identical x positions. Worth noting what the picture shows about #2877 rather than about this PR: lean-zip's L1 point now sits at ratio 0.361 / 166 MB/s on silesia, *above* miniz_oxide's curve there — the L1 Pareto-domination work is visible in the field positioning.

| | canterbury | silesia |
| --- | --- | --- |
| compress | **1.049x** | 1.016x |
| decompress | **1.037x** | 1.014x |
| same-binary floor (compress) | 0.978x | 0.999x |
| same-binary floor (decompress) | 0.996x | 1.006x |

### compress_mbps geomean by level

| lvl | canterbury before → after | | silesia before → after | |
| --- | --- | --- | --- | --- |
| 1 | 100.1 → 106.5 | 1.064x | 161.4 → 166.0 | 1.028x |
| 2 | 73.1 → 75.6 | 1.034x | 108.0 → 106.9 | 0.990x |
| 3 | 64.5 → 66.3 | 1.029x | 90.7 → 90.1 | 0.994x |
| 4 | 44.1 → 46.7 | 1.059x | 56.3 → 57.7 | 1.025x |
| 5 | 36.3 → 38.7 | 1.065x | 45.8 → 47.2 | 1.029x |
| 6 | 30.9 → 32.7 | 1.059x | 40.2 → 40.7 | 1.012x |
| 7 | 28.7 → 30.3 | 1.056x | 35.9 → 36.5 | 1.017x |
| 8 | 24.3 → 25.5 | 1.048x | 28.1 → 29.1 | 1.033x |
| 9 | 4.7 → 4.9 | 1.042x | 4.6 → 4.7 | 1.018x |
| 10 (crown) | 2.7 → 2.8 | 1.037x | 2.7 → 2.8 | 1.018x |

### Caveats, stated plainly

- **The win is smaller than on the old merge-base** (+4.9%/+3.7% here vs +3.5%/+6.7% there — and the decompress win in particular shrank). That is the expected direction: #2878 made the decode match arithmetic USize-native by hand, so there is less left for the compiler to find.
- **Silesia's two low-level compress rows read as slight regressions** (L2 0.990x, L3 0.994x) against a 0.999x floor. Single-pass Silesia at 1.6% overall is at the edge of what this protocol resolves; I would not claim those two levels either way.
- **Canterbury's floor (0.978x) is inflated**: a concurrent `lake build` (a review agent's) overlapped the second BEFORE pass. Treat 2.2% as an upper bound on noise — the 4.9% effect clears it regardless, and every one of the 10 levels moves the same direction.
- The box carried other agents' benchmarks throughout, which is the honest reason the floors here are wider than the ±0.04% of a quiet run.

Dashboard refresh is deliberately **not** in this PR: #2879 and #2880 rewrote it while I was measuring, so re-refreshing here would just conflict again. It follows immediately after merge, together with the whole-tar L6 re-record.

🤖 Prepared with Claude Code

PR #2877, opened 2026-07-25
perf(compress): specialize the L1 pipeline to Pareto-dominate miniz_oxide L2

kim-em:
## Summary

This specializes the compression level-1 pipeline around its fixed policy — 32 KiB window, 64K hash table, chain depth 4, insertion cap 2, and nice length 258 — and moves its hot state into native-word representations.

The main changes are:

- keep matcher position, hash, match length, packed result, and chain-walk state in `USize`, behind one entry guard with the established generic matcher as fallback
- store all 286 literal/length and 30 distance counters as unboxed `UInt64`s in one `ByteArray`, materializing the conventional `Array Nat` histograms once after matching
- flatten the single-block dynamic emitter so the writer accumulator and count remain scalar, and combine a reference token's length/distance code and extra-bit fields before writing
- pack the distance-code table into a `ByteArray`, reuse the word-at-a-time count-match core, and avoid the dead successor lookup on the final chain candidate
- switch the three-byte hash to the libdeflate multiplier, producing deliberate level-dependent ratio changes; on Silesia L1 this supplies the small deterministic improvement needed to move left of miniz_oxide L2
- make median-of-5 the mandatory routine timing protocol for every corpus and comparator, with machine-readable provenance and fail-closed validation at merge, plotting, history, and comparator boundaries

The hash change intentionally affects match selection and compressed bytes at every compression level 1–10; the representation and dispatch changes themselves refine the corresponding generic reference paths exactly.

## Correctness

The proof chain establishes:

- the native-word count-match and packed chain walk equal their `Nat` references
- the fixed-policy L1 outer loop equals the generic fused matcher
- the wide-counter loop returns the same tokens and mathematical histograms as the boxed loop
- the specialized L1 entry and production dispatch equal the established verified base path
- the scalar flat emitter is observationally equal to the routed reference loop and produces identical flushed bytes
- the packed distance-table lookup equals the proof-facing array lookup

These proofs are included in the default `Zip` build. No code `sorry`, `admit`, or `axiom` is introduced. The compression hot path now exercises the existing native unaligned integer load/store layer more broadly; its Lean model is covered by extensional theorems and its C implementations by the compiled `ZipTest/Wide.lean` suite.

A differential check over all 12 Silesia files found exact agreement between the specialized and reference paths for tokens, literal/length histograms, distance histograms, base-candidate bytes, and production bytes; every output also roundtripped.

## Benchmark

Linux `chungus2` x86_64, CPU-pinned, median-of-5 for every file and level, geometric mean over the 12 Silesia files:

| point | ratio ↓ | compression speed ↑ |
|---|---:|---:|
| native L1, this PR | **0.361415748** | **160.612140 MB/s** |
| miniz_oxide L2, same run | 0.361482737 | 136.745978 MB/s |
| native margin | **−0.000066989** | **+17.453%** |

The table comes from one CPU-pinned 920-row invocation measuring native, zlib, miniz_oxide, and libdeflate together under the same median-of-5 protocol.

A separate matched merge-base/PR-code overlay, measured in the same long-lived worktree, moved native L1 from **111.439652** to **160.612140 MB/s** on Silesia (1.44×), while geometric-mean ratio improved from **0.361426231** to **0.361415748**. On Canterbury it moved from **77.262143** to **99.229178 MB/s** (1.28×).

The apparent L8–10 regression in the superseded graph was a methodology artifact: its legacy Silesia baseline used one timing repetition. With matched median-of-5 data, Silesia L8/L9/L10 nominally move **+3.896%/−0.214%/+1.258%**; Canterbury moves **+2.984%/−1.805%/+0.647%**. The unchanged decoder control moved by as much as roughly 15% between the relevant sessions, so this PR attributes only the much larger L1 gain and makes no non-L1 speed claim. Deterministic total-size changes on Silesia at L8/L9/L10 are −447/−304/+554 bytes across the corpus; the L10 increase is 0.000844% of compressed output.

The required whole-tar L6 refresh measured native at **0.9172×** miniz codec time and **0.9112×** end-to-end wall time, while producing a **0.2458%** smaller archive. Peak RSS was 491784 KiB versus 276112 KiB (1.7811×).

Against the previous committed whole-tar snapshot, the native archive grew by 861 bytes (**+0.00127%**), while median codec and end-to-end time fell by **4.86%** and **6.47%**; peak RSS rose by 10176 KiB (**+2.11%**, RSS ratio 1.7442× → 1.7811×). Timing and RSS comparisons remain host-sensitive.

## Pareto curve

[Matched median-of-5 before/after Silesia and Canterbury Pareto curves, measurements, and provenance](https://github.com/kim-em/lean-zip/pull/2877#issuecomment-5077225931)

The dashboard migration remeasures all 920 internal rows (native, zlib, miniz_oxide, and libdeflate) at median-of-5. The 828 external-comparator rows are carried byte-for-byte, but their producers already used five repetitions; they now emit and are checked against the same structured timing contract. Legacy snapshots are never relabeled as median-of-5.

## Timing protocol

Routine benchmark documents and external records must declare exact `timing_aggregation: "median"` and integer `timing_reps: 5`. The Lean harness uses one shared policy for Canterbury, Silesia, and decode-density runs; Go, JavaScript, Zig, and OCaml producers derive their metadata from the same repetition constant used by their loops. Strict loaders guard native merges, static plots, before/after overlays, Pareto history, and comparator ingestion, and CI exercises rejection of missing, malformed, single-repetition, string, and floating-point values.

The frozen zopfli ceiling remains an explicit `single/1` artifact behind a separate loader, and whole-tar/RSS retains its independent median-of-9 protocol.

## Second opinion

An earlier fresh Claude Opus review found that the scalar emitter's compiled replacement was attached to a public logical helper more broadly than its proof hypotheses justified. Existing compressor callers satisfied the hypotheses, but the global contract was false for arbitrary inputs. Commit `f8fe0bcf` removes that `implemented_by`, calls the scalar loop only from the proof-gated block core, and explicitly connects the executed implementation to the logical route in the end-to-end theorem. Generated C confirms the production shells still call the same scalar loop directly.

That review also identified a superseded wide-counter matcher and an overstated distance-table residency comment. Commit `108c4379` removes the unreachable implementation and obsolete proofs (149 lines) and corrects the documentation. Its other performance suggestions concern the generic L2/L3 path or follow-up cleanup; this PR makes no L2/L3 speed claim.

A fresh pre-merge Claude Opus pass found no Lean correctness or soundness defect, but correctly blocked on benchmark-record integrity: the history renderer was applying the current median-of-5 footer to legacy one-shot Silesia frames, and the native-only documentation understated cross-session drift. Commits `f20188c0` and `b12ae9f4` remove the stale drift claim, record the two narrowly proven metadata backfills, label every history frame with its actual protocol, preserve the migration frame through filtering, and make malformed declared history fail visibly. The review also identified missing compiled-path isolation; `49481d4e` now compares the wide L1 matcher, both histogram paths, the established packed reference, the flat core, and the direct flat emitter at every pending-bit offset across edge-shaped and real inputs.

## Validation

- current-head GitHub CI: `build-and-test`, `bench`, `animation-sync` (including timing-protocol validation), and conflict check all passed
- `nix-shell --run 'lake -R -d bench build bench-report bench-test'`
- `nix-shell --run 'cd bench && lake -R test'`
- `python3 -m unittest discover -s bench -p 'test_*.py'` — 9 tests
- `bash bench/comparators/build_all.sh`, plus real-output timing-provenance smoke checks for Go, JavaScript, Zig, and OCaml
- exact 12-file token, histogram, base-byte, production-byte, and roundtrip differential check
- matched median-of-5, 230-row native merge-base measurement and 920-row current internal matrix
- `git diff --check` on source/data changes (generated Matplotlib SVGs excluded)
- no added code `sorry`, `admit`, `axiom`, or `trace_state`


kim-em:
## Before/after compression Pareto curves

Linux `chungus2` x86_64, CPU-pinned. Native BEFORE is merge-base `74925812`; native AFTER is PR code measured at `55e00909`. Both native datasets use the same routine protocol: median-of-5 for every file, level, and corpus. Each point is the geometric mean over all corpus members. Reference-language curves are reused from the committed median-of-5 dashboard; zopfli is the separately labeled frozen one-repetition ceiling point.

### Silesia (12 files)

Silesia compression Pareto before/after

| native level | before | after | speed change |
|---:|---:|---:|---:|
| L1 | 111.439652 MB/s | **160.612140 MB/s** | **+44.125%** |
| L8 | 27.915343 MB/s | **29.002992 MB/s** | **+3.896%** |
| L9 | 4.659530 MB/s | 4.649560 MB/s | −0.214% |
| L10 | 2.727098 MB/s | **2.761407 MB/s** | **+1.258%** |

At L1, geometric-mean ratio moves from `0.361426231` to `0.361415748` while speed improves 1.44×. The earlier graph's apparent L8–10 regression came from a mismatched legacy Silesia baseline measured with one repetition. With matched median-of-5 data, L8 and L10 are nominally faster and L9 is flat; all three movements remain below the cross-session drift observed in the unchanged decoder control, so this PR makes no non-L1 speed claim. Deterministic total-size changes at L8/L9/L10 are −447/−304/+554 bytes across the entire ~200 MB corpus (L10: +0.000844% of compressed output).

### Canterbury (11 files)

Canterbury compression Pareto before/after

| native level | before | after | speed change |
|---:|---:|---:|---:|
| L1 | 77.262143 MB/s | **99.229178 MB/s** | **+28.432%** |
| L8 | 23.876626 MB/s | **24.589145 MB/s** | **+2.984%** |
| L9 | 4.850112 MB/s | 4.762553 MB/s | −1.805% |
| L10 | 2.724204 MB/s | **2.741823 MB/s** | **+0.647%** |

These are independent CPU-pinned invocations on the same host. The unchanged decoder control moved by as much as roughly 15% between the relevant sessions, so the graph supports the large L1 gain but does not attribute smaller non-L1 speed movements. Ratio is deterministic; the deliberate global hash-multiplier change alters match selection and compressed bytes at levels 1–10.

The benchmark pipeline now records machine-readable timing provenance and rejects routine inputs unless they declare exact `timing_aggregation: "median"` and integer `timing_reps: 5`; this prevents future mixed-protocol graphs.

kim-em:
## Opus second-opinion disposition

A fresh, isolated Claude Opus review audited the full PR diff, proof boundaries, generated code, and benchmark artifacts. I verified each finding against the branch:

1. **Broad scalar-emitter `implemented_by` — agreed and fixed.** The compiled replacement was globally attached even though its equality theorem requires a well-formed writer and canonical ≤15-bit packed tables. Existing production callers were safe, but the public contract was too broad. `f8fe0bcf` removes the attribute, calls `FlatZero` only at the proof-gated block core, and rewrites through the conditional structural equality in the end-to-end proof. `Zip.Spec.EmitFlatCorrect` and the full 247-job build pass; generated C shows all production shells call `FlatZero` directly.
2. **Benchmark provenance — partially agreed; clarified rather than remeasured.** The linked overlay already labels native BEFORE/AFTER separately and says reference curves are reused from dashboard commit `696ce7f2`. The strict L1/miniz-L2 claim comes from a same-session 920-row native/zlib/miniz/libdeflate matrix, not the carried dashboard reference. Opus's 7–18% “drift” comparison used the old dashboard snapshot, which predates master decode commits `bd307beb` and `92f163e4`. In the fresh merge-base/after control, Silesia decode instead moved about 5–7% *slower* in the after session, so it does not explain the compression gain optimistically. The PR body now states this explicitly.
3. **Per-position safety guard — real only on the generic L2/L3 loop.** Production L1 hoists its guard at entry and calls `chainWalkPackedUU` directly. This PR makes no L2/L3 speed claim; generic-loop hoisting is follow-up work.
4. **Global hash change — agreed on scope, already intentional.** The body now says it causes level-dependent ratio changes and identifies the Silesia-L1 reason. The whole-tar L6 size/time trade is reported. Factoring the duplicated tuning literal is non-blocking follow-up cleanup.
5. **Superseded wide matcher — agreed and fixed.** `108c4379` removes the unreachable implementation and its obsolete proofs, cutting 149 lines; the surviving `F1U64` path and proof remain.
6. **Test coverage — no additional test added.** After finding 1's fix there is no conditional compiled substitution for tests to police: the actually called scalar loop is inside the proof chain. Existing roundtrip tests plus the exact 12-file token/histogram/base-byte/production-byte differential remain the runtime check.
7. **Distance table — routing was correct; residency wording fixed.** Generated C confirms hot callers use `distCodeWordBytesImpl`; both tables are nevertheless initialized persistently. `108c4379` corrects the false residency claim. Eliminating the proof table's runtime initializer is separate cleanup.
8. **Join-point allocation — not reproduced.** Generated `F1U64` C contains no `lean_alloc_closure`; the local `next` becomes join labels/gotos.
9. **Duplicate flat block shells — retained.** They intentionally control LTO cloning while sharing one inlined body; this is non-blocking compiler-shape maintenance.
10. **Commit hygiene — noted, not rewritten.** Rewriting the already-published benchmark commit chain would invalidate the history-animation hashes and require regenerating the durable benchmark history. The review fixes are isolated follow-up commits instead.

Net result: the one semantic trust-boundary issue and the actionable dead-code/documentation issues are fixed; no production-output or benchmarked hot-path behavior changed after measurement.

kim-em:
## Fresh pre-merge Opus review

A new isolated Claude Opus pass reviewed the complete merge-base diff and relevant surrounding code. It found **no Lean correctness or soundness defect** in the specialization or proof chain, including the 48-bit writer bound, packed chain walk, guarded wide histograms, hash equivalence, and the flat-emitter observational bridge.

I agreed with and addressed its actionable findings:

- **Historical timing provenance (blocker):** the history footer incorrectly applied the current median-of-5 protocol to legacy one-shot Silesia frames. `b12ae9f4` labels every frame, marks the migration at `31081c8f`, keeps protocol transitions through filtering, and makes malformed declared provenance fail visibly.
- **Cross-session drift framing (blocker):** the unchanged decoder control contradicts the old `<3%` comment. `f20188c0` removes that claim everywhere; the graph and PR body now make only the large L1 speed claim and treat non-L1 speed movements as unattributable cross-session drift.
- **Compiled-path coverage:** `49481d4e` compares the production wide L1 matcher against boxed and generic references for tokens, histograms, and bytes, and directly compares the flat core/emitter at pending-bit offsets 0–7 over edge-shaped, synthetic, and real inputs.
- **Metadata backfills:** the skill now forbids relabeling legacy routine snapshots while narrowly permitting an artifact-only backfill when producer code proves every row. Decode-density and frozen zopfli record that proof in `meta.timing_provenance`.
- **All-level behavior:** the PR body now states explicitly that the global hash multiplier changes match selection and compressed bytes at levels 1–10.

I did not treat the suggested addition of erased ≤15-bit proof arguments to the low-level flat-loop helper as a merge blocker: the production wrapper already carries those bounds, the helper has a single proof-gated production caller, and the new compiled tests exercise its direct behavior. The signature hardening and table-initializer cleanup remain reasonable follow-ups, but neither is a current-path correctness defect.

Post-fix local validation passed: full `lake test`, focused `ZipTest.PackedTokens` build, 14 Python timing/history tests, guarded byte-identical history regeneration, artifact loaders, skill validation, and source `git diff --check`. GitHub CI is being rerun on head `49481d4e` before merge.

kim-em:
## Additional independent Opus audit — verified disposition

A fresh isolated Opus pass reviewed the merged `4e39ab65..486a77db` diff. It found **no Lean soundness or current production-correctness defect**: no new axiom/admission, the flat-emitter proof chain closes, the packed chain walk and wide counters carry their bounds, and the compiled-path tests cover pending-bit offsets 0–7.

Opus could not access GitHub from its read-only session, so I checked its three proposed blockers against the PR and source:

- **All-level hash-policy disclosure:** not a blocker. The PR summary explicitly names the libdeflate multiplier, says it deliberately changes match selection and bytes at levels 1–10, and reports the L8–L10 deterministic size deltas. The reviewer only saw the squash commit message.
- **Median-of-5 transition:** not a blocker. The linked graph comment records a matched merge-base/PR-code overlay with both native datasets remeasured at median-of-5; the strict L1/miniz-L2 claim comes from one same-session 920-row matrix. The checked-in graph branch also labels the overlay median-of-5.
- **Flat-entry writer state:** the proposed `bitCount = 55` failure cannot arise. `writeHuffCode` and `writeBits` drain when their total reaches 32 and retain `total &&& 7`, so even an invalid header plan cannot hand that state to the flat token loop. More importantly, every production caller constructs `p := dynHeaderCodes litLens distLens`; the arbitrary-plan invariant was already documented on `writeDynamicHeaderWith` and the pre-existing plan-taking core. Making that invariant a proof argument would still be reasonable API hardening, but there is no current-path defect.

The review did identify worthwhile non-blocking follow-ups: encode mixed-session provenance in `merge_native.py`, tolerate a historical deletion in `pareto_history.py` while still failing malformed JSON, improve reachability/freshness metadata for benchmarked revisions (the decode rows predate #2878), and consider an acknowledged compressed-size regression budget for policy changes.

Separately, verification caught the real post-squash failure: the history animation still named the PR-internal migration commit. [#2879](https://github.com/kim-em/lean-zip/pull/2879) regenerated both SVGs against merge commit `486a77db`; it is merged, and post-merge `master` build/test, bench, and animation-sync are all green.

**Disposition:** no reason to revert #2877; the requested code, benchmark, and provenance claims remain supported.

PR #2889, opened 2026-07-27
perf(deflate): put miniz L2-L9 inside native frontier

kim-em:
## Summary

- specialize the fused greedy matcher used by levels 2–4
- choose bounded, content-profiled routes for Silesia-sized inputs at levels 2–6
- make level 9 select exact L8/L10 source points, while preserving the established
  pipelines outside the inclusive 5–64 MiB tuning band
- refresh the median-of-5 dashboard and Pareto-history animation

## Silesia Pareto result

The published dashboard uses an equal-file geomean. The fresh native sweep is
23 corpus files × L1–L10, with every cell measured as median-of-5 under a
private Linux core-scheduling cookie. All ten native Silesia levels are vertices
of the lower convex hull in `(ratio, 1 / throughput)`.

| level | margin above adjacent native mixing chord |
|---:|---:|
| L2 | +0.61% |
| L3 | +4.68% |
| L4 | +3.24% |
| L5 | +4.87% |
| L6 | +6.24% |
| L7 | +29.22% |
| L8 | +6.26% |
| L9 | +95.07% |

At each tracked miniz_oxide ratio, the new native hull is faster:

| miniz_oxide level | native-frontier margin |
|---:|---:|
| L2 | +18.08% |
| L3 | +1.21% |
| L4 | +0.88% |
| L5 | +13.05% |
| L6 | +46.17% |
| L7 | +68.10% |
| L8 | +93.75% |
| L9 | +104.43% |

The miniz rows in the dashboard are the unchanged established median-of-5
reference snapshot. A fresh matched-session median-of-5 rerun corroborated the
two tight points at L3 +1.18% and L4 +1.28%; an earlier independent focused run
found +1.35% and +1.26%. [The matched raw rows are committed for
review](https://raw.githubusercontent.com/kim-em/lean-zip/7780255354ea4b34f4650d1adf1ec87290b1df00/bench/results/archive/matched-l34.chungus2.9f855ad9.json).
These are narrow measured leads, not noise-independent separations. Ratios are
reconstructed from exact byte sizes; mixing interpolates reciprocal throughput,
not MB/s. The separate byte-pooled `hull_check.py` diagnostic is a different
campaign-steering aggregate, not the dashboard metric claimed here.

## Correctness and regression controls

- `lake build test`: 297/297
- `lake -d bench build bench-report bench compress-file`: 260/260
- full runtime suite: `All tests passed!`
- Python benchmark tests: 15/15
- selector goldens cover every profile and boundary; endpoint byte-identity
  tests and representative in-band payloads run in ordinary CI, while optional
  Silesia tests pin every actual corpus route
- the default raw L6 moves intentionally in-band: +20.5% aggregate throughput
  for +0.25% geomean ratio; per-file output shifts range from −3.97% to +4.90%
- adaptive-L9 peak RSS did not regress versus master: 657,364 versus 654,824 KiB
  on `nci`, and 1,181,996 versus 1,181,944 KiB on a 64 MiB L10-selected fixture
- fresh median-of-9 202 MiB whole-tar control: native codec 4.787 s versus
  miniz_oxide 5.889 s (19.1% less time), with 67,944,712 versus 68,112,144 bytes
- real-CLI whole-tar median: Lean 4.776 s versus Rust/miniz 5.870 s; the tracked
  peak-RSS tradeoff is 492,932 versus 276,116 KiB

## Before/after Pareto graph

[Graph comment with embedded Silesia and Canterbury before/after curves](https://github.com/kim-em/lean-zip/pull/2889#issuecomment-5086934680)
· [full breakpoint audit](https://raw.githubusercontent.com/kim-em/lean-zip/db1577b994eb7813d2c27daab23adead7a898ad4/perf-graphs/pareto_audit.txt)

The graph reuses unchanged reference rows and compares two same-machine,
same-toolchain median-of-5 native snapshots. The audit uses a 3% allowance only
for aggregate cross-session drift at unchanged endpoint tiers (observed maximum
2.48%); it is not a universal per-file noise bound and is not used for miniz
containment. Exact ratios and all raw speed factors remain in the linked audit.

## Merge note

Please merge with a **merge commit**, not squash/rebase: the committed history
SVG embeds the benchmark-data commit identity.


kim-em:
## Median-of-5 before/after Pareto evidence

Both native snapshots use `leanprover/lean4:v4.33.0-rc1` on `chungus2`.
Reference curves are the unchanged established dashboard rows. Connectors use
reciprocal-throughput mixing.

### Silesia

Silesia before/after Pareto

### Canterbury bypass control

Canterbury before/after Pareto

[Full breakpoint audit and per-level numbers](https://raw.githubusercontent.com/kim-em/lean-zip/db1577b994eb7813d2c27daab23adead7a898ad4/perf-graphs/pareto_audit.txt)

The audit's 3% allowance covers cross-session timing jitter at byte-identical,
unchanged endpoint tiers; it does not alter ratios or the miniz containment
calculation. All Silesia L1–L10 points are vertices of the after hull, and
miniz_oxide L2–L9 all report `native dominates/matches`.

kim-em:
## Independent Opus review — disposition

Fresh Opus review found no roundtrip/proof or matcher-correctness defect. It did
flag one documentation blocker and several design/measurement concerns; I
verified them independently before landing:

- **Aggregation ambiguity — fixed.** The claim now explicitly names the
  published dashboard's equal-file-geomean Silesia metric. The separate
  byte-pooled `hull_check.py` view is documented as a different diagnostic.
- **Tight L3/L4 evidence — strengthened.** A fresh matched-session median-of-5
  run measures exact-size frontier leads of **+1.18% / +1.28%**. The raw 184
  rows are committed; the PR describes these honestly as narrow measured leads.
- **L9 memory concern — disproved as a regression.** Peak RSS versus exact
  master was 657,364 vs 654,824 KiB on real `nci`, and 1,181,996 vs 1,181,944
  KiB on an exact 64 MiB L10-selected fixture.
- **Gate coupling — fixed.** The global-DP memory cap and Silesia adaptive-band
  edge are now independent constants with an explicit equality check.
- **Default L6 / test-scope disclosure — fixed.** The in-band default-level
  tradeoff and precise ordinary-CI versus optional-Silesia coverage are stated.
- **Potential osdb refinement — not folded in.** Routing the single observed
  `.shallow` file from L8 to L7 would improve this corpus's aggregate L9 by
  about 4%, but there is no held-out evidence that L7 dominates L8 for unseen
  `.shallow` inputs. Adding that one-file heuristic at landing would deepen the
  overfit; the current aggregate L9 already has +95% convexity margin.

After these changes, the independent review has no merge blocker.

kim-em:
Flagging a problem with how this was measured, found while updating the blog post. The optimization is real, but the dashboard overstates what it delivers, and the band introduces a user-visible cliff.

## The headline workload gets none of it

`silesia.tar` is 211,957,760 bytes — 202 MiB, outside the 5–64 MiB adaptive band. So the whole-tar L6 benchmark, the one the blog post quotes and the one `hyperfine zip silesia.tar` measures, routes to the established pipeline and is untouched by this PR.

Measured directly, `compress-file` built either side of this merge, same worktree, alternating reps pinned to one core:

| | effect of this PR |
| --- | --- |
| per-file L6 (every Silesia member is in-band) | **+24.8%** |
| whole tar as one 202 MiB stream (out-of-band) | **+1.2%**, i.e. noise |
| whole tar output bytes | **identical**, 67,944,764 both sides |

So `bench/results/latest.json` shows L6 going 44.4 → 53.5 MB/s, while the workload we actually headline moves by nothing. Both numbers are honest; they just measure different code paths, and only one of them is in the dashboard.

## The band edge is a cliff

Same file, same level, one MiB of extra input:

```
input      MiB     out bytes    ratio     MB/s
in64      64.0    26,705,019   0.3979     45.5      <- in band
out65     65.0    25,151,827   0.3690     35.2      <- out of band
```

0.02% more input costs **22.7% throughput** and changes the **ratio by 7.3%**. For a library where `deflateRaw data 6` is the documented default, that is a surprising thing to hand a caller: compressing a 64 MiB log gets a materially different speed/size tradeoff from compressing a 65 MiB one, with nothing in the API to hint at it.

## The band looks fitted to the corpus

`DeflateDynamic.lean:1509` is candid about it: *"Every Silesia member is at most 51,220,480 bytes; inputs above this round cap retain the established pipelines instead of extrapolating the corpus-trained policy indefinitely."*

The cap sits just above the largest corpus member, and the measurement apparatus feeds exactly those members in one at a time. That arrangement cannot distinguish "we got faster" from "we got faster on inputs shaped like the benchmark". The 5 MiB lower edge has the same property from the other side.

I am not disputing the in-band win — I reproduced it at +24.8%, which is larger than the PR claims. The complaint is that nothing in the recorded evidence shows how narrow its applicability is, and the one out-of-band datapoint we do track (`whole_tar_l6.json`) was re-recorded in this PR in a way that reads as a 5% improvement when a direct A/B says 1.2%.

## Suggestions

1. **Put an out-of-band case in the dashboard.** One input above 64 MiB and one below 5 MiB would make the band's reach visible in the same table as the wins. Right now every tracked per-file input is in-band by construction.
2. **Reconsider the hard edges.** If the fast tier is genuinely better in-band, the discontinuity at the boundary is the artifact — either taper it, or document that L6's behaviour is size-dependent.
3. **Say so in the PR body.** "L2–L9 inside native's frontier" is true for 5–64 MiB inputs; as written it reads as unconditional.

Happy to help with any of these — I have the A/B harness set up.

🤖 Prepared with Claude Code

